PR #35, opened 2016-01-16
add podspec

ohayon:
https://github.com/JakeLin/IBAnimatable/issues/7


JakeLin:
@ohwutup  Thank you, I am also working on https://github.com/JakeLin/IBAnimatable-PackageManagementExamples to create example projects to use package management tools.

PR #37, opened 2016-01-17
Add MaskType.Star.

lexrus:
Add `.Star`, `.Star4`, `.Star6`, `.Star8`, `.Star10` to MaskType. We can mask views like this:
screen shot 2016-01-17 at 11 51 31 pm

Conclusion:
1. These stars are masked with CAShapeLayer which mask shadows as well. Thus we can't have shadows under star views. (Stars have no shadow in the real world neither. 🤓)
2. Border of star is not rendered properly. I already have a workaround in my Playground which is not suitable(maybe next PR) for IBAnimatable:
screen shot 2016-01-18 at 12 03 04 am


JakeLin:
Thanks you very much again. May have another sub `CAShapeLayer` to support a fake shadows, but I don't want to over engineer that. I think current solution is very good. :+1:

PR #38, opened 2016-01-18
Move IBAnimatable.playground into xcodeproj.

lexrus:
So that users can play with IBAnimatable directly.
screen shot 2016-01-18 at 10 49 51 am


JakeLin:
Thanks again. :+1:

PR #41, opened 2016-01-19
Make Framework iOS 8.x compatible

YoniTsafir:
iOS 8.x is still 15-20% of the users as of writing this.
Not every app developer can afford losing all this traffic.

The only thing that currently breaking compatibility is the use of UIStackView, then by marking AnimatableStackView `@available(iOS 9, *)` I was able to make the entire framework iOS 8.x compatible.


JakeLin:
@theyonibomber Good one. I thought about it. Thank you very much. Another similar one is to support Objc, I won't start unless someone asks for. We should Swift more.

@lexrus good pickup.

PR #30, opened 2016-01-16
Fix typos and improve readability of README.

lexrus:
Fixed typos in README.md.
Improved readability of README.zh.md according to [中文文案排版指北](https://github.com/sparanoid/chinese-copywriting-guidelines).

Additionally, we should set `APPLICATION_EXTENSION_API_ONLY` to YES for Frameworks. See https://github.com/Alamofire/Alamofire/issues/268

PR #43, opened 2016-01-19
Improve compatibility for extension again.

lexrus:
Fix #30 again.


lexrus:
Uh... I guess your Travis is not configured properly. Otherwise here must be a comment from Travis.

JakeLin:
Thanks again, we can use it in today extension now.

Issue #8, opened 2015-12-28
More MaskDesignable


JakeLin:
Currently, `MaskDesignable` only supports `Circle`, can support more mask like `Triangle` and `Star`.

lexrus:
Triangle mask is easy to support but every triangle must has a direction obviously.

``` swift
enum MaskType {
  ...
  case TriangleUp
  case TriangleDown
  case TriangleLeft
  case TriangleRight
}
```

JakeLin:
One difficulty I am having is to support different parameters with different options. For example, for `Circle`, we don't need an angle for it. But for `Triangle` or `Star`, we may need a starting degree.

I am working on `Star` based on your PR. Will ask your favour to review it once I submit a PR. I cam up an idea to put parameter in the MaskType string. Because IB doesn't support enum as `IBInspectable`'s type, we have to use `String` then convert the `String` to the enum. But we can also put more value than the raw value of the enum. e.g. `Star(6)` for a star with 6 points. That may provide a workaround for similar situation.

lexrus:
Yeah. I were thinking about using associated types but fail back to strings at last.
There's another workaround:

``` swift
#if TARGET_INTERFACE_BUILDER
  @IBInspectable public var maskType: String = ""
  @IBInspectable public var triangleDirection: String = ""
  @IBInspectable public var starPoints: Int = 0
#else
  public var maskType: MaskType = .None
#endif
...

enum TriangleDirection {
  case Up, Right, Down, Left
}

typealias StarPoints = UInt

enum MaskType {
  case None
  case Circle
  case Star(StarPoints)
  case Triangle(TriangleDirection)
}
```

JakeLin:
I really like your idea for 

```
#if TARGET_INTERFACE_BUILDER
  @IBInspectable public var maskType: String = ""
#else
  public var maskType: MaskType = .None
#endif
```

I will try to replace all `String` to actual `enum` in one PR. But will introduce breaking changes to the API when we use them in code.  Thanks again.

JakeLin:
For triangle direction, we already have a property called `rotate` in `RotationDesignable`, I steal this idea from Sketch. Unfortunately, it doesn't render properly in Interface Builder. 

[...]

PR #44, opened 2016-01-21
Workaround to redirect setter of animationType.

lexrus:
Just a tricky & dirty PoC for #8.
We should develop a Xcode Plugin for IBInspectable enum or wait for Xcode 8...


JakeLin:
Thanks @lexrus I will keep going with this and will give IB team this feedback.

Issue #46, opened 2016-01-22
Method slideOut won't slide out everytime


tbaranes:
Following the code, the `slideOut` methods are always `300 * force`, but if we are running this example on an iPad with a force to 1, it will never get the outside screen.

Is there a reason to have hardcoding that constant?

JakeLin:
Will add `SlideFadeOutLeft` which is like `SqueezeFadeOutLeft`.

tbaranes:
:+1:

tbaranes:
After check the implementation of `SqueezeFadeOutLeft`, we will have the same issue, the `slideOut` should go outside of the screen, whatever is the distance between it's current position, and the next one, isn't it?

JakeLin:
@tbaranes you are right. that is the problem of `SlideOut` animations. The slide out animations require the width or height of the device. And the animating element itself doesn't have the info. All we can use `UIScreen. mainScreen. bounds` to get. Or we can add some animations like `moveX()`, `moveY()`. That may fix this issue. Do you have any suggestions?

tbaranes:
We shouldn't use `UIScreen` to get the frame since the container where we are doing the animation could have any frame depending of the context.

Something like that would be more appropriate:

``` swift
  public func slideOutLeft(completion: AnimatableCompletion? = nil) {
    let x = -frame.width * force
    animateOutWithX(x, alpha: 1, completion: completion);
  }

  public func slideOutRight(completion: AnimatableCompletion? = nil) {
    guard let unwrappedSuperview = self.superview else {
      return
    }

    let x = unwrappedSuperview.frame.maxX * force
    animateOutWithX(x, alpha: 1, completion: completion);
  }
```

The only bad point is that we can't use `Animatable` without an `UIView` but it won't be a big issue, is it?

tbaranes:
After a better looking in the code, the solution would be more something like this:

``` swift
  public func slideOutDown(completion: AnimatableCompletion? = nil) {
    guard let unwrappedSuperview = self.superview else {
      return
    }

    let y = (unwrappedSuperview.frame.height - frame.minY) * force
    animateInWithY(y, completion: completion)
  }
```

I will make some tests, and a PR if that's working on monday

JakeLin:
@tbaranes My concern is the animating view may not be the directly subview of the the root view. For example, VC.view -> UIStackView(not 100% width and height) -> UIView(one item within StackView) -> AnimatableView.

[...]

PR #49, opened 2016-01-22
Removing hardcoding constants for out animation

tbaranes:
As suggested in #46, this solution will be more relievable.
What do you think?


tbaranes:
My bad, I misunderstand the implementation. I will open another clean PR once I have a better fix.

PR #51, opened 2016-01-25
Add rotate animation

tbaranes:


JakeLin:
Cool @tbaranes I am thinking `repeatCount` can have `Infinite` or have `.RotateForever` in enum. That would be useful for  lots of cases. But don't know to how support `Infinite` in `@IBInspectable` property.

lexrus:
How about `let Infinite = -1`?

tbaranes:
I though about this, but I conclude that the user will set infinite as the repeatCount. @lexrus suggestion is a bit hidden (if you don't now, hard to find), but maybe the nicest one solution?

JakeLin:
`Float` does have `Float.infinity`, that's why I chose `Float` for `repeatCount` instead of `Int`. And `repeatCount` should be `Int` realistically. We can use `Float.infinity` in code. but can't do it in Interface Builder. I think `let Infinite = -1` is not a bad idea for IB because we never have negative `repeatCount`. 

On the other hand, I found out the documentation is also important for `IBAnimatable` project, if we have `let Infinite = -1`, we need to update the API document to describe this workaround.

JakeLin:
The reason why I said documentation is also important  because Interface Builder doesn't support validation for the value and the range. When I observed the others used `IBAnimatable` and the API document can help out the supported options and value range.

PR #54, opened 2016-01-27
Fix left gradient

tbaranes:


JakeLin:
@tbaranes Thanks again

Issue #50, opened 2016-01-23
Replace String with enum for the properties are not using in IB


JakeLin:
Because `@IBInspectable` in Interface Builder only support certain types. eg. CGFloat, UIColor, String and Int. But can not support enum inherited from `String`. eg. `enum AnimationType: String`.

@lexrus came up a great idea to use enum in programming API and use `String` in IB.

```
#if TARGET_INTERFACE_BUILDER
  /**
    String value of predefined Animation Type, all supported types are in `AnimationType` enum
  */
  var animationType: String? { get set }
#else
  /**
   Predefined Animation Type, all supported types are in `AnimationType` enum
   */
  var animationType: AnimationType? { get set }
  var animationTypeRaw: String? { get set }
#endif
```

Then when the App runs on simulator (not in IB), the compiler actually uses _non_ `TARGET_INTERFACE_BUILDER`. We have to convert the String value in Storyboard to `AnimationType`. Method Swizzling on `NSKeyedUnarchiver` is one way to do that.

I have tried to use Swift protocol extension.

```
// Use Method Swizzling to convert `String` to emun when loading runtime attributes from Storybaord.
// http://nshipster.com/swift-objc-runtime/
public extension NSKeyedUnarchiver {
  public override class func initialize() {
    struct Static {
      static var token: dispatch_once_t = 0
    }

    // make sure this isn't a subclass
    if self !== NSKeyedUnarchiver.self {
      return
    }

    dispatch_once(&Static.token) {
      let originalSelector = Selector("setValue:forUndefinedKey:")
      let swizzledSelector = Selector("iba_setValue:forUndefinedKey:")

      let originalMethod = class_getInstanceMethod(self, originalSelector)
      let swizzledMethod = class_getInstanceMethod(self, swizzledSelector)

      let didAddMethod = class_addMethod(self, originalSelector, method_getImplementation(swizzledMethod), method_getTypeEncoding(swizzledMethod))

      if didAddMethod {
        class_replaceMethod(self, swizzledSelector, method_getImplementation(originalMethod), method_getTypeEncoding(originalMethod))
      } else {
        method_exchangeImplementations(originalMethod, swizzledMethod)
      }
    }
  }

  // MARK: - Method Swizzling
  func iba_setValue(value: AnyObject?, forUndefinedKey key: String) {
    if (key == "animationType") {
      setValue(value, forKey: "animationTypeRaw")
    }
  }
}
```

But `initialize` never gets called.

[...]

PR #57, opened 2016-02-01
Enable reverse rotate

tbaranes:


IBAnimatable/RotationDesignable.swift
```diff
@@ -15,7 +15,7 @@ public protocol RotationDesignable {
 
 public extension RotationDesignable where Self: UIView {
   public func configRotate() {
-    if !rotate.isNaN && rotate > 0 && rotate < 360 {
+    if !rotate.isNaN && rotate > -360 && rotate < 360 {
```
JakeLin, line 18: Good idea, I will update README.md

JakeLin:
@tbaranes Thanks again and again. if you have time, can you have a look at https://github.com/JakeLin/IBAnimatable/blob/master/Documentation/Roadmap.md ? Any feedback is more than welcome.

tbaranes:
Great documentation, some points I think you could add:
- `Collection View and Collection View Cell` -> + Tableview and TableViewCell
- I won't stroke the slideOut due to the issue #46 

I love this both, that would be really awesome! These both features makes me think a little bit about [iosViews](https://github.com/Charimon/iosViews)

```
Importing Sketch to Swift playground
Importing Sketch to Storyboard? Maybe too hard.
```

I don't have that much idea for others features right now (the MacOS support would be awesome), but I will make a PR if I think about something, if I have sometimes, I will see to make some implementation of some of them :)

Really great library. Thanks for your hard work @JakeLin !

JakeLin:
@tbaranes Thanks for your comments and letting me know about iosViews project, I will look at it to find some ideas. 

The tasks of  "Importing Sketch to Storyboard" are bigger than what iosViews project can do if I understand correctly. Because we need to convert layers into semantic elements e.g. buttons.  There is not one-to-one mapping for flat layer to UIKit elements. And also Sketch doesn't support Auto Layout.  But we maybe able to start with something simple, Sketch can export CSS, or we maybe able to export `IBAnimatable` properties.

tbaranes:
Yes, iosViews is different of IBAnimatable goal (in fact, I never get it working, but it looks great in their README). They are only exporting some bezier path to draw lines... they aren't exporting any `UIKit` elements. As you said, a good start would be to export simple element (a button, a label...) with the right `IBanimatable` configuration.

Is there a place where you detailed more the big features of the roadmap (the small issues are already straightforward)? Some of them have already an issue related (#46 #50). 

By the way, you should add a link to the issue (when an issue is related) in your roadmap, that would be useful in order to check the current state, discussions...

JakeLin:
You are right, I will link the issues to the roadmap and another missing thing is the priority. Next thing I will focus on transitions to develop the architecture to support `IBInspectable` transition animations.

PR #59, opened 2016-02-05
Wave shape

tbaranes:
Not sure if you are interested in this kind of mask, but here it is: MaskWave
Examples of output: two waves (one up / one down) with different width / offset

screen shot 2016-02-05 at 20 40 09
screen shot 2016-02-05 at 20 41 06

In order to add this mask, I did some refactor in the `MaskDesignable`
- Put public all the maskStuff -> until we can pass parameters from the xib, we can at least make the mask more configurable in code
- Divide some codes in order to reuse it for the different shape (drawPath() / drawBorderPath())

Missing:
- Designable properties to manage the direction (up / down), width, and the offset of the wave. I didn't know if you have any thought how to handle this? If so, I will be happy to update the PR.
- Update README? Roadmap? Not sure I updated everything necessary

### DIFF
--- a/IBAnimatable/AnimatableLabel.swift
+++ b/IBAnimatable/AnimatableLabel.swift
@@ -5,7 +5,7 @@
 
 import UIKit
 
-@IBDesignable public class AnimatableLabel: UILabel, Animatable {
+@IBDesignable public class AnimatableLabel: UILabel, Animatable, RotationDesignable {
   
   // MARK: - Animatable
   @IBInspectable public var animationType: String?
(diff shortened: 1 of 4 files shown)


JakeLin:
Wave is useful to be used in the design like `Siri`. 

We had discussion on your concern before in #8 We haven't got the conclusion to use designable properties for each mask specific property or not. For example, if we have direction (up / down), width, and the offset for `Wave`, we may have `points` for `Star` and `Polygon`, the advantage is they are full customizable. And the disadvantage is we have some property they are not relative to some options of the enum. But that is what we are doing for Animations. `repeatCount` can only use for Shake, Pop, Morph, Squeeze, Flash, Wobble, Swing and Rotate.

And another thought I had is to use string to let the use set the parameter. for example, `Star` means 5-point star, and `Star(6)` string in IB means 6-point star. The trick can use in `Wave` as well like `Wave` is default shape, and `Wave(up,40,10)` for `directionUp=true`, `waveWidth=40.0` and `waveOffset=10`. What do you think about this workaround?

I like to merge the `Wave` PR and improve the interface in following discussions.

tbaranes:
I think you can already merge this PR. It's already usable, and a good starting point. The interfaces improvements will definitely come once we find the best solution to customise property following the mask type. I will think about some workaround to improve that, and join you directly on the thread #8.

In the current state, it can be a bit annoying if we need some more customisation, but we can always using it programatically until it's fully compatible with the interface builder.

JakeLin:
@tbaranes thanks again, let's discuss the inspectable properites on #8

PR #60, opened 2016-02-05
Fix slide out

tbaranes:
As discussed in #46, I implemented a first fix of the `slideOut` animations based on the `UIScreen` size. We should think about a better solution later. At least, it will always work until then :)


JakeLin:
Thanks. UISreen should fit most of cases.

PR #61, opened 2016-02-06
Rotate CW or CCW

marksands:
The rotation animation is great! But we modified it so that we can rotate clockwise or counterclockwise.

Is this something that you think is desirable?

### DIFF
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -174,8 +174,10 @@ public extension Animatable where Self: UIView {
       wobble(completion)
     case .Swing:
       swing(completion)
-    case .Rotate:
-      rotate(completion)
+    case .RotateCW:
+      rotate(clockwise: true, completion: completion)
+    case .RotateCCW:
+      rotate(clockwise: false, completion: completion)
     }
   }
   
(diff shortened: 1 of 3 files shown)


IBAnimatable/Animatable.swift
```diff
@@ -548,11 +550,11 @@ public extension Animatable where Self: UIView {
       }, completion: completion)
   }
 
-  public func rotate(completion: AnimatableCompletion? = nil) {
+  public func rotate(clockwise clockwise: Bool, completion: AnimatableCompletion? = nil) {
```
JakeLin: I prefer to have a default value to `true`, what do you think?

IBAnimatable/Animatable.swift
```diff
@@ -174,8 +174,10 @@ public extension Animatable where Self: UIView {
       wobble(completion)
     case .Swing:
       swing(completion)
-    case .Rotate:
-      rotate(completion)
+    case .RotateCW:
```
JakeLin: Can we keep `Rotate` as `RotateCW` and add `RotateCCW` for counterclockwise?

JakeLin:
@marksands Thanks for your PR, it is very useful to have anticlockwise/counterclockwise for `Rotate`. :+1:  Could you please have a look at my small comments and?

IBAnimatable/Animatable.swift
marksands: @JakeLin Yes, I agree

IBAnimatable/Animatable.swift
marksands: :+1:

marksands:
@JakeLin Look over the new changes and let me know what you think! :+1:

JakeLin:
@marksands perfect, thanks.👍

PR #62, opened 2016-02-06
Mask triangle

tbaranes:
screen shot 2016-02-06 at 18 03 48


JakeLin:
Well done @tbaranes 👍

lexrus:
I think we need a direction/angle property for this triangle so that we can draw icons for folders tree.
screen shot 2016-02-08 at 10 59 29 am

JakeLin:
@tbaranes @lexrus  can we use `rotate` property although it is not rendered in IB properly yet?

tbaranes:
Yes, I was thinking about rotating it directly. We can make it easier to use and add some parameters, but once again, we should find a happy ending for #8 before.

PR #63, opened 2016-02-06
Mask polygon

tbaranes:
screen shot 2016-02-06 at 18 11 22

### DIFF
--- a/IBAnimatable/MaskDesignable.swift
+++ b/IBAnimatable/MaskDesignable.swift
@@ -21,6 +21,8 @@ public extension MaskDesignable where Self: UIView {
         maskCircle()
       case .Star:
         maskStar()
+      case .Polygon:
+        maskPolygon()
       case .Triangle:
         maskTriangle()
       case .Wave:
@@ -37,6 +39,11 @@ public extension MaskDesignable where Self: UIView {
   
   public func maskCircle() {
     layer.cornerRadius = ceil(min(bounds.width, bounds.height))/2
+  }
+  
+  public func maskPolygon() {
+    let polygonPath = maskPolygonBezierPath(Int(arc4random() % 10 + 3))
+    drawPath(polygonPath)
   }
   
   // See https://www.weheartswift.com/bezier-paths-gesture-recognizers/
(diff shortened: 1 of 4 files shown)


IBAnimatable/MaskDesignable.swift
```diff
@@ -37,6 +39,11 @@ public extension MaskDesignable where Self: UIView {
     layer.cornerRadius = ceil(min(bounds.width, bounds.height))/2
   }
   
+  public func maskPolygon() {
+    let polygonPath = maskPolygonBezierPath(Int(arc4random() % 10 + 3))
```
JakeLin: Shall we use `arc4random_uniform` here?

IBAnimatable/MaskDesignable.swift
```diff
@@ -97,6 +104,24 @@ public extension MaskDesignable where Self: UIView {
     return CGPoint(x: radius * cos(angle) + offset.x, y: radius * sin(angle) + offset.y)
   }
   
+  // MARK: Polygon
+  
+  private func maskPolygonBezierPath(degree: Int) -> UIBezierPath {
```
JakeLin, line 116: can it be `public`?

JakeLin:
Thanks again, can you please have a look at my small comments when you have time? We do need a workaround to support additional parameters for `Polygon` as well.

IBAnimatable/MaskDesignable.swift
tbaranes, line 116: I don't think it should be. That method is just creating the path before drawing it... why do you want it public?

IBAnimatable/MaskDesignable.swift
tbaranes: Definitely not, but until we can customise the parameters, it's the an easy way to create / test polygon.

IBAnimatable/MaskDesignable.swift
JakeLin, line 116: You are right.

IBAnimatable/MaskDesignable.swift
JakeLin: I think `arc4random_uniform` is a little bit better in this case. It would be `let polygonPath = maskPolygonBezierPath(Int(arc4random_uniform(10) + 3))`, please have a look at http://nshipster.com/random/ 

Will create a doc/an issue for "customise the parameters" and let us vote.

tbaranes: Just pushed the change

JakeLin:
I think I have changed the Roadmap.md, if you don't mind can you rebase it for me?
something like

`git pull --rebase upstream master`
`git mergetool` // if there is conflict
// use `opendiff` or whatever to check the conflict and manual merge it if needed
`git rebase --contine`
`git push -f origin mask_polygon`

Thanks. <del>we</del> You are great👍(Was a typo😅)

tbaranes:
I think, we are good! Could you just check the roadmap to be sure I didn't fail the merge?
The most of the conflicts was due to my both PR (triangle vs Polygon 😅)

Thanks!

JakeLin:
@tbaranes  great. 👍

PR #64, opened 2016-02-06
SideImageDesignable's right image

tbaranes:
I did some refactoring of the `SideImageDesignable` in order to reuse the code to handle the right image. 
Also, a future improvement could be make designable the `viewMode`, don't you think? Related  #8


JakeLin:
:+1:  Thanks again.

Issue #24, opened 2016-01-01
Default gradient colors


JakeLin:
Predefined gradient colors from http://uigradients.com/

tbaranes:
I like this idea, but I'm asking myself how that will render when used. I suppose it will be a big enum with each gradient, then `DesignableGradient` will have a designable `defaultGradient`, but in order to use it, we will have to enter a string in the `defaultGradient` field? Some of them are understandable, but not all... don't you think?

JakeLin:
@tbaranes Sorry for the late reply, yesterday, I brought my son to see the lion dancing. 

I have thought through that before I created this card. It will automatically generate a enum `GradientType` from https://github.com/Ghosh/uiGradients/blob/master/gradients.json vis JavaScript or Swift script. Similar to what I did for 
https://github.com/JakeLin/SwiftWeather/blob/master/SwiftWeather/WeatherIcon.swift . Then we can easily update it when the json has been updated.

Do you think `presetGradient` or `predefinedGradient` is a better name than `defaultGradient`?  Because of IB's limitation, we have to enter a string like `SunnyDays`, `GreenBeach` refer to gradients.json. To let the user understands the more, we need to put a link in our API document to http://uigradients.com/ then they can preview all of them and copy the gradient name (string for enum rawValue).

Please let me know if you have any questions. Thanks.

tbaranes:
Got it. Interesting implementation. 

I like `predefinedGradient`, but all of them seems good names for this feature.

lexrus:
We can use Ruby:

``` ruby
require 'open-uri'
require 'JSON'

JSON_URL = 'https://raw.githubusercontent.com/Ghosh/uiGradients/master/gradients.json'

def hexToRgb(hex)
  if m = /#(..)(..)(..)/.match(hex)
    return "red: #{m[1].hex} green: #{m[2].hex} blue: #{m[3].hex}"
  end
end

gradients = JSON.load(open(JSON_URL))

puts "private let predefinedGradients: [String: [String]] = ["

gradients.each do |g|
  name = g['name']
  from, to = g['colors']
  puts "  \"#{name}\": [\"#{from}\", \"#{to}\"],"
end

puts "]"
```

It outputs:

``` swift
private let predefinedGradients: [String: [String]] = [
  "Sunny Days": ["#EDE574", "#E1F5C4"],
  "Green Beach": ["#02AAB0", "#00CDAC"],
[...]

PR #79, opened 2016-02-09
Predefined gradients

tbaranes:
Here are the predefined gradients (#24).
It's working perfectly well in the current state, but some unanswered questions (improvements?) are still persisting:
- The script that generates the enum / switch is just printing out the swift code. It's easy to use: just copy / paste them at the right place. I prefer this way since it's usable and more secure (we choose what we do), but maybe it would be better that the script erase the override the last output (directly in the project)? 
- `DesignableGradientView`: I didn't get why that class wasn't implemented the `GradientDesignable` directly instead of rewriting the process. I saw that `GradientDesignable` wasn't implemented the `startPoint` properly (the property is not used). Is there any reasons for these implementations or is it a bug? Whatever is the reason, it will probably need another PR to take of this

I'm waiting your feedback in order to improve all that stuff :)

PS: For the color extension, I just pick a working code on the net. If you have any preference for the implementation, I'm open to any suggestions!


JakeLin:
👍, please merge it if you feel ready.

I like the idea to generate the code and paste it to the file manually. 

`DesignableGradientView` originally is a temporary view to wrap a layer within it. Will fix it. Add an issue against that problem. https://github.com/JakeLin/IBAnimatable/issues/81

JakeLin:
@tbaranes please have a look, I have made some small modifications to your PR.

tbaranes:
Good idea to make `UITableViewCell` and `AnimatableView` conform `GradientDesignable` :+1:

JakeLin:
Shall we have a release/tag for that? That is a very useful feature. I am going to demonstrate this feature in Melbourne Cocoaheads

tbaranes:
Awesome! Yes, we can make a new release. A lot of new features has been added since the last one. Do you want to create a CHANGELOG or directly explain what did change directly in the release desc?

JakeLin:
CHANGELOG should be better and easier to maintain the history.

PR #80, opened 2016-02-09
MoveTo and MoveBy animations

tbaranes:
#74


JakeLin:
Thanks for developing this feature. I may upset you because I have second thought, sorry because I didn't think through carefully.

In SpriteKit, there are some nice API like `moveBy:duration:` and `moveTo:duration:` https://developer.apple.com/library/prerelease/ios/documentation/SpriteKit/Reference/SKAction_Ref/index.html#//apple_ref/occ/clm/SKAction/moveTo:duration: 

There are two types of move, `moveBy` and `moveTo`, in our case, we can use `CGPoint` for the IBInspectable property to have both x and y like 

```
  var shadowOffset: CGPoint { get set }
```

Since we use `NaN`, we can only move the set one (x or y) by checking the value.

tbaranes:
Didn't know about these API in sprite kit, I will check this out soon :+1:

tbaranes:
@JakeLin Just did a refactor to replace the both `CGFloat` by a `CGPoint`. 
For the `moveTo` / `moveBy` from SpriteKit, you need a `SKSpriteNode` to use them. How could we use it in our context? You can also directly make the change in the branch, it's to you.

JakeLin:
@tbaranes thanks, please leave it to me. I will keep going with this branch. We don't use `SKSpriteNode` at all, we just get the idea from SpriteKit to create our API interface. `moveTo` and `moveBy` will be the methods and in the enum `AnimationType`. I need to prepare a talk for Melbourne Cocoaheads, probably, will do it on weekend.

tbaranes:
Oh, just understood what did you mean by looking at SpriteKit. Anyway, I'm gonna let you ending this PR. Let me know if you want some help.

Have fun with your talk! :tada:

JakeLin:
I have pushed some code to the branch, but please don't merge, there is some problem to change the frame with AutoLayout. Will ask for review once I found a better solution. 

Currently in this branch, we have `MoveTo` animation as the interface, later we can add `MoveBy`, both will take `position` property to move the element.

After both above are done, we can even add `MoveFrom` and `MoveFromBy` animations. Those interface are similar to Sprite Kit's.

JakeLin:
@tbaranes I have updated the animations to `MoveTo` and `MoveBy`. And use `x` and `y` instead of `position` because of a bug of IB, when we set a value in `position` e.g. `position.x` to 20, IB will set `position.y` to 0 automatically. 

When using `MoveTo`, `x` and `y` are absolute position and when using `MoveBy`, `x` and `y` are the offset to move.

Reviewer @tbaranes , please have a look. Thanks. Let me know if you have any questions or comments.

tbaranes:
Nice work, I like the idea of `moveTo` / `moveBy`.

Let's just fix the star before launching 1.2, but we can already start working on the architecture of the 2.0 :tada:

JakeLin:
Yes, I will update the last bit of `MaskDesignable` in README.md and realse 1.2 tomorrow morning. Let's have some fun with transitions. 👯

PR #82, opened 2016-02-10
Fix a bug: when `leftPadding` is not set, the app crashes.

JakeLin:
When `leftPadding` is not set (NaN), the app crashes when we open the create account scene.


tbaranes:
Nice catch!

PR #83, opened 2016-02-10
First CHANGELOG

tbaranes:
Following the discussion in #79, I came up with a first CHANGELOG suggestion. 
I pick a basic format, but we should probably update it and choose a better format, that's including:
- Title format for each version
- Subpart by versions: "API breaking changes", "Enhancements", "Bugfixes" or just a linear list
- Each "news": link to the issue / PR ? Author? Just a short desc? ...
- ...

Feel free to update the file directly in the branch to improve it until we pick the final one.

I filled the "next" block with the last PR / issues closed. Feel free to complete the list if I miss some history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,22 @@
 # Next
 
-- New animations: Rotate, RotateCCW - [#51](https://github.com/JakeLin/IBAnimatable/issues/51)
+### API breaking changes
+
+### Enhancements
+
+- New animations: Rotate, RotateCCW [#51](https://github.com/JakeLin/IBAnimatable/issues/51)
 - New masks: Star, Triangle, Polygon, Wave
 - SidebarImage now support right images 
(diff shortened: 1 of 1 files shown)


JakeLin:
👍Great, I like it. I think that's simple and well documented. I don't mind to put "Enhancements", "Bugfixes" all together since the user should figure them out very quickly. But for "API breaking changes", I would create a section for it to let the user know the important changes.

tbaranes:
I don't really care about keeping  "Enhancements", "Bugfixes" together, but adding "API breaking changes" doesn't make it look nice, here two solutions:
- Put separately  "Enhancements" and "Bugfixes"
- Find a nice word that represent these both points

Here a changeling conforming to the first option.
Do we have any breaking changes for the next release?

JakeLin:
We don't have any breaking changes for next release.  Your original one was very good, now is even better, let's starts with this format. Thanks again, great work.

Issue #81, opened 2016-02-10
DesignableGradientView should conform to GradientDesignable


JakeLin:
`DesignableGradientView` was a temporary view to wrap a layer. need to conform to  `GradientDesignable` protocol.

tbaranes:
`GradientDesignable` should also implement `startPoint` process which is implemented in `DesignableGradientView`

JakeLin:
Yes, move it in `DesignableGradientView`

tbaranes:
It still needs some improvements, but DesignableGradientView conforms to `GradientDesignable`.
Also, the `startPoint` is now working when conforming to `GradientDesignable`.

PR #84, opened 2016-02-10
Various fixes on designable gradient view

tbaranes:
#81

### DIFF
--- a/IBAnimatable/AnimatableImageView.swift
+++ b/IBAnimatable/AnimatableImageView.swift
@@ -113,21 +113,12 @@ import UIKit
     autoRunAnimation()
   }
   
-  // MARK: - CALayer
-  public override class func layerClass() -> AnyClass {
-    return CAGradientLayer.self
-  }
-  
-  var gradientLayer: CAGradientLayer {
-    return layer as! CAGradientLayer
-  }
-  
   // MARK: - Private
   private func configInspectableProperties() {
     configAnimatableProperties()
     configBorder()
     configTintedColor()
     configBlurEffectStyle()
-    configGradientWithLayer(gradientLayer)
+    configGradient()
   }
 }
--- a/IBAnimatable/AnimatableImageView.swift
+++ b/IBAnimatable/AnimatableImageView.swift
@@ -110,8 +110,16 @@ import UIKit
   
   public override func layoutSubviews() {
     super.layoutSubviews()
-    
     autoRunAnimation()
+  }
+  
+  // MARK: - CALayer
+  public override class func layerClass() -> AnyClass {
+    return CAGradientLayer.self
+  }
+  
+  var gradientLayer: CAGradientLayer {
+    return layer as! CAGradientLayer
   }
   
   // MARK: - Private
@@ -120,6 +128,6 @@ import UIKit
     configBorder()
     configTintedColor()
     configBlurEffectStyle()
-    configGradient()
+    configGradientWithLayer(gradientLayer)
   }
 }
--- a/IBAnimatable/GradientDesignable.swift
+++ b/IBAnimatable/GradientDesignable.swift
@@ -14,22 +14,23 @@ public protocol GradientDesignable {
 
 public extension GradientDesignable where Self: UIView {
 
-  public func configGradientWithLayer(gradientLayer: CAGradientLayer) {
+  public func configGradient() {
     let predefinedGradient = configPredefinedGradient()
     if let unwrappedStartColor = startColor, unwrappedEndColor = endColor {
-        configGradientWithStartColor(unwrappedStartColor, endColor: unwrappedEndColor, gradientLayer: gradientLayer)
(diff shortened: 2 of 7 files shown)


IBAnimatable/GradientDesignable.swift
```diff
@@ -14,34 +14,49 @@ public protocol GradientDesignable {
...
+      gradientLayer.endPoint = CGPoint(x: 0.5, y: 0)
+    case .BottomLeft:
+      gradientLayer.startPoint = CGPoint(x: 0, y: 1)
+      gradientLayer.endPoint = CGPoint(x: 1, y: 0)
+    case .Left:
+      gradientLayer.startPoint = CGPoint(x: 0, y: 0.5)
+      gradientLayer.endPoint = CGPoint(x: 1, y: 0.5)
+    case .TopLeft:
+      gradientLayer.startPoint = CGPoint(x: 0, y: 0)
+      gradientLayer.endPoint = CGPoint(x: 1, y: 1)
```
JakeLin, line 59: Good idea to put them in protocol extension.

IBAnimatable/AnimatableImageView.swift
```diff
@@ -110,16 +110,24 @@ import UIKit
...
-    
     autoRunAnimation()
   }
   
+  // MARK: - CALayer
+  public override class func layerClass() -> AnyClass {
+    return CAGradientLayer.self
+  }
+  
+  var gradientLayer: CAGradientLayer {
```
JakeLin: I am not 100% on the idea to create a layer property for each `subclass`, can we encapsulate it in `DesignableGradientView`, `DesignableGradientView` is a subview of other UI elements conform to `GradientDesignable`.

There are some reason I can think of for encapsulating `gradientLayer` within `DesignableGradientView`
1. Better Autolayout support, CALayer doesn't support Autolayout directly, if we rotate the device, we may need to manipulate the layer again to fit the new orientation.
2. The subclasses e.g. `AnimatableImageView` do not need to have any knowledge of gradient implementation. According to [Open/closed principle](https://en.wikipedia.org/wiki/Open/closed_principle) 
3. Keep all interface consistent like `configXXXXX()` Although we have created a lot of UI elements start with `AnimatableXXXX`, but another goal of the project is to provide protocol extensions to create custom UI elements easily, e.g. `MaterialButton` or `BootstrapButton`. To achieve that we need to make all API as clean as possible. 

I feel the challenges when we work on a library, we need think about how to implement and at the same time, we also consider how the others use the library. But we get better architecture skills on the way while doing that. Let's have some fun with those challenges. Please let me know if you have any questions.

tbaranes: I completely agree with all the points. At the beginning, I was looking for a better implementation which could answer to that points.

However, I'm not sure I get what do you mean by making `DesignableGradientView` encapsulating all the others subviews. Something like creating a `DesignableGradientView` and adding it at the subviews of the view that conform to `GradientDesignable`? If so, I don't really like that solution since it will make a lot of boilerplate code in each subviews. I really like the fact that it's really easy to make  a new view conform a protocol without that kind of code.

For the moment, I removed all the layer stuff from the views that conform `GradientDesignable`, and moved it directly in the protocol extension. It's clearly not enough since it won't handle the rotation... but I'm gonna think how to improve that.

tbaranes:
I completely agree with all your points. At the beginning, I was looking for a better implementation which could answer that points.

However, I'm not sure I get what do you mean by making DesignableGradientView encapsulating all the others subviews. Something like creating a DesignableGradientView and adding it at the subviews? If so, I don't really like that solution since it will make a lot of boilerplate code in each subviews. I really like the fact that it's really easy to make a new view conform a protocol without that kind of code.

For the moment, I removed all the layer stuff from the views that conform GradientDesignable, and moved it directly in the protocol extension. It's clearly not enough since it won't handle the rotation... but I'm gonna think how to improve that.

JakeLin:
@tbaranes original, I used `DesignableGradientView` which encapsulates all logic, then all other subviews can use the following code to add a subview

```
gradientView.autoresizingMask = [.FlexibleWidth, .FlexibleHeight]
self.insertSubview(gradientView, atIndex: 0)
gradientView.configGradient()
```

Then it can support Autolayout, think about we have a background view configures with gradient color. Then we rotate the phone from landscape to portrait. a subview can automatically re-size but a sublayer can't.

I like what you did to move all logic into `GradientDesignable` protocol which can be reused for any class conform to it. I merge it first and we can think about how to improve it. I even want to remove `DesignableGradientView` it originally was a temporary view and should not be `public`, the user should always use `AnimatableView` which supports gradient colors. Please let me know if you have ideas.

tbaranes:
I agree to remove `DesignableGradientView`, it's was useful until now since it was the only view handling the gradient properly. Now that all the classes conforming to `DesignableGradient` are (almost) working, it's just duplicating `AnimatableView`...

Let's create an issue to continue this discussion to keep a history!

PR #90, opened 2016-02-13
Swift Package Manager support

### LINKED ISSUE #5: Swift package manager support
Support Swift package manager to install `IBAnimatable`.

tbaranes:
I also updated the "installation" part in the README to have a commun documentation between the three third party. Not sure which one you wanted at the beginning, so I just keep the CocoaPods one.

Should close #5


JakeLin:
Well done.

lexrus:
@tbaranes @JakeLin 
As I know, we can not submit an iOS App built with beta toolchain. In other words, we can't build IBAnimatable module with the current stable Xcode 7.2.1.

JakeLin:
I didn't know that, but I have merged it into master, how about leaving a comment in README.md to let the user know it can only work for Xcode 7.3 and need to wait for Xcode 7.3 GM to submit to App Store?

tbaranes:
This issue is happening only when build with the SPM, isn't it? If so, I think people who are already using this dependency manager are already aware. Nevertheless, I agree with Jake to leave a comment in the README.

lexrus:
Yes, that's right. They must knew it. :smile:
We can just wait for Xcode 7.3 without touching the README, it's around the corner.

tbaranes:
By the way, @JakeLin, to completely support SPM, the tag of each release must contains only the version number: "1", "1.1", "1.1.1"... SPM doesn't accept "V".

I don't think anyone we'll try to use `IBAnimatable` with SPM right now, so, it's not a big deal to not be fully compatible before the next release.

JakeLin:
Thanks @tbaranes ,will update all release tags without "V", learning a lot from you guys every day.

lexrus:
Have you guys tried integrate IBAnimatable with SPM in Xcode 7.3?

JakeLin:
@lexrus not yet, I use CocoaPods intead. But will make an example in https://github.com/JakeLin/IBAnimatable-PackageManagementExamples. Have you got any experience on that?

lexrus:
Looks like there is no `swift build` in current toolchain embed in Xcode 7.3.1

```
error: unable to invoke subcommand: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build (No such file or directory)
```

Thus, we can not build IBAnimatable as a Swift module in an iOS project yet unless you switch to the 3.0 toolchain. But you can't publish the App with beta toolchain right?

JakeLin:
@lexrus yes, I was working on another project https://github.com/JakeLin/Underscore and found out, for Swift 2.2, `swift build` is only in development snapshots. For release packages, it was only in "January 11, 2016" release and before, but not in the other releases after that. I can't find the information why Apple did that. We may update README.md to tell the user or I think if the user uses SPM, they may know the constraint. I don't see many users use SPM for iOS or Mac, most of them are using it for Linux.

PR #91, opened 2016-02-13
Fixed gradient designable rotation

### LINKED ISSUE #86: Improving GradientDesignable implementation
History that bring this issue: #84  / #81 
Historically, `GradientDesignable` wasn't working even if a subview was conforming the protocol. In order to fix this, we moved all the gradient's intelligence directly in the protocol, by creating creating a sublayer directly in the protocol. Using this new implementation, we can make the API nicer and removed useless views like `DesignableGradientView` (#87)

That solution is working well for basic usage. We still need to improve it in order to support rotation... and if possible, let `GradientDesignable` handle everything itself in order to keep the protocol reusable everywhere, without breaking changes.

Any suggestions?

tbaranes:
Should closed #86 

Here an implementation which seems a good compromise. I didn't find any way to handle this issue without a custom subview. So in order to keep everything easy to use, I ended up by created a private custom view only used in `GradientDesignable`.

This way, the layer handles well the rotation.. didn't find any bugs related to this implementation.

### DIFF
--- a/IBAnimatable/GradientDesignable.swift
+++ b/IBAnimatable/GradientDesignable.swift
@@ -58,12 +58,14 @@ public extension GradientDesignable where Self: UIView {
       gradientLayer.startPoint = CGPoint(x: 0, y: 0)
       gradientLayer.endPoint = CGPoint(x: 1, y: 1)
     }
+    
+    let gradientView = GradientView(frame: self.bounds, layer: gradientLayer)
+    self.insertSubview(gradientView, atIndex: 0)
   }
   
   private func createGradientLayer() -> CAGradientLayer {
     let gradientLayer: CAGradientLayer = CAGradientLayer()
     gradientLayer.frame = self.bounds
-    self.layer.insertSublayer(gradientLayer, atIndex: 0)
     return gradientLayer
   }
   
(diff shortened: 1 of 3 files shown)


IBAnimatable/GradientDesignable.swift
```diff
@@ -354,3 +356,23 @@ public extension GradientDesignable where Self: UIView {
     }
   }
 }
+
+private class GradientView: UIView {
+
+  // MARK: - Life cycle
+  
+  init(frame: CGRect, layer: CAGradientLayer) {
```
JakeLin, line 364: Do we need this `init` with layer method here? can we just remove both `init` and add

```
gradientView.autoresizingMask = [.FlexibleWidth, .FlexibleHeight]
gradientView.layer.insertSublayer(gradientLayer, atIndex: 0)
```

between line 62 and 63?

JakeLin:
:+1: except for a small commit above. You implementation was good, it is up to you to change or not.

IBAnimatable/GradientDesignable.swift
tbaranes, line 364: I like the idea that the protocol extension doesn't have the knowledge how `GradientView` works. I mean, it has to know how create a gradient and customise the layer... but the view part is not really related to the protocol. That's the only reason I created a custom init, but it could be also a `configGradientViewWithLayer(layer: CAGradientLayer)`, that may conform more the rest of the project.

What do you think?

JakeLin, line 364: Cool, it is `private` and not used in any where else. Good one.

tbaranes, line 364: Which solution is the good one? :laughing:

Issue #76, opened 2016-02-08
Custom properties for enum value


JakeLin:
TBD

Some discussions 
https://github.com/JakeLin/IBAnimatable/issues/8
https://github.com/JakeLin/IBAnimatable/pull/59

tbaranes:
The current ideas suggested (which could be working) in order to make that thread going further:
1. `Wave(up,40,10)`: MaskType(param1,param2,param3)
2. More AnimatableView's children to add custom designable properties following the shape
3. Other suggestions?

I prefer the first solution, but it could be maybe too hard to use, or improving a bit the syntax to be a bit more swift: `Wave(direction:up,width:40,offset:10)` (still annoying since the syntax is really strict)

tbaranes:
Another approach:

``` swift
public enum Mask {
    case Circle()
    case Star(branches: Int)
    case Wave(up: Bool, width: Int, offset: Int)
}

Mask.Circle()
Mask.Star(branches: 5)
Mask.Wave(up: true, width: 40, offset: 40)
```

I still missing a nice way to make this working with the interface builder, but I find this approach more swifty.

JakeLin:
@tbaranes @lexrus came up similar ideas before to use Swiftified enum. But since IB doesn't support enum yet, it can't fix our problem.

I prefer option one, and it can at least move us forward. If we can support option one, we can subclass `AnimatableView` to support option two if needed. 

To get the benefit of Swiftified enum, we can have 

```
public enum Mask {
    case Circle()
    case Star(branches: Int)
    case Wave(up: Bool, width: Int, offset: Int)
}
```

in code then map the `String` value e.g. `Star(6)` to `Star(branches:6)` for IB. And fall back to default value `Star(branches:5)` if the user put it wrongly like `Star5`, `Star(12` or Star2)

PR #92, opened 2016-02-13
Customising IB mask properties

### LINKED ISSUE #8: More MaskDesignable

tbaranes:
Here a first version of mask customisation from IB following the related threads. Should closed #8 #50 #76 

The following implementation has been done for:
-  `StarMask`, IB usage: `Star(6)`, `Star(12)`...
-  `WaveMask`, IB usage: `Wave(up,50,100)`

If one parameters is missing, or miswrite, we will fall back to the basic implementation with default values.
I also allowed some miswrite errors: `Star(6` is accepted, as well as `Wave(up, 50, 100)`

Finally, I didn't implement the swifty enum. I concluded that we will still need a string enum in order to avoid hard coding strings or multiple constants for each mask. So, even if it's more swifty to have the following one, it will just create duplicates code.

```
public enum Mask {
    case Circle()
    case Star(branches: Int)
    case Wave(up: Bool, width: Int, offset: Int)
}
```

Missing:
- README update: don't really know how to explain this feature... so I prefer let you handle this (if you don't mind)

I'm waiting your feedbacks on that :)


JakeLin:
@tbaranes , thanks for the PR, I found out it didn't work before because `MaskType(rawValue: unwrappedMaskType)` will return `nil` if we specify the parameter like `Star(6)`, I have fixed it by adding additional check like 

```
if unwrappedMaskType.hasPrefix(MaskType.Star.rawValue) {
        maskStarFromString(unwrappedMaskType)
      }
```

I have tested them, please have a look at my changes. Will update README.md later once we get that in master.

tbaranes:
Thanks for the fix Jake, I miss that point after refactoring the PR... my bad.
I'm gonna merge the PR once travis has finished its builds.

Issue #71, opened 2016-02-08
Predefined flat colors 


JakeLin:
Flat colors like https://flatuicolors.com/ or more predefined colors.

PR #93, opened 2016-02-13
Predefined colors

tbaranes:
Issue #71 

I wasn't sure where `predefinedColor` should be. I finally put it in `FillDesignable` since it seems to be the best place. It works well, and as `predefinedGradient`, it will set its color only if `fillColor` isn't set.

### DIFF
--- a/IBAnimatable/FillDesignable.swift
+++ b/IBAnimatable/FillDesignable.swift
@@ -7,6 +7,7 @@ import UIKit
 
 public protocol FillDesignable {
   var fillColor: UIColor? { get set }
+  var predefinedColor: String? { get set }
   var opacity: CGFloat { get set }
 }
 
@@ -14,6 +15,8 @@ public extension FillDesignable where Self: UIView {
   public func configFillColor() {
     if let unwrappedFillColor = fillColor {
       backgroundColor = unwrappedFillColor
+    } else if let unwrappedPredefinedColor = predefinedColorFromString(predefinedColor) {
+      backgroundColor = unwrappedPredefinedColor
     }
   }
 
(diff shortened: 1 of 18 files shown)


JakeLin:
Very good, later on, we can add on Material Colorado when we work on v4 material design elements.

IBAnimatable/FillDesignable.swift
```diff
@@ -34,6 +37,62 @@ public extension FillDesignable where Self: UITableViewCell {
...
+
+private extension FillDesignable {
+  
+  func predefinedColorFromString(predefinedColor: String?) -> UIColor? {
+    guard let unwrappedColorTypeString = predefinedColor, colorType = ColorType(rawValue: unwrappedColorTypeString) else {
+      return nil
+    }
+    
+    switch colorType {
+    case .Emerland:
```
JakeLin: Do you think we should prefix them as FlatEmerland to support MaterialRed later on ?

tbaranes: I don't think so, at least, not until we have to. These flat colors are known named, we risk to confuse their usage if we change the name.

JakeLin: @tbaranes Sorry, I am not sure are we on the same page? I mean for all flat colors, can we add a prefix for all of them like `FlatEmerland` and `FlatPomegranate` instead of `Emerland` and `Pomegranate`. 

The advantage of add `Flat` as prefix, the user knows they are from flat colors (https://gist.githubusercontent.com/tkrugg/a577c32e93eecc6c7991/raw/f7866132e26f9bffcda9caf13dced55e4a99e145/flatuicolors.json). And when we add other set of colors, e.g. Material Design color (https://www.google.com/design/spec/style/color.html#color-color-palette), then all of them will prefix as `Material`, e.g.  `MaterialRed500` and  `MaterialRed50`, those names may not be the final one, just an example to describe the usage. Another benefit is when the user uses the enum in code, when they can start typing `.Flat` and Xcode will auto complete the rest for them.

But there are disadvantage like getting the name longer than it should be.

Please let me know what you think.

tbaranes: I understand your point. I like the idea to have the completion when tapping .Flat. I just updated the PR. For the name, it's predefined colors, so I don't think it's a big issue to have longer name.

tbaranes:
I like the idea to have the completion when tapping .Flat. I just updated the PR to add the prefix "Flat" for each color. The name is a bit longer, but I don't think it's a big deal since it's just _Predefined flat colors_

tbaranes:
Since, the issue #94  is related to this one, I updated the implementation of all the `Animatable...` `FillDesignable` in order to have only one way to conform the protocol.

JakeLin:
Well done @tbaranes , just merged it into master.

PR #99, opened 2016-02-14
Fixed border designable and mask border

tbaranes:
We need to delete and create a new layer when the frame isn't the same as in the xib, or when we rotate. Otherwise, the border line won't be drawn with the right frame.

I also used this PR to remove the gradientSubview when creating a new one to avoid a stack of useless `GradientView`.

EDIT: This PR now handles also the follow bugs / improvements:
- Making `BorderDesignable` testable directly in playground
- Making `BorderDesignable` compatible with `MaskDesignable` (#101)


lexrus:
Tom, do you have any idea how can we draw the border properly when maskType is Star or Triangle?

tbaranes:
Oh, that's a good question. Should we really handle maskType borders using `BorderDesignable`? I mean, draw the bottom border of a star is a bit confusing. A simple way (I imagine) would be to add a property `maskBorderColor` to `MaskDesignable`. 

We should probably open an issue to discuss this. We will probably think about lot of solution, but not a lot will be a nice solution. What do you think?

lexrus:
Well, if there isn't a nice solution yet, shall we mention this the README or docs? Someone may try to draw rating stars with IBAnimatable, and finally disappointed.
screen shot 2016-02-14 at 10 08 10 pm

tbaranes:
@lexrus I just push a fixed for the mask border (#101). What do you think of this approach?

One bad points which could be improved later:
- We have to call `configBorder` after a `configMask` if the view implements the both protocols. 

screen shot 2016-02-14 at 17 33 12

I also update the implementation of `BorderDesignable` in each view in order to make it testable from playground. I don't see any bad points for the runtime, did you?

lexrus:
Awesome.
Can we use KVO in IB?

IBAnimatable/AnimatableButton.swift
```diff
@@ -34,9 +34,23 @@ import UIKit
...
+    didSet {
+      configBorder()
+    }
+  }
+  
+  @IBInspectable public var borderSide: String? {
+    didSet {
+      configBorder()
+    }
+  }
```
JakeLin, line 53: If I remember correctly, I started with that and wanted to avoid calling `configBorder` multiple times, and move them out in `configInspectableProperties`. But should be that way for changing those properties after `awakeFromNib`.

JakeLin:
@tbaranes :+1: Like it, Well done. calling `configBorder` after a `configMask` is OK for me, when I worked on the library, I tried to force the subclass to call some methods like something `required` override, but must call instead of override.  But I didn't find a solution, I think there is a guideline to subclasses to conform to those protocols. they need to set to default value and call `configXXXX()` in correct timing.

@lexrus I am not sure what do you want from KVO here? let `configBorder` automatically call after `configMask`

IBAnimatable/AnimatableButton.swift
tbaranes, line 53: I think that too, I really thought it was a bug when I first tried in playground. I suppose an external user will have the same effect. Also, it probably don't call that much, don't you think?

tbaranes:
Happy that you like it guys. If it's ok for  you, we can already merge this and think further how to improve the global system later if necessary :tada:

JakeLin:
Well done. Getting very closer to 1.2

Issue #96, opened 2016-02-14
Fix layout issues of Main.storyboard for iPad


lexrus:
<img width="624" alt="screen shot 2016-02-14 at 4 22 52 pm" src="https://cloud.githubusercontent.com/assets/219689/13032697/d8d45196-d337-11e5-9227-63b7da19bf95.png">

tbaranes:
Closing this until we really wants to create a layout specific for iPad.

PR #100, opened 2016-02-14
Fixed iPad layout

tbaranes:
That PR fixed the issue #96. Now the layout on iPad won't be anymore buggy (backgrounds fits...)

Nevertheless, the iPad's layout is still not really nice. Some easy customisation could improve that, for example: more static cells in the tableview, making the subviews always centered. That's not a priority, but could be another issue "nice to have".


lexrus:
Nice!

JakeLin:
Well done.

PR #107, opened 2016-02-16
Fix potential crash while using DesignableGradient

tbaranes:
I got a crash in my project just after updated the pod. When I read the code, it makes sense that there an error. The weird part is that I retest the gradient on IBAnimatable's simulator, and it's working perfectly.

So I don't really know why is it working in the sample project (simulator cache? Xcode derived data? ...), but here a fix to be sure that won't happen to anyone else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ TBD
 TBD
 
 #### Bugfixes
-TBD
+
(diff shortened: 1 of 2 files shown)


JakeLin:
No idea why the demo app is OK as it should not be. :+1:  for this PR. Thanks.

PR #110, opened 2016-02-24
AnimatableNavigationController supports custom transition animations

JakeLin:
There is the first PR to support custom transition animations. It has a flexible architecture to support more custom transitions.

`AnimatableNavigationController` is a subclass of `UINavigationController`, it has two IBInspectable properties:
`transitionAnimationType` for `TransitionAnimationType` and `transitionDuration` for duration

Currently, we have four custom animations: `CubeFromLeft`, `CubeFromRight`, `CubeFromTop` and `CubeFromBottom`. Will add a lot more animations soon.

To see those supported custom transition animations in the demo App. Tap on "Forgot Password" button in the login screen.

relate to https://github.com/JakeLin/IBAnimatable/issues/19


tbaranes:
Awesome. It looks really nice and easy to use! Good job  :+1:

Issue #19, opened 2015-12-29
Design transition animations in Interface Builder


JakeLin:
Not only system built-in transitions, more custom transitions support.
- [x] `AnimatableNavigationController` to support custom transition animations for Push and Pop
- [x] `AnimatableViewController` to support custom transition animations for Present and Dismiss
- [x] <del>`AnimatableTableViewController` to support custom transition animations for Present and Dismiss</del>
- [x] `Present***Segue` to support custom transition animations for Present and Dismiss
- [x] `***InteractiveAnimator` to support gesture to Pop or Dismiss VC. e.g. Pan, ScreenEdgePan and Pinch.
- [x] <del>More custom animations.</del>

tbaranes:
Do you have any preferences about which transitions should be implemented or examples of the results expected?

lexrus:
Some concepts:
https://github.com/entotsu/TKSubmitTransition
https://github.com/NewAmsterdamLabs/ZOZolaZoomTransition
https://github.com/AlvaroFranco/AFBlurSegue

I think image zooming segue is very useful in most Apps. And it's handy to dismiss it with swap like Jake said in #20 .

tbaranes:
Other list of concepts: http://transitiontreasury.com

lexrus:
https://dribbble.com/shots/2131325-Bike
https://dribbble.com/shots/1216761-Liquid-Animation

JakeLin:
First thing, we need to setup an architecture for that. I got three ideas to achieve that. 
1. An animator conforms to `UIViewControllerAnimatedTransitioning` and drag `NSObject` into a `UINavigationController` on IB (on top of the Controller), set that `NSObject` to that animator, set that navigation controller's delegate to it. Then all transitions within that navigation controller will use the animation specified in that animator.
2. Add two properties into `AnimatableViewController` like `transitOutAnimation` and `transitInAnimation`, let `AnimatableViewController` conform to `UIViewControllerAnimatedTransitioning` and pick up the animation according to those two properties.
3. Subclass `UISegue` for different transition animations. e.g. `FadeInOutPushSegue` for fade in when push and fade out when pop. `FadeInOutPresentSegue` for fade in when present and fade out when dismiss.

I am not sure will they work or not, need to create spike for them, I prefer the option 3 because when the user connect the segue they can select the predefined transition for the segue. 

When we have this architecture, we can refer to a lot of nice implementation on Github to add them on bit by bit. 

[...]

PR #114, opened 2016-02-29
Present Segue

JakeLin:
In this PR, we add `PresentFadeSegue`, `PresentFadeInSegue` and `PresentFadeOutSegue` to present viewController with Fade, FadeIn and FadeOut transition animations. Also update the transition.story to test these transitions. Now can we use segue to configure a custom present transition. Because Segue doesn't support IBInspectable, all segues have to use default duration 0.35. 

Will work Transition playground to demonstrate all supported transition animations.

Relate to #19


tbaranes:
:+1:

PR #119, opened 2016-03-09
Add interactive animator for Pop 

JakeLin:
In this PR, we add interactive animator for Pop from `UINavigationController`.
1. User can configure `interactiveGestureType` in `AnimatableNavigationController` to specify the interactive animator.
2. Some animators like `FadeAnimator` and `SystemCubeAnimator` provide default interactive animator to pop.
3. Currently provides `PanInteractiveAnimator` to support pan gesture to pop.

To test it, run the app and tap on "Forget password" to see all support animators. `Fade`, `FadeIn` and `FadeOut` use `PanHorizontally` (left or right) to pop. And `SystemCubeFromLeft` uses `PanFromRight`, `SystemCubeFromRight` uses `PanFromLeft`, and other animators have their own gestures.

`SystemFlipAnimator` doesn't support interactive animator.

Will work on interactive animator for Dismiss, the architecture is almost done for transitions.

Refer to #19


lexrus:
Superb! I gonna try it tonight.

lexrus:
IMO, the liner transition is kinda weird. We can improve it by setting timing function to EaseOutQubic.

``` swift
transition.timingFunction = CAMediaTimingFunction(controlPoints: 0.215, 0.61, 0.355, 1)
// or just use easeOut
transition.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseOut)
```

BTW. The animation is too fast, especially in iPad Pro. Can we increase the default duration?

JakeLin:
@DianQK thanks, have refactored the enum to remove `from` prefix.

@lexrus have change the duration to `0.5`, it is much nicer, we can change it make it better. The subclass of Animator can also override its own duration to fit particular animation effect.

Also changed the timing for the system built-in animations to `EaseOutQubic`, it is much better than linear. The original idea to use linear was to support pan gesture to pop/dismiss, the animation can follow the finger. It is similar to the system "Swipe to pop" gesture. But maybe only few animations like `slide` will need the animation to follow the finger. So `EaseOutQubic` is better in most of cases.

Thanks guys for your valuable comments.

PR #122, opened 2016-03-16
Interactive animator for presented view controller

JakeLin:
In this PR, we support Interactive animators to dismiss presented view controller. The user can configure `interactiveGestureType` in Interface Builder or use `Present***WithDismissInteractionSegue` to support gesture.

Features:
1. The user can configure `interactiveGestureType` in `AnimatableViewController` in Interface Builder to define a gesture for dismissal.
2. New Segues: `PresentFadeWithDismissInteractionSegue`,`PresentFadeInWithDismissInteractionSegue` and `PresentFadeOutWithDismissInteractionSegue` to support interactive gesture to dismiss view controller.
3. Update the transition demo to support gesture to dismiss. Tap on "Forgot Password" to see the demo.

Refer to #19 

The architecture for transition animations are done. Will add more interactive animators and transition animators. May release 2.0 first and leave  interactive animators and transition animators as 2.\* releases.


tbaranes:
I have to say that I didn't follow all the implementation, it seems nice, but I will make a deeper look by implementing directly another animation (when I will have some spare time). 

I completely agree with you about to release a first v2.0 then making more minor versions supporting new animations. Releasing a first version will (maybe) help to more feedbacks about these features.

JakeLin:
Thanks @tbaranes, I will write some documents in Wiki to describe how to add new animators. Because we are using Apple's API, all animators only need to override two methods. It will be very simple. For interactive animators, we need some refactor when we support more gestures. `Navigator` and `Presenter` are still using `PanInteractiveAnimator`, it will be a protocol or abstract class. but won't change the API at all.  

Will release 2.0 once we complete the documentation for README and CHANGELOG.

PR #123, opened 2016-03-18
Fix slideIn/Out animations on external screens

tbaranes:
Using airplay / cable, the slideIn/Out animations won't use the right frame, here a fix! :)


JakeLin:
Thanks, great, :+1:

PR #134, opened 2016-03-28
System curl transition

tbaranes:
This PR add a new transition animator: `SystemCurlAnimator`. 

First of all, you really did a great work @JakeLin on that animator implementation! It was really easy to implement it, I quickly read your wiki, but I mostly use the existing code to implement the new one. The wiki seems complete and easy to understand :tada: 

Some though about the current implementation of Animators:
- `TransitionFromDirection`: a bit annoying when we doesn't need all the direction. In my case, I'm only using  `Top` and `Bottom`, since I didn't want to make all my properties optional, I just make a useless hack, and say `Right` will have the same behaviour as `Up`... not really nice
- `AnimatorFactory`: we currently have 13 different animators, and that number will continue to grow. I can't imagine how that enum will looks like with 20, 30, 50 different animators. I don't have a better solution right now, but we definitely need to find another way to implement it. Maybe keeping the main enum, and creating several subenums in the main one?
- Animator example: same point as the previous one. It's easy to use now with 13 examples, but once we have 50? We need a nicer way to generate the example. An easy way (I think) would be to add sections and using grouped tableview

Anyway, nothing that important. We can think about it later if we think about a nice solution :+1: 

I'm gonna take a deeper look in that engine, and let you know if I foud others points to improve

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ TBD
 #### Enhancements
 
 - Support Xcode 7.3 and Swift 2.2
-- Add `SystemCurlAnimator` to support `SystemCurlFromTop` and `SystemCurlFromBottom` transition animations
+- Add `SystemPageCurlAnimator` to support `SystemPageCurlFromTop` and `SystemPageCurlFromBottom` transition animations
 
 #### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ TBD
 
 #### Enhancements
 
-Support Xcode 7.3 and Swift 2.2
+- Support Xcode 7.3 and Swift 2.2
+- Add `SystemCurlAnimator` to support `SystemCurlFromTop` and `SystemCurlFromBottom` transition animations
 
 #### Bugfixes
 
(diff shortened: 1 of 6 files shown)


JakeLin:
Thanks again @tbaranes 
- For `TransitionFromDirection`, I think you did a good job to merge `Right` with `Top`, I didn't consider some transition doesn't have all four directions. Maybe we can support the supported ones and all the rest use the default one, e.g. `Top` or `Left` depends on the transition. In Interface Builder the user can't configure the unsupported ones anyway since we treat them as invalid value.
- For the `TransitionAnimationType`, I think about to use `SystemCube(.Left)` to reduce the number of options. But Interface Builder doesn't support it and we have to use hack like `String("SystemCubeFromLeft")` ->  `SystemCube(.Left)` to do the conversion. Do you think we should use `SystemCube(.Left)` for better programmatic API?
- Haven't got idea to improve Animator example yet. 😤

Please have a look at the engine. I don't mind any breaking changes in the early stage.

Feel free to merge it when you think is ready. 👍

tbaranes:
I'm gonna merge this, it's working, engine improvements will be for another PR! :dancers: 
- `TransitionFromDirection`: it's ok with me to make a default value for the unused ones, it will probably never used so...
- Can't we used the same system as `DesignableMask` to handle that issue? I like your idea to use `SystemCube(.Left)`, it's really easy to understand, and will make the code more compact. If we can't have it in IB, maybe we should supports the both systems?
- If we break the `TransitionAnimationType` in `Cube`, `Flip`... what about a grouped tableview with sections:
  - Section 1: Cube
    - Cube from left
    - Cube from top
    - ...
  - Section 2: Flip
    - Flip from left
    - Flip from top
    - ...
  - ...

JakeLin:
@tbaranes Good idea, if we use `SystemCube(.Left)` then we should be able to group them by the main type like `SystemCube`.

The reason they start with `System` because all `System` transition animations can only support push/pop not present/dismiss, I don't know why, maybe an issue with `CATransition`

PR #135, opened 2016-03-28
Privatise MaskDesignable methods

### LINKED ISSUE #101: Mask border
As @lexrus pointed out in #99, we currently doesn't draw well the borders when a maskType is specified. 
I join Lex to say that is a really annoying miss since a new user can be disappointed by the library.

Any thought of how we can handle this issue?

tbaranes:
Should close #101 
I also used this PR to group the different code shape together.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ All notable changes to this project will be documented in this file.
 
 #### API breaking changes
 
-TBD
+- Removing `MaskDesignable` public methods:
+   - `maskCircle()`, use instead: `view.maskType = String(MaskType.Circle)`
+   - `maskStar(_:)`, use instead: `view.maskType = `Star(6)`
+   - `maskPolygon()`, use instead: `view.maskType = `String(MaskType.Polygon)`
+   - `maskTriangle()`, use instead: `view.maskType = `String(MaskType.Triangle)`
+   - `maskWave(_:_:_:)`, use instead: `view.maskType = `Wave(up, 16, 40)`
+
 
 #### Enhancements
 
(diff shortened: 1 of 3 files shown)


JakeLin:
Well, I was going to bed and received this PR, well done @tbaranes :+1:

CHANGELOG.md
```diff
@@ -7,7 +7,13 @@ All notable changes to this project will be documented in this file.
 
 #### API breaking changes
 
-TBD
+- Removing `MaskDesignable` public methods:
+   - `maskCircle()`, use instead: `view.maskType = String(MaskType.Circle)`
```
lexrus, line 11: Hi Tom,
What's the difference between MastType.Circle.rawValue and String(MaskType.Circle)?

JakeLin, line 11: @lexrus I think `rawValue` is for the actual value it saves. e.g. if the enum inherits from `Int` then rawValue is a number. But `String()` is the presentation of the option. Here is the code I test

```
enum A : Int {
  case A1
}

// A1
String(A.A1)

// 0
A.A1.rawValue
```

If the enum inherits from `String`, I don't see the difference, but if the enum doesn't inherit from any type, `rawValue` is not available.

```
// B1
String(B.B1)

// Value of type `B` has no member `rawValue`
B.B1.rawValue
```

I think I changed them in this commit https://github.com/JakeLin/IBAnimatable/pull/110/commits/954a55070ef61c5e0bd2787b4df22218ef57d69c

tbaranes, line 11: You are right @lexrus, no reason to do `String()` instead of `rawValue`. Thanks for the change @JakeLin!

PR #136, opened 2016-03-28
Fixed gradient cut off

### LINKED ISSUE #129: Gradient color cut off in UITableViewControllers
As @sambing mentioned, this is a bug in Gradient color. When use Predefined gradient color in a `UITableViewController`, it color will cut off.

simulator screen shot mar 22 2016 10 05 23 pm

Expected behaviour: Should display the gradient color from edge to edge.

The sample app can be found in https://github.com/JakeLin/IBAnimatable-PackageManagementExamples/tree/master/IBAnimatable-CocoaPods

Related to https://github.com/JakeLin/IBAnimatable/issues/120

tbaranes:
Should close #129 
The issue here was the same as the border. Drawing the gradient in `configInspectableProperties()` isn't enough because we don't know the _real_ frame. In our case, the frame was `(600,600)` (xib frame), but not the final one.

Moving the `configGradient()` in `layoutSubviews()` fixed this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ All notable changes to this project will be documented in this file.
 
 #### Bugfixes
 
-TBD
+- Fixed a bug that was making `GradientDesignable` not filling the dedicated frame
 
 ### [2.0](https://github.com/JakeLin/IBAnimatable/releases/tag/2.0)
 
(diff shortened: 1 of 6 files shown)


JakeLin:
:+1:  Good found, thanks @tbaranes

PR #137, opened 2016-03-28
Refactoring TransitionAnimationType

tbaranes:
Following the discussion in #134, I tried to review the implementation of `TransitionAnimationType` in order to make it thiner. The current implementation is answering that needs: the enum is really smaller, and the switch is more generic / readable.

The architecture seems nice, but there are some points that need a generic solution:
- `TransitionAnimationType` from string -> how to implement it?
- Iterating on TransitionAnimationType and handles all the cases (which means with the different directions...) to update the examples. That also implicate a way to divide the final array in different section for each transition family and _readable_ name
  For now, I just hard coded one case of `SystemCube`, and `Fade`, it's working as expected.

Any suggestions for these issues? Until now, I didn't find any nicer solution than the implementation currently used in master.

If you want to make some tests, use this branch, it's still very experimental :thought_balloon: 

/!\ NOT READY TO BE MERGED /!\

### DIFF
--- a/IBAnimatable/PresenterManager.swift
+++ b/IBAnimatable/PresenterManager.swift
@@ -19,12 +19,12 @@ class PresenterManager {
   }
   
   // MARK: - Private
-  private var cache = [TransitionAnimationType: Presenter]()
+  private var cache = [String: Presenter]()
   
   // MARK: Inertnal Interface
   func retrievePresenter(transitionAnimationType: TransitionAnimationType, transitionDuration: Duration = defaultTransitionDuration, interactiveGestureType: InteractiveGestureType? = nil) -> Presenter {
     // Get the cached presenter
-    let presenter = cache[transitionAnimationType]
+    let presenter = cache[transitionAnimationType.stringValue]
     if let presenter = presenter {
       // Update the `transitionDuration` and `interactiveGestureType` every time to reuse the same presenter with the same type
       presenter.transitionDuration = transitionDuration
@@ -34,7 +34,7 @@ class PresenterManager {
     
     // Create a new if cache doesn't exist
     let newPresenter = Presenter(transitionAnimationType: transitionAnimationType, transitionDuration: transitionDuration, interactiveGestureType: interactiveGestureType)
-    cache[transitionAnimationType] = newPresenter
+    cache[transitionAnimationType.stringValue] = newPresenter
     return newPresenter
   }
 }
--- a/IBAnimatable/TransitionAnimationType.swift
+++ b/IBAnimatable/TransitionAnimationType.swift
@@ -20,13 +20,60 @@ public enum TransitionAnimationType {
     return String(self)
   }
 
-  static func fromString(string: String) -> TransitionAnimationType? {
-    if String(TransitionAnimationType.Fade) == string {
-      return .Fade
-    } else if String(TransitionAnimationType.SystemCube(direction: .Left)) == string {
-      return .SystemCube(direction: .Left)
+  static func fromString(transitionType: String) -> TransitionAnimationType? {
+    if transitionType.hasPrefix("Fade") {
+      return fadeTransitionAnmationType(transitionType)
+    } else  {
+      return fromStringWithDirection(transitionType)
+    }
+  }
+
+}
+
+// MARK: - TransitionAnimationType from string
+
+private extension TransitionAnimationType {
+  
+  static func transitionDirection(forTransitionType transitionType: String) -> TransitionFromDirection? {
+    let range = transitionType.rangeOfString("(")
+    let transitionType = transitionType.stringByReplacingOccurrencesOfString(" ", withString: "")
+                                        .lowercaseString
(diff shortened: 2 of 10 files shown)


IBAnimatable/TransitionAnimationType.swift
```diff
@@ -8,18 +8,25 @@ import Foundation
...
+  case SystemPageCurl(direction: TransitionFromDirection)
+
+  var stringValue: String {
+    return String(self)
+  }
+
+  static func fromString(string: String) -> TransitionAnimationType? {
+    if String(TransitionAnimationType.Fade) == string {
+      return .Fade
+    } else if String(TransitionAnimationType.SystemCube(direction: .Left)) == string {
```
JakeLin: I think that may not work since it is a fully qualified like SystemCube(IBAnimatable.TransitionFromDirection.Left)", we may need to do some like what we did in `maskStarFromString` in `MaskDesignable`, in IB, we may configure the string as `"SystemCube(Left)"`

IBAnimatable/PresenterManager.swift
```diff
@@ -19,12 +19,12 @@ class PresenterManager {
   }
   
   // MARK: - Private
-  private var cache = [TransitionAnimationType: Presenter]()
+  private var cache = [String: Presenter]()
```
JakeLin, line 22: Using `TransitionAnimationType.stringValue` as the key is good.

IBAnimatableApp/TransitionTableViewController.swift
```diff
@@ -12,16 +12,17 @@ class TransitionTableViewController: UITableViewController {
   // MARK: - Lifecycle
   override func viewDidLoad() {
     super.viewDidLoad()
-    
-    for type in iterateEnum(TransitionAnimationType) {
-      transitionAnimations.append(type)
-    }
+    transitionAnimations.append(.Fade)
```
JakeLin: Let's think about this implementation here. We need to group them into sections. But how 🤔

JakeLin:
@tbaranes nice attempt , I really like the new `TransitionAnimationType`, we maybe able to do similar thing for `MaskType` as so on.

IBAnimatableApp/TransitionTableViewController.swift
tbaranes: Generic and nice result is hard in this case! Don't find a nice solution yet 🤔

IBAnimatable/PresenterManager.swift
tbaranes: Agree, gonna change this.

tbaranes, line 22: Since `stringValue` is a var, we can't directly use `TransitionAnimationType.stringValue`, or I'm missing something. Any suggestions ?

IBAnimatable/TransitionAnimationType.swift
tbaranes: Yeah, sure. That part is really not finished :sweat_smile: I let it that way to talk that new implementation, but didn't find a nice solution yet. 
In order to do as `MaskDesignable`, we are will need to hard code all the enums, is that a good solution?

IBAnimatable/PresenterManager.swift
JakeLin, line 22: @tbaranes Sorry, I may confuse you, I think using `String` as the key is good, and we save `TransitionAnimationType.stringValue` as the value of the key, it is unique. I like your solution here.

tbaranes, line 22: Oh, cool then :laughing:

tbaranes:
Here two new commits who may finalise that PR if it seems correct to everyone:
- `TransitionAnimationType`: it will ask a bit more work than before, we have to add a new `if` for each new transition. I tried to make it generic (at least the direction). I don't like that much this generator, but don't have any better idea right now, and I'm not sure that we can make it more generic using an enum with params since we are using hard coding string from xib. 
- Example project: here again, a bit more work, one line will be add manually for each new transition, but it's worth it even if it's not generic. The interface is cleaner and more readable for the user:

simulator screen shot 29 mar 2016 20 25 28

Everything is working as expected! What do you think about that implementation? :)
Also, shouldn't we create a new release with or without that PR? The changelog is already not that bad :laughing:

JakeLin:
Very cool implementation. I like it very much. 

We may have more `else if` branches in `static func fromString(transitionType: String) -> TransitionAnimationType?` in the future to support some transition more than direction. For example, ZoomIn/Out.

I will update the wiki to mention `generateTransitionTypeData` when we develop new transition animations. 

Please merge it when you feel is ready. 👍👍👍 Well done.

tbaranes:
Let's merge that first step! It's still improvable with less breaking change :laughing: 

Could you update the CHANGELOG to explain the breaking change? Since it's a big one, I prefer let you write it to be clear. Also, if it's ok for you, we can release the IBAnimatable 2.1 :+1:

JakeLin:
Cool, @tbaranes doing 2.1 now.

PR #139, opened 2016-03-29
Fixed mask frame

tbaranes:
Same context as the border, gradient...: we need to `configMask()` in the `layoutSubviews` in order to be sure that we draw the right frame.
That bug is easily reproductible when drawing a mask (wave for example) taking all the width on a second screen.


JakeLin:
Good one, :+1:  thanks again.

Issue #126, opened 2016-03-19
More system transition animators (animation controllers)


JakeLin:
To have more system transition animations can be configured in Interface Builder.

Unlock all system built-in animations, We have `SystemCubeAnimator` and `SystemFlipAnimator` now, we can easily unlock all system built-in animations in [SystemTransitionType](https://github.com/JakeLin/IBAnimatable/blob/master/IBAnimatable/SystemTransitionType.swift). They all have same implementation with different parameters. Refer to http://iphonedevwiki.net/index.php/CATransition

Documents:
- [How to develop an animator (animation controller)](https://github.com/JakeLin/IBAnimatable/wiki/How-to-develop-an-animator-%28animation-controller%29)
- [How to develop an animator using system built in animation](https://github.com/JakeLin/IBAnimatable/wiki/How-to-develop-an-animator-using-system-built-in-animation)

tbaranes:
If we already have all of them, it would be a huge work :laughing: 
Seems we are gonna have a lot of fun!

JakeLin:
@tbaranes too good to be true if we can implement all of them. I think we can do frequent releases for every one or few animators. Once we have implemented some of popular transition animators, it is good enough to use in most of transitions.

tbaranes:
Just to keep in mind: the animation `SystemPageCurl` could be refactored in order to use `animateWithCATransition`, in order words creating a new enums instead of using the enums. The result won't change but will create a breaking change, nevertheless, it will make sense.

Also, here the full list of the system animators and their implementation state:
- [x] Fade
- [x] MoveIn
- [x]  Push
- [x]  Reveal
- [x] Flip
- [x]  Cube
- [x]  PageCurl
- [x]  PageUnCurl
- [x]  RippleEffect
- [x]  SuckEffect
- [x]  CameraIris
- [x]  CameraIrisHollowOpen
- [x]  CameraIrisHollowClose
- [x]  Rotate

That list doesn't contains the private transition, source from [stackoverflow](http://stackoverflow.com/a/12593393/4177109)

JakeLin:
Don't mind for the breaking changes for renaming `SystemPageCurl`, should follow the system names. :+1:  for the check list.

tbaranes:
Should we renamed that issue `More system transition animators` and create a new one for the `custom transition animators`?

Once all the PR merged, we also may make a new released since all the system ones are implemented.
What do you think? :dancers:

JakeLin:
[...]

PR #143, opened 2016-03-30
System animator ripple effect

tbaranes:
#126 WIP

### DIFF
--- a/IBAnimatable/TransitionAnimationType.swift
+++ b/IBAnimatable/TransitionAnimationType.swift
@@ -15,6 +15,7 @@ public enum TransitionAnimationType {
   case SystemCube(direction: TransitionFromDirection)
   case SystemFlip(direction: TransitionFromDirection)
   case SystemPageCurl(direction: TransitionFromDirection)
+  case SystemRippleEffect
 
   var stringValue: String {
     return String(self)
@@ -23,6 +24,8 @@ public enum TransitionAnimationType {
   static func fromString(transitionType: String) -> TransitionAnimationType? {
     if transitionType.hasPrefix("Fade") {
       return fadeTransitionAnmationType(transitionType)
+    } else if transitionType.hasPrefix("SystemRippleEffect") {
+      return .SystemRippleEffect
     } else  {
       return fromStringWithDirection(transitionType)
     }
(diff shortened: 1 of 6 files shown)


tbaranes:
Wrong mania, my bad.

IBAnimatable/TransitionAnimationType.swift
```diff
@@ -23,6 +24,8 @@ public enum TransitionAnimationType {
   static func fromString(transitionType: String) -> TransitionAnimationType? {
     if transitionType.hasPrefix("Fade") {
       return fadeTransitionAnmationType(transitionType)
+    } else if transitionType.hasPrefix("SystemRippleEffect") {
```
JakeLin: Can we check exact match first like 

```
if let transitionType = TransitionAnimationType(rawValue: transitionType) {
  return transitionType      
} else {
  return fromStringWithDirection
}
```

even including `Fade`, `FadeIn` and `FadeOut` in first if branch.

JakeLin:
:+1: except for a small comment

IBAnimatable/TransitionAnimationType.swift
tbaranes: Not sure what did you mean. Something like that?

```swift
    if transitionType.hasPrefix("SystemRippleEffect") {
      return .SystemRippleEffect
    } else if transitionType.hasPrefix("SystemSuckEffect") {
        return .SystemSuckEffect
    } else if transitionType.hasPrefix("Fade") {
      return fadeTransitionAnimationType(transitionType)
    } else if transitionType.hasPrefix("SystemCameraIris") {
        return cameraIrisTransitionAnimationType(transitionType)
    } else  {
      return fromStringWithDirection(transitionType)
    }
```

tbaranes: I agree that would be better, but since the enum isn't a string, we can't create a `transitionType` this way. Any suggestions?

JakeLin: You are right, because `TransitionAnimationType` doesn't inherit from `String`, we can't use `init(rawValue: String)`. I haven't found any solutions for this. I am happy with your current implementation. Please merge it when you feel it is ready. 👍

tbaranes: Let's keep that in mind for later, we definitely need a better solution!

PR #145, opened 2016-03-30
System suck effect animator

tbaranes:
#126 WIP

You really did an awesome job with your transition engine @JakeLin, it's so easy to add a new animation!

For this one, I'm not sure about the reverse animation. It looks weird to me, but don't really know what should we use instead.


JakeLin:
@tbaranes you are so productive. 👍

tbaranes:
It's always a pleasure to work on IBAnimatable :tada: 
Waiting for travis, then merging that PR.

PR #146, opened 2016-03-31
Removing private system transition animations

tbaranes:
I was trying to implement new system transitions, but it wasn't working. I finally ending on stack overflow reading that [answer](http://stackoverflow.com/a/12593393/4177109). That explains we can't implements all the system transition since the most of them are private.

In this PR, I just removed the system transition type we can't implement. If we want have this kind of effect, we have to create our own animators.

I also clean up the list in #126


JakeLin:
Thanks @tbaranes I didn't know that, most of them can be done in our own animators if needed. :+1:

PR #147, opened 2016-03-31
Camera iris transition animation (without hollow state)

tbaranes:
#126  WIP

I missed it yesterday, here it is: system camera iris transition

### DIFF
--- a/IBAnimatableApp/TransitionTableViewController.swift
+++ b/IBAnimatableApp/TransitionTableViewController.swift
@@ -50,7 +50,7 @@ private extension TransitionTableViewController {
     transitionAnimationsHeaders.append("Suck Effect")
     transitionAnimations.append(["SystemSuckEffect"])
     transitionAnimationsHeaders.append("Camera Iris")
-    transitionAnimations.append(["SystemCameraIris(Open)", "SystemCameraIris(Close)"])
+    transitionAnimations.append(["SystemCameraIris", "SystemCameraIris(HollowOpen)", "SystemCameraIris(HollowClose)"])
     transitionAnimationsHeaders.append("Ripple Effect")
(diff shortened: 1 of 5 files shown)


IBAnimatableApp/TransitionTableViewController.swift
```diff
@@ -50,7 +50,7 @@ private extension TransitionTableViewController {
     transitionAnimationsHeaders.append("Suck Effect")
     transitionAnimations.append(["SystemSuckEffect"])
     transitionAnimationsHeaders.append("Camera Iris")
-    transitionAnimations.append(["SystemCameraIris(Open)", "SystemCameraIris(Close)"])
+    transitionAnimations.append(["SystemCameraIris", "SystemCameraIris(HollowOpen)", "SystemCameraIris(HollowClose)"])
```
JakeLin, line 53: do we need to add `SystemCameraIris` to the end like `transitionAnimations.append(["SystemCameraIris", "SystemCameraIris(HollowOpen)", "SystemCameraIris(HollowClose)", "SystemCameraIris"])`

JakeLin:
:+1:  I didn't know the `.None` option either.

IBAnimatableApp/TransitionTableViewController.swift
tbaranes, line 53: Why should we?

JakeLin, line 53: @tbaranes sorry, `SystemCameraIris` is the first one, we don't extra one.

JakeLin, line 53: @tbaranes sorry, `SystemCameraIris` is the first one, we don't need the extra one.

PR #148, opened 2016-03-31
System curl refactoring

tbaranes:
As said in #126, `SystemPageCurl` wasn't using the `SystemTransitionType` which an error, here a refactor making that transition cleaner.

Didn't update the CHANGELOG in order to avoid conflicts with the others PR, gonna make it after the merge if that's ok for everyone.


IBAnimatable/TransitionAnimationType.swift
```diff
@@ -62,6 +64,16 @@ private extension TransitionAnimationType {
     return nil
   }
   
+  static func pageTransitionAnimationType(transitionType: String) -> TransitionAnimationType? {
+    let transitionType = cleanTransitionType(transitionType)
+    if transitionType.containsString("uncurl") {
```
JakeLin, line 69: Do we need to support "unCurl"/ "UnCurl" or just make the checking case-insensitive?

JakeLin:
:+1: I will update the CHANGELOG.md once it is in master. Thanks again.

PR #149, opened 2016-03-31
Use framework in demo app

JakeLin:
### Why
- As @molescat advised, the demo App should use IBAnimatable as a framework like all other app using it with CocoaPods, Carthage or SPM.
- According to Kevin's (from IB team) tweet [If you're looking for faster IBDesignable build times, put custom UI code in it's own framework. IB only builds the target w/ the designable](https://twitter.com/kbcathey/status/615656846167138304) Using framework can speed up a lot. I have tested it, the first time to render the main storyboard is about 2 seconds, and down to less than 1 second for rest of time to open the main storyboard. It used to be 5~10 seconds or even longer 😱.
### What has been changed
- Remove all IBAnimatable elements from `IBAnimatableApp` target. They are in `IBAnimatable` target only.
- Set `IBAnimatable.framework` as "Embedded Binaries" for `IBAnimableApp` target.
- Fix some constraints for Xcode 7.3
- Enlarge some touchable area for some buttons, thanks to @molescat for pointing out those issues.

@tbaranes you may need to rebase your branches depending which PR will get into master first. I would like to get this PR into master earlier than yours 😁, because I started this branch couple weeks ago and rebased it from time to time. It still took me 10 minutes to rebase it and merge the Xcode project file this morning, especially the Internet connection wasn't good in the train. I don't want to rebase it again selfishly👻.

### DIFF
--- a/IBAnimatable/TransitionPageType.swift
+++ b/IBAnimatable/TransitionPageType.swift
@@ -1,7 +1,4 @@
 //
-//  TransitionPageState.swift
-//  IBAnimatableApp
-//
 //  Created by Tom Baranes on 31/03/16.
 //  Copyright © 2016 Jake Lin. All rights reserved.
 //
--- a/IBAnimatableApp.xcodeproj/project.pbxproj
+++ b/IBAnimatableApp.xcodeproj/project.pbxproj
@@ -7,58 +7,24 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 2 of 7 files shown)


tbaranes:
Ups, I didn't read that PR first so I already merged the two others :sweat: 
Otherwise, great idea :+1:

JakeLin:
Thanks @tbaranes I rebase and merge this PR now.

seanwoodward:
Even better to separate the repos too, #151

PR #150, opened 2016-04-01
fix to allow single sided borders

seanwoodward:


IBAnimatable/BorderDesignable.swift
```diff
@@ -25,16 +25,23 @@ public protocol BorderDesignable {
...
+    
     guard let unwrappedBorderColor = borderColor else {
+      clearBorders()
       return
     }
     
     if borderWidth.isNaN || borderWidth <= 0 {
+      clearBorders()
       return
     }
```
tbaranes: Couldn't we merge that `if` in the `guard`?

tbaranes:
Thanks @seanwoodward, nice catch :+1:

IBAnimatable/BorderDesignable.swift
```diff
@@ -25,16 +25,23 @@ public protocol BorderDesignable {
...
+        .forEach { $0.removeFromSuperlayer() }      
+    }
+    
     guard let unwrappedBorderColor = borderColor else {
+      clearBorders()
       return
     }
     
     if borderWidth.isNaN || borderWidth <= 0 {
+      clearBorders()
```
tbaranes: Couldn't we merge that if in the guard?
 Add a line note

IBAnimatable/BorderDesignable.swift
tbaranes: Couldn't we merge that if in the guard?

tbaranes, line 41: Since the `guard` and the `if` are doing the thing, couldn't we merge them?

tbaranes:
Nice catch :+1: 
I made a minor feedback just to improve that implementation, but looks good!

I'm gonna let @JakeLin make the final review :dancers:

JakeLin:
Good one, :+1:

agree with @tbaranes ,  it is better to group them within one. Can we optimise that in this PR as well? Thanks.

JakeLin:
@seanwoodward thanks again for your PR, do you have time to make the last change according to @tbaranes 's comment? Or we can merge this PR then make the change after. We are about to release 2.2, 😉 thanks.

tbaranes:
I think we can merge it. We can handle that comment, or even do it directly in master since it's not a big change.

JakeLin:
@seanwoodward thanks again for your PR. @tbaranes working on release 2.2 now.

PR #152, opened 2016-04-02
Removing useless system type transitions

tbaranes:
This PR contains only two minor refactors:
- Make `FlipAnimator` using the method `animateWithCATransition` which clean a little bit the code
- Removing `oglFlip`, `alignedFlip`, `alignedCube` because the transition is the same as `Flip` and `Cube`. I try the three of them, but can't see any differences in the animation. If there any reasons, I don't see a reason to support them since it will just confuse the user: which one between the three should I use to make a flip transition?

For that last point, I may missing the main point :laughing: 

PS: I also clean the todo list in the issue #126


JakeLin:
:+1:  Thanks, agree with you, we are making a library to help the users not confuse them.

PR #153, opened 2016-04-02
SystemMoveIn, SystemPush, and SystemReveal animators

tbaranes:
In this PR, we continue to add more system animators (#126):
- SytemMoveIn
- SystemPush
- SystemReveal

I did one PR for the three because the implementation is exactly the same: one system animator supporting the four directions.


IBAnimatable/SystemMoveInAnimator.swift
```diff
@@ -0,0 +1,53 @@
...
+      self.reverseAnimationType = .SystemMoveIn(direction: .Right)
+      self.interactiveGestureType = .PanFromRight
+    case .Right:
+      self.transitionAnimationType = .SystemMoveIn(direction: .Right)
+      self.reverseAnimationType = .SystemMoveIn(direction: .Left)
+      self.interactiveGestureType = .PanFromLeft
+    case .Top:
+      self.transitionAnimationType = .SystemMoveIn(direction: .Top)
+      self.reverseAnimationType = .SystemMoveIn(direction: .Bottom)
+      self.interactiveGestureType = .PanFromTop
```
JakeLin, line 34: I think it should be `. PanFromBottom`, need to fix for line 38, and `SystemPushAnimator` as well.

JakeLin:
👍 except for a smalll comment. Well done, you are so productive.

IBAnimatable/SystemMoveInAnimator.swift
tbaranes, line 34: Do you think? If that so, we have the wrong implementation of `InteractiveGesture` for each animators. Gonna try it on real device, and get back to you.

JakeLin, line 34: The pop gesture should be reverse as push animation. You are right, I implemented `handleGesture` in `PanInteractiveAnimator` wrongly, please merge this PR when you free ready, I fix them all in next PR. Thanks.

PR #154, opened 2016-04-02
SystemRotate animators

### LINKED ISSUE #126: More system transition animators (animation controllers)

tbaranes:
Finally, the last system animator, should we close #126 to create another PR for the custom transition animators?

Nevertheless, if some of you could try that that transition, it doesn't seems to work for the 180 and CCW. I suppose that's due to the `subtypes`, but I didn't find any documentation of how it should be writing...

The `SystemRotate(90)` is working well, but I'm a bit confuse about the others.

PS: Didn't update the CHANGELOG to avoid conflicts with the others PR.

### DIFF
--- a/IBAnimatable/AnimatorFactory.swift
+++ b/IBAnimatable/AnimatorFactory.swift
@@ -32,6 +32,8 @@ struct AnimatorFactory {
       return SystemCameraIrisAnimator(hollowState: hollowState, transitionDuration: transitionDuration)
     case .SystemRippleEffect:
       return SystemRippleEffectAnimator(transitionDuration: transitionDuration)
+    case .SystemRotate(let degree):
+      return SystemRotateAnimator(withDegree: degree, transitionDuration: transitionDuration)
     }
   }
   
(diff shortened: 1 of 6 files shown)


JakeLin:
Thanks, I also tried it out, all animations look like `SystemRotate(90)`, it maybe a problem of `subtype`, but the only document I found is http://iphonedevwiki.net/index.php/CATransition I don't know how to fix it. 

I am OK to pull it into master and make it as a  known issue in API doc. 👍

tbaranes:
The pop of `180ccw` was looking a bit different, but I have no idea why :laughing: 
I'm gonna merge that, let's make that issue known, maybe we will have a fix later!

PR #156, opened 2016-04-02
Fix the FromTop and FromBottom directions in system built-in animations

JakeLin:
After tested, the implementation of `PanInteractiveAnimator` is right, but the directions `FromTop` and `FromBottom` in `SystemCubeAnimator`, `SystemMoveInAnimator`, `SystemPushAnimator` and `SystemRevealAnimator` are wrong, I don't know why, but after I reverse them in `TransitionFromDirection`, everything works as expected. 

@tbaranes please have a look. Thanks.


tbaranes:
One more weird behaviour :cry: 
Good to go :+1:

Issue #155, opened 2016-04-02
Custom transition animators (animation controllers)


JakeLin:
Since we have all supported system built-in transition animations.  Refer to [More system transition animators (animation controllers)](https://github.com/JakeLin/IBAnimatable/issues/126). We can start implement more custom transition animators.

VCTransitionsLibrary can be a good library to start with, we can port https://github.com/ColinEberhardt/VCTransitionsLibrary to Swift and make it designable in Interface Builder.

Other concepts
http://transitiontreasury.com
https://github.com/entotsu/TKSubmitTransition
https://github.com/NewAmsterdamLabs/ZOZolaZoomTransition
https://github.com/AlvaroFranco/AFBlurSegue
https://dribbble.com/shots/2131325-Bike
https://dribbble.com/shots/1216761-Liquid-Animation

Any feedback is welcome, thanks.

Document:
- [How to develop an animator (animation controller)](https://github.com/JakeLin/IBAnimatable/wiki/How-to-develop-an-animator-%28animation-controller%29)

tbaranes:
Plus creating these customs transitions animators, I'd like to keep in mind some points purely related to the code that (I think) we should improve:
- Core's improvements:
    - [ ] Find a way to avoid that multiple `if / else if` to check which transitions. Once again, it's ok for now since we don't have _that much_ transitions, but it will fast become big
- Demo apps improvements:
   - [ ] Improve the demo app example: the list will grow and grow, in a short time, it will really be confusing. It's just an interface question, but we can improve it. 
  - [ ] Add the title when playing with a transition. That will make it even easier to use!

tbaranes:
Here a list of which transitions we can port from [VCTransitionsLibrary](https://github.com/ColinEberhardt/VCTransitionsLibrary):
- [x] Flip
- [x] Turn
- [x] Fold
- [x] Explode
- [x] Cards
- [x] NatGeo
- [x] Portal

JakeLin:
@tbaranes the behaviour of `Flip` in VCTransitionsLibrary is different from `SystemFlip`, we maybe able to use `Flip` because the other once called `SystemFlip`.

tbaranes:
Here another list of transitions: https://github.com/onmyway133/fantastic-ios-animation/blob/master/Animation/transition.md
There definitely some transitions that can have their place in IBAnimatable!

tbaranes:
Some others lovely animations: https://github.com/PiXeL16/RevealingSplashView

[...]

PR #157, opened 2016-04-03
Explode transition

tbaranes:
I always wanted to make an explode transition :laughing: 

Some points about that PR:
- You should definitely try it before validate it, it seems good to me, but we can adjust some default 
  values
- I didn't update the CHANGELOG, I think we shouldn't ship that transition with the new release, but keep it for the 2.3 in case we have breaking changes coming for custom transitions
- Didn't set any `InteractiveGestureType`, a nice gesture would be the pinch, but we need to create it before use it
- In the demo app, I group the transitions without parameters to make them clearer, but in a long term process, not sure what it will looks like
- Another thing we may improve is to make some values designable: number of snapshots, min / max, angle rotation... should we or is making it too complex? If we should, any thought about how to implement them?

Also, one question about the segue implementation, don't you think we can have a deeper abstraction? I mean, for most of them (currently implemented), they are all exactly the same, the only difference is `TransitionAnimationType` which is the most important part, but it duplicates a lot of code for only one parameter. I suppose the main reason for this is that we need one custom class by segue, but is there any way to improve it?

Once again, thanks for your great work on the transition abstraction, it's really easy to create new one :+1: 

Related to #155

### DIFF
--- a/IBAnimatable/ExplodeAnimator.swift
+++ b/IBAnimatable/ExplodeAnimator.swift
@@ -72,11 +72,14 @@ private extension ExplodeAnimator {
       snapshots.forEach{
         let xOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
         let yOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
-        $0.frame = CGRectOffset($0.frame, xOffset, yOffset)
-        $0.alpha = 0.0
-        
         let angle = self.randomFloatBetween(lower: -10.0, upper: 10.0)
-        $0.transform = CGAffineTransformScale(CGAffineTransformMakeRotation(angle), 0.01, 0.01)
+
+        let translateTransform = CGAffineTransformMakeTranslation($0.frame.origin.x - xOffset, $0.frame.origin.y - yOffset)
+        let angleTransform = CGAffineTransformRotate(translateTransform, angle)
+        let scaleTransform = CGAffineTransformScale(angleTransform, 0.1, 0.1)
+
+        $0.transform = scaleTransform
+        $0.alpha = 0.0
       }
     },
     completion: { _ in
--- a/IBAnimatable/ExplodeAnimator.swift
+++ b/IBAnimatable/ExplodeAnimator.swift
@@ -35,8 +35,7 @@ extension ExplodeAnimator: UIViewControllerAnimatedTransitioning {
       return
     }
     
-    containerView.addSubview(toView)
-    containerView.sendSubviewToBack(toView)
+    containerView.insertSubview(toView, atIndex: 0)
     
     let snapshots = createSnapshots(toView: toView, fromView: fromView, containerView: containerView)
     containerView.sendSubviewToBack(fromView)
@@ -57,7 +56,7 @@ private extension ExplodeAnimator {
     
     let fromViewSnapshot = fromView.snapshotViewAfterScreenUpdates(false)
     for x in 0.stride(to: Int(size.width), by: Int(size.width / xFactor)) {
-      for y in 0.stride(to: Int(size.height), by: Int(size.width / yFactor)) {
+      for y in 0.0.stride(to: Double(size.height), by: Double(size.width / yFactor)) {
         let snapshotRegion = CGRect(x: CGFloat(x), y: CGFloat(y), width: size.width / xFactor, height: size.height / yFactor)
         let snapshot = fromViewSnapshot.resizableSnapshotViewFromRect(snapshotRegion, afterScreenUpdates: false, withCapInsets: UIEdgeInsetsZero)
         snapshot.frame = snapshotRegion
@@ -69,16 +68,16 @@ private extension ExplodeAnimator {
   }
   
   func animateSnapshotsExplode(snapshots: [UIView], completion: () -> Void) {
-    UIView.animateWithDuration(2, animations: {
-      for view in snapshots {
+    UIView.animateWithDuration(transitionDuration, animations: {
+      snapshots.forEach({
         let xOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
         let yOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
-        view.frame = CGRectOffset(view.frame, xOffset, yOffset)
-        view.alpha = 0.0
+        $0.frame = CGRectOffset($0.frame, xOffset, yOffset)
+        $0.alpha = 0.0
         
         let angle = self.randomFloatBetween(lower: -10.0, upper: 10.0)
-        view.transform = CGAffineTransformScale(CGAffineTransformMakeRotation(angle), 0.1, 0.1)
-      }
+        $0.transform = CGAffineTransformScale(CGAffineTransformMakeRotation(angle), 0.01, 0.01)
+      })
     }) { _ in
       for view in snapshots {
         view.removeFromSuperview()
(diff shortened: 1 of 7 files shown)


JakeLin:
@tbaranes thanks, will look at it soon. 

I was thinking to work on more [More interactive animators/controllers](https://github.com/JakeLin/IBAnimatable/issues/125) after realse 2.2. and we can set it to Pinch once we have Pinch gesture.

To make some values designable, I think we can use the way we did for `wave` animation, put all parameter inside like `Explode(transitionDuration, numberOfSnapshots, min, max, angleRotation...)`, how do you think? The current one is a good starting point. enhance it later.

Good pick-up for the segue implementation, I tried to encapsulate them to use less number of segues. but Interface Builder doesn't support `@IBInspectable` property in segue. It means we can't change parameters in Segue in IB, we can only change it in code. To let the user configure different custom transition animations in IB, currently, we need different segues. If IB can support `@IBInspectable` property in segue, we can group most of them together. For the implementation, e.g. templating most of the implementation for all similar segues, I think we can still improve, although I haven't got idea yet.

tbaranes:
I see for the segue, we can add that point to the list of "transitions code" that needs improvements. Still, it's not blocking for now.

Sounds good to use the `Wave` implementation for each transition. I'm gonna keep a try during the week. :+1:

IBAnimatable/ExplodeAnimator.swift
```diff
@@ -0,0 +1,94 @@
...
+  
+  public func animateTransition(transitionContext: UIViewControllerContextTransitioning) {
+    let (tempfromView, tempToView, tempContainerView) = retrieveViews(transitionContext)
+    guard let fromView = tempfromView, toView = tempToView, containerView = tempContainerView else {
+      transitionContext.completeTransition(true)
+      return
+    }
+    
+    containerView.addSubview(toView)
+    containerView.sendSubviewToBack(toView)
```
JakeLin: I think we can do `containerView.insertSubview(toView, atIndex: 0)`, but may not have any difference.

IBAnimatable/ExplodeAnimator.swift
```diff
@@ -0,0 +1,94 @@
...
+private extension ExplodeAnimator {
+  
+  func createSnapshots(toView toView: UIView, fromView: UIView, containerView: UIView) -> [UIView] {
+    let size = toView.frame.size
+    var snapshots = [UIView]()
+    let xFactor: CGFloat = 10.0
+    let yFactor = xFactor * size.height / size.width
+    
+    let fromViewSnapshot = fromView.snapshotViewAfterScreenUpdates(false)
+    for x in 0.stride(to: Int(size.width), by: Int(size.width / xFactor)) {
```
JakeLin: Shall we use `Double(size.width)` instead of `Int(size.width)` because `CGFloat` is more close to `Double` although width or height should not have decimal points.

IBAnimatable/ExplodeAnimator.swift
```diff
@@ -0,0 +1,94 @@
...
+        let xOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
+        let yOffset = self.randomFloatBetween(lower: -100.0, upper: 100.0)
+        view.frame = CGRectOffset(view.frame, xOffset, yOffset)
+        view.alpha = 0.0
+        
+        let angle = self.randomFloatBetween(lower: -10.0, upper: 10.0)
+        view.transform = CGAffineTransformScale(CGAffineTransformMakeRotation(angle), 0.1, 0.1)
+      }
+    }) { _ in
+      for view in snapshots {
```
JakeLin: Use `snapshots.forEach{ $0.removeFromSuperview() }` instead?

IBAnimatable/ExplodeAnimator.swift
```diff
@@ -0,0 +1,94 @@
...
+        snapshot.frame = snapshotRegion
+        containerView.addSubview(snapshot)
+        snapshots.append(snapshot)
+      }
+    }
+    return snapshots
+  }
+  
+  func animateSnapshotsExplode(snapshots: [UIView], completion: () -> Void) {
+    UIView.animateWithDuration(2, animations: {
```
JakeLin: 2 second is a little bit slow for me. .7 - .8 seems OK for me, but don't have strong opinion on it.

tbaranes: Ups, that's a debug variable, it should use `transitionDuration` instead.

IBAnimatable/PresentExplodeSegue.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Created by Tom Baranes on 03/04/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public class PresentExplodeSegue: UIStoryboardSegue {
+  public override func perform() {
+    destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Explode, interactiveGestureType: .Default)
```
JakeLin: This one should not have `interactiveGestureType`, it means it will use the default as `nil`

`destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Explode)`

JakeLin:
Except for few small comments, 👍 it is a great PR.

tbaranes:
Thanks for the feedback! I just commit the change following your comments.
Gonna work later on the designable part, and coming back to you :dancer:

JakeLin:
@tbaranes I have played around and made some changes, I also rebased from master. Please have a look at the last comments when you have time. Thanks.

IBAnimatable/PresentExplodeWithDismissInteractionSegue.swift
```diff
@@ -10,7 +10,7 @@ import UIKit
 
 public class PresentExplodeWithDismissInteractionSegue: UIStoryboardSegue {
   public override func perform() {
-    destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Explode)
+    destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Explode, interactiveGestureType: .Default)
     sourceViewController.presentViewController(destinationViewController, animated: true, completion: nil)
   }
 }
```
tbaranes, line 16: In that case, shouldn't we have it for the other Explode segue?

JakeLin, line 16: we can remove it now and add it back later when we have pinch gesture and want to use pinch to dismiss Explode animation. 

There are one or two types of segue for one type of animation, one is normal one without dismiss gesture. If this animation supports gesture, then we add another one called `****WithDismissInteractionSegue`, then the user can use gesture to dismiss when we developer uses `****WithDismissInteractionSegue`.

tbaranes:
Sorry, but did I miss a comment? 
I go ahead and try an implementation for passing params to an animator based on the `Mask` implementation. I don't think it has that much sense for the `ExplodeAnimator` to customise its xFactor and angle... but it's a first step of that kind of customisations for the animators.

JakeLin:
@tbaranes sorry, I overrode you commit, because I rebased from master and did a force push to `feature/explode_transition`. But don't worry, I think you can find it, please have a look at http://stackoverflow.com/questions/10099258/how-can-i-recover-a-lost-commit-in-git

tbaranes:
Are you sure about that? I don't see anything missing.

JakeLin:
@tbaranes I did found some commits before using those git commands even those were overridden by other developer's force push. I am so sorry to override your stuff. 😢, btw, I can't find them in my repo because you didn't push and I didn't fetch.

tbaranes:
Well, since I don't see anything missing in the code, I suppose it's ok :laughing: 
Also, it was working as expected yesterday, so...

JakeLin:
@tbaranes I have merged this PR first because I am going to do another PR, after that, I will work on Pinch gesture and add it to Explode transition. Thanks for your great work.

PR #158, opened 2016-04-03
Get release 2.2 ready

JakeLin:
We are going to release 2.2 🐣🐣🐣 
In this PR, we have **all** supported system built-in transition animations. Many thanks to @tbaranes

@tbaranes @seanwoodward  please have a look at the changes for `BorderDesignable` when you have time. refer to #150  Thanks

### DIFF
--- a/IBAnimatable/BorderDesignable.swift
+++ b/IBAnimatable/BorderDesignable.swift
@@ -25,22 +25,17 @@ public protocol BorderDesignable {
 
 public extension BorderDesignable where Self: UIView {
   public func configBorder() {
-    func clearBorders() {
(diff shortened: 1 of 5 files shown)


IBAnimatable/BorderDesignable.swift
```diff
@@ -25,23 +25,18 @@ public protocol BorderDesignable {
...
+
     guard let unwrappedBorderColor = borderColor else {
-      clearBorders()
       return
     }
     
     if borderWidth.isNaN || borderWidth <= 0 {
-      clearBorders()
       return
     }
```
tbaranes, line 34: I still think that we should remove that merge that `if` in the `guard` using `where`. It just making the code bigger.

tbaranes:
Just one comment! Otherwise, perfect :+1:

Issue #125, opened 2016-03-19
More interactive animators/controllers


JakeLin:
Currently, we have `PanInteractiveAnimator` to support custom transition animations for Push and Pop. We may add more interactive animators
- [x] `ScreenEdgePanInteractiveAnimator` 
- [x] `PinchPanInteractiveAnimator` 

Others like `Tap` (multiple times?) or `LongPress`? Feedback are welcome.

PR #160, opened 2016-04-07
Support PanScreenDdgeGesture for dismiss or pop

JakeLin:
In this PR
1. Refactor `InteractiveGestureType` to use more Swifty way. Remove `PanFromLeft`, `PanFromRight` and so on, and use `Pan(Left)` and `Pan(Right)` instead.
2. Create `InteractiveAnimator` abstract (kind of, but Swift doesn't have it) class to encapsulate all logic for `InteractiveAnimator` subclasses
3. Change the default transition duration to 0.75, it looks nicer for most of transition animations.
4. Add `ScreenEdgePanInteractiveAnimator` to support `PanFromRight`, `PanFromTop`, `PanFromBottom`, `PanHorizontally` and `PanVertically` to `ScreenEdgePan(Left)`, `ScreenEdgePan(Right)`, `ScreenEdgePan(Top)`, `ScreenEdgePan(Bottom)`, `ScreenEdgePan(Horizontal)` and `ScreenEdgePan(Vertical)` for `ScreenEdgePan` gesture transition controller.

It is very easy to add new `***InteractiveAnimator` like `PinchInteractiveAnimator` because most of logic is in `InteractiveAnimator` abstract class.

@tbaranes could you please have a look when you have time. I will add Pinch gesture to `ExplodeAnimator` in next PR, thanks.

Refer to #125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ All notable changes to this project will be documented in this file.
 
 #### API breaking changes
 
+- Change `PanFromLeft`, `PanFromRight`, `PanFromTop`, `PanFromBottom`, `PanHorizontally` and `PanVertically` to `Pan(Left)`, `Pan(Right)`, `Pan(Top)`, `Pan(Bottom)`, `Pan(Horizontal)` and `Pan(Vertical)` for `Pan` gesture transition controller.
+ 
 #### Enhancements
+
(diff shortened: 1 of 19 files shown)


tbaranes:
Awesome :+1:

JakeLin:
@tbaranes thanks for the review, I will work on Pinch gesture now.

Issue #66, opened 2016-02-08
Border with more than one side and less than 4 sides


JakeLin:
Currently, designable border can only supports one side or 4 sides.

JakeLin:
Close since #168 has merged. 😆

PR #162, opened 2016-04-08
Borders in Interface Builder and multiple sides

seanwoodward:
Additional fix for border display in Interface Builder. Also add support for multiple sides per enhancement issue #66.


seanwoodward:
Initial visual tests worked perfectly. Adding code to other projects results in unexpected behaviors. Will fix and resubmit.

PR #163, opened 2016-04-08
Border dispaly in Interface Builder and Multiple Sides

seanwoodward:
dditional fix for border display in Interface Builder. Also includes support for multiple sides per enhancement issue #66.


seanwoodward:
This one should work better.

IBAnimatable/BorderDesignable.swift
```diff
@@ -25,39 +25,70 @@ public protocol BorderDesignable {
...
     
-    if let unwrappedBorderSide = borderSide, side = BorderSide(rawValue: unwrappedBorderSide) {
-      configBorderWithSide(side, borderColor: unwrappedBorderColor)
-    } else {
+    if borderWidth.isNaN || borderWidth <= 0 {
+      clearBorders()
+      return
+    }
+    
+    clearBorders()
```
JakeLin, line 45: Seems you have merged the changes from upstream master in https://github.com/JakeLin/IBAnimatable/pull/163/commits/b0bbf139c3efb2029f39975969230a8f2d170432 but don't know why our changes didn't apply to your branch. Because we call `clearBorders()` unconditionally. The code in master would be slightly better.

```
    // Clear borders
    layer.sublayers?.filter  { $0.name == "borderSideLayer" || $0.name == "borderAllSides" }
        .forEach { $0.removeFromSuperlayer() }      

    guard let unwrappedBorderColor = borderColor where borderWidth > 0 else {
      return
    }
```

can you rebase again? may need to add the code below back. 

```
layer.borderColor = nil
layer.borderWidth = 0
```

IBAnimatable/BorderSide.swift
```diff
@@ -11,3 +11,49 @@ public enum BorderSide: String {
...
+      return
+    }
+    
+    self = BorderSides(sides)
+    
+  }
+  
+  init(side: BorderSide?) {
+    guard let side = side else { 
+      self = .None
```
JakeLin, line 48: Should the default value be  `. AllSides` because we display all sides when we don't specify any. That's the currently behaviour.

JakeLin:
@seanwoodward Thank you for your PR. I like the idea because sometimes, we need to have more than one sides. Can you please have a look at my comments? I give it a 👍. 

I will update the CHANGELOG.md to reflect that changes once we merge in.

seanwoodward:
I'll redo and re-submit.

PR #164, opened 2016-04-09
Fold transition

tbaranes:
Here is the next one: Fold transition, related to #155 

That PR implements the `Fold` transition, it's working as expected, and can be customise with using params: `Fold(nbFolds)` which will change the number of pan during the transition, default is 2.

Missing:
- Reverse: I didn't find a way to detect if the transition is a push or a pop. Did you implement a system for that? The navigation is currently in hard coding (a private property set to true)
- Direction: not sure if that's needed, but after some tests I think it could be nice to setup the direction. Moreover, it could be used to solve the reverse issue if there no existing code to detect the that. 

The main code won't move (at least, if we dont implement the direction which is way more complex 😆) that much, so if you have any feedback, let me know 👐 

/!\ This PR is not finished yet, don't merge it /!\

### DIFF
--- a/IBAnimatable/FoldAnimator.swift
+++ b/IBAnimatable/FoldAnimator.swift
@@ -12,9 +12,11 @@ public class FoldAnimator: NSObject, AnimatedTransitioning {
   public var reverseAnimationType: TransitionAnimationType?
   public var interactiveGestureType: InteractiveGestureType? = .Default
   
-  // MARK: - Private
+  // MARK: - Private params
   private var reverse: Bool = false
   private var folds: Int = 2
+  
+  // MARK: - Private fold transition
   
   private var transform: CATransform3D = CATransform3DIdentity
   private var size: CGSize = CGSize.zero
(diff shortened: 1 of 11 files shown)


JakeLin:
@tbaranes thanks for the PR, great one 😎

Currently, the way to support reverse is to use direction eg. left vs right, in vs out. Because if we can support reverse transition effect, we should be able to expose the parameter, in `FoldAnimator` we can support `FoldAnimator(.Right, nbFolds)` and `FoldAnimator(.Left, nbFolds)`

 `FoldAnimator(.Right, nbFolds)`  will set `reverse = false` and  `FoldAnimator(.Right, nbFolds)`  will set `reverse = true`

Please let me know if you have any questions.

tbaranes:
👍 Will work on it next week!

tbaranes:
Here a small update:
- Small refactoring of the `AnimationTransitonType` to handle default direction, and direction + params
- The fold now handles direction to manage the `revere` animation

Still missing: vertical fold animations, it's only working with the horizontal one

Not sure about the buddy build failure, do I have something to do?

tbaranes:
The vertical animation is here! Ready to merge from my side, what do you think about it @JakeLin ? 🎉

IBAnimatable/FoldAnimator.swift
```diff
@@ -0,0 +1,254 @@
...
+  
+  func valuesForAxe(initialValue: CGFloat, reverseValue: CGFloat) -> (CGFloat, CGFloat) {
+    return horizontal ? (initialValue, reverseValue) : (reverseValue, initialValue)
+  }
+  
+  func width() -> CGFloat {
+    return horizontal ? size.width : size.height
+  }
+  
+  func height() -> CGFloat {
```
JakeLin: can them be properties?

JakeLin:
@tbaranes  well done, 👍 vertical animations fit most of the use cases. please get it in when you free ready. Thanks.

tbaranes:
Thanks Jake! If travis build passed, I will merge it 👯

JakeLin:
@tbaranes thanks again, I have fixed buddybuild by uploading the provisioning profile and p12 files. It is green again.

PR #166, opened 2016-04-10
Update Functions.swift

TheDarkCode:
Slightly faster version.

### DIFF
--- a/IBAnimatableApp/Functions.swift
+++ b/IBAnimatableApp/Functions.swift
@@ -8,28 +8,14 @@ import Foundation
 // Source: https://gist.github.com/TheDarkCode/2f65c1a25d5886ed210c3b33d73fe8a9
 // Based on earlier version: http://stackoverflow.com/a/28341290/749786
 public func iterateEnum<T: Hashable>(_: T.Type) -> AnyGenerator<T> {
-    
     var x = 0
-    
     return AnyGenerator {
-        
         let y = withUnsafePointer(&x) {
-            
             UnsafePointer<T>($0).memory
-            
         }
-        
-        if y.hashValue == x {
-            
+        defer {
             x += 1
-            return y
-            
-        } else {
-            
-            return nil
-            
         }
-        
+        return next.hashValue == x ? y : nil
     }
-    
 }
--- a/IBAnimatableApp/Functions.swift
+++ b/IBAnimatableApp/Functions.swift
@@ -1,17 +1,35 @@
 //
-//  Created by Jake Lin on 3/1/16.
-//  Copyright © 2016 Jake Lin. All rights reserved.
+//  Created by Mark Hamilton on 4/9/16.
+//  Copyright © 2016 dryverless. All rights reserved.
 //
 
 import Foundation
 
-// http://stackoverflow.com/a/28341290/749786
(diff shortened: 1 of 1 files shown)


JakeLin:
@TheDarkCode thanks for your PR, do you mean we can save one times `i = i + 1`?  I am OK with your PR, before I can get it into master, please use two spaces as a tab and remove some empty lines, we try to follow this code style. https://github.com/raywenderlich/swift-style-guide , thanks.

TheDarkCode:
I'll make the changes in the morning and resubmit. The line that I removed, which saves time very slightly, is: `let currentI = i`.

Another way of doing it is keeping what the original had for the most part, but making an extension to Int which defers the postfix increment until after the initial value is returned to the ternary operator.

``` swift
func iterateOverEnum<T: Hashable>(_: T.Type) -> AnyGenerator<T> {
    var i = 0
    return AnyGenerator {
        let next = withUnsafePointer(&i) { UnsafePointer<T>($0).memory }
        return next.hashValue == i.postfix(1) ? next : nil // or i.postfixIncrement()
    }
}

public extension Int {
    mutating func postfixIncrement() -> Int {
        defer {
            self += 1   
        }
        return self   
    }

    mutating func postfix(increment: Int) -> Int {
        defer {
            self += increment   
        }
        return self   
    }
}
```

JakeLin:
thanks @TheDarkCode it saves a local variable and one assignment, I like you solution better, don't like the extension to `Int` too much because it is too wide to all `Int`. But good to know this powerful and fun extension. To do something similar to the extension, Can we use `defer` here?

```
defer {
  i += 1
}
return next.hashValue == i ? next : nil
```

tbaranes:
Just a question... is there any reason to keep that function since we don't use it anymore?

Otherwise, nice PR 😆

JakeLin:
@tbaranes yes, because we can have more demo scenes like transition animations ViewController in the future to demo more features like `Animations`, `PredefinedGradient`, `PredefinedColor` and so on.

I would like to keep this `func` in `IBAnimatableApp` target only because `IBAnimatable` would not use it.

TheDarkCode:
@JakeLin In regards to the `defer` inside `AnyGenerator`, yes, that works fine. Also, I'm actually rebuilding your entire project as part of an SDK I'm writing, so we should chat at some point and see where the directions mesh.

@tbaranes The function is actually super useful for lots of things so while it might not be so obvious here, it makes a huge difference across projects when you are dealing with lots of enums. Like if you want to make helper functions that return all possible error types, or all options for whatever, that function becomes critical.

I uploaded a playground which uses all the different versions [here](https://github.com/TheDarkCode/Example-Swift-Apps/tree/master/Playgrounds/EnumExtensions.playground)

JakeLin:
@TheDarkCode cool, are you going to use `defer`? if yes, please let me know once you push the changes.

Sure, love to chat with you, and looking forward to hearing more from your SDK.

TheDarkCode:
@JakeLin Done. Sent you an e-mail with my contact info as well.

~~Opps, looks like that version still has 4 spaces not 2.~~ Fixed.

JakeLin:
@TheDarkCode thanks for your PR.

PR #168, opened 2016-04-11
Border Fix and Multiple Sides

seanwoodward:
Corrects border behavior and adds support for multiple sides (enhancement issue #66).
A better approach perhaps. BorderSides is implemented in BorderDesignable.swift as it is used only by the configBorder() function. Border configuration is consolidated to the configBorder() function to make logic of the three basic borders (masked, all sides, or multi-sided) easier to follow.

Close #162 #163


JakeLin:
Very cool, I like it, 👍 thanks @seanwoodward

PR #171, opened 2016-04-17
Portal animator

tbaranes:
Next animator: Portal #155 

Here what's new:
- Two new directions: `Forward` and `Backward`. All the others animators will now used `Left` by default, and ignore these two. They are useful for the portal animation since `Left`... doesn't really make sense for that kind of animation
- PortalAnimator: classic usage, support one parameter: zoomScale
- Portal segue, as usual

Ready to go for me 👍

### DIFF
--- a/IBAnimatable/TransitionFromDirection.swift
+++ b/IBAnimatable/TransitionFromDirection.swift
@@ -12,6 +12,8 @@ public enum TransitionFromDirection {
   case Right
   case Top
   case Bottom
+  case Forward
+  case Backward
   
   var stringValue: String {
     switch self {
@@ -25,6 +27,8 @@ public enum TransitionFromDirection {
     case .Bottom:
       // The actual transition direction is oposite, need to reverse
       return kCATransitionFromTop
+    default:
+      return ""
     }
   }
 }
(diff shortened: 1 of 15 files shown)


IBAnimatable/TransitionFromDirection.swift
```diff
@@ -25,6 +27,8 @@ public enum TransitionFromDirection {
     case .Bottom:
       // The actual transition direction is oposite, need to reverse
       return kCATransitionFromTop
+    default:
+      return ""
```
JakeLin, line 31: I was thinking `fatalError`, but I think returning `kCATransitionFromLeft` as the default value is not a bad idea, how do you think @tbaranes ?

JakeLin:
I like the idea of using `Left` as default . Great job 👍 thanks.

IBAnimatable/TransitionFromDirection.swift
tbaranes, line 31: Not sure that makes sense, I mean `Forward` and `Backward` direction will never use any strings since its just for CA. So, returning `Left` by default can be confusing, isn't it?

JakeLin, line 31: @tbaranes I think this property `stringValue` is  only used in `CATransition` and will not fall into `default` at all. Let's leave it `""`. Please merge it when you are ready, thanks.

PR #175, opened 2016-04-21
Fix a bug for fold animator with pan gesture when dismiss

JakeLin:
Fix a bug for fold animator with pan gesture when dismiss the ViewController, it is a bug in https://github.com/ColinEberhardt/VCTransitionsLibrary as well, will do a PR to them to fix it.

It was OK for popping a ViewController, but not for dismissing a ViewController, the transition state was broken.

The fix is in 
https://github.com/JakeLin/IBAnimatable/compare/bugfix/fold-animator-with-pan-gesture-when-dismiss?expand=1#diff-e3c0877d13bebebec521d0689eb955f7R97

Also fixed some minal code style issues.

@tbaranes can you have a look at this? thanks.

### DIFF
--- a/IBAnimatable/BorderDesignable.swift
+++ b/IBAnimatable/BorderDesignable.swift
@@ -138,7 +138,7 @@ public extension BorderDesignable where Self: UIView {
       borderPath.addLineToPoint(linePoints.end)
     }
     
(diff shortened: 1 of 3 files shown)


tbaranes:
Seems good 👍

PR #180, opened 2016-04-26
Fix a bug to support single side of border for AnimatableTextField

### LINKED ISSUE #179: Issue with AnimatableTextField
After I updated pod from 1.0.1 to latest 2.2 version, borderSide doesn't work for me any longer in IB.   If I set borderWidth to 1, borderColor to grey and borderSide to Bottom.   I get border on all 4 sides regardless.   If I revert back to 1.0.1 it works again.   Any ideas?

JakeLin:
Should fix #179 

To fix this issue, we need to set `borderStyle = .None` for `UITextField`

Also, I try to fix a warning for `BorderDesignable` for `AnimatableStackView` - "CATransformLayer changing property borderColor in transform-only layer, will have no effect" in this PR.

Actually, `BorderDesignable` doesn't work for `AnimatableStackView`, I think this is a bug of `UIStackView` someone tweeted about it https://twitter.com/fra3il/status/654127926644899841 But we can wait for the comming WWDC to see whether Apple fix it.

@tbaranes can you have a look if you have time, thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to this project will be documented in this file.
 
 - Use `weak` for the `viewController` in `InteractiveAnimator` to avoid retain cycle.
 - Fixed the right image of `SideImageDesignable` [#176](https://github.com/JakeLin/IBAnimatable/issues/176)
-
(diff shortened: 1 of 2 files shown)


tbaranes:
👍 and thanks for the warning! It was annoying me 😆

PR #184, opened 2016-05-01
Turn transition

tbaranes:
Related to #155

New transition from VCTransitionsLibrary `Turn`:
- Take a direction
- Default gesture is the Pan (not sure about this, but why not)
- No custom parameters

It's very similar to the `SystemFlip` but seems to be a bit smoother and nicer. So I think we can let the both and let the user choose.

Ready to be merged 🎉


JakeLin:
@tbaranes thank you very much for this PR, 👍 I think it is better than `SystemFlip`, `System*` transitions give us a chance to implement them quickly, but custom transition animations is nicer. 😘

I like the `Pan` gestures for `Turn` too.

I think there is a typo in README.md for `Portal(Left`, `Portal` support `Forward` and `Backward` as paramter. I just fix it in master.

PR #189, opened 2016-05-06
Initialise animator outside of the framework by making their init public

tbaranes:


JakeLin:
👍 forgot to make them `public` 😓

tbaranes:
I forgot that too 😆

PR #190, opened 2016-05-06
Remove unused parameters in Presenter

tbaranes:
Don't see any regression, is there a reason for these two parameters?


JakeLin:
👍  please merge it.

Issue #188, opened 2016-05-05
Can we use animators with child view controllers?


tbaranes:
In my current app, I'd like to use the `Fold` animations when transitioning between two child view controllers. From what I know about the animators, I don't think we can. For example, for a fade transition, I will have the following code (approximately):

``` swift
        oldViewController.willMoveToParentViewController(nil)
        addChildViewController(newViewController)
        addSubview(newViewController.view, toView:self.containerView!)
        newViewController.view.alpha = 0
        newViewController.view.layoutIfNeeded()
        UIView.animateWithDuration(animationDuration, animations: {
            newViewController.view.alpha = 1
            oldViewController.view.alpha = 0
        }, completion: { finished in
            oldViewController.view.removeFromSuperview()
            oldViewController.removeFromParentViewController()
            newViewController.didMoveToParentViewController(self)
        })
```

But that's a bit sad to have to rewrite it since we already did it in `FadeAnimator`, or if I wanted to use a more complex transition... any suggestions about this?

That issue may be related to #187

JakeLin:
@tbaranes Have you seen this article before, https://www.objc.io/issues/12-animations/custom-container-view-controller-transitions/ it looks like something similar to your requirements. Because all `***Animator`s conform to `UIViewControllerAnimatedTransitioning` it will work with the container ViewController uses `- (id <UIViewControllerAnimatedTransitioning>)containerViewController:(ContainerViewController *)containerViewController animationControllerForTransitionFromViewController:(UIViewController *)fromViewController toViewController:(UIViewController *)toViewController;`

tbaranes:
Right, I'm gonna give a try. Thanks!

tbaranes:
I finally get it working using your link, thanks to you. I will make a PR, I think it could be nice to have that feature in IBAnimatable.

PR #192, opened 2016-05-06
Support for container transition

tbaranes:
Related to #188 

This PR contains only one new class: `Container` (need to find a better name) allowing the usage of `IBAnimatable` animators to make transition between childViewController. The usage is really straightforward:

``` swift
let transitionContext = ContainerTransition(animationType: .Fold(direction: direction, params: []),
                                                    container: containerView,
                                                    parentViewController: self,
                                                    fromViewController: fromViewController,
                                                    toViewController: toViewController) {
      self.currentViewController = toViewController
}
transitionContext.animate()
```

More details:
- Doesn't support interactive gesture
- Doesn't support for interface builder
- If nil is passed as `fromViewController`, it will set the `toViewController` as default without animation
- Handle all the logic to transit from a view controller to a new one, with an animation
- Support every `TransitionAnimationType`

That feature doesn't really follow the main goal of `IBAnimatable` (a library for designers), but I'm sure it's worth it for every developers using it programatically.  What do you think?

If you are ok to merge it, I will update the CHANGELOG, README, and also add an example in the demo project.

### DIFF
--- a/IBAnimatable/Container.swift
+++ b/IBAnimatable/Container.swift
@@ -1,0 +1,110 @@
+//
+//  Created by Tom Baranes on 06/05/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+class ContainerTransition: NSObject, UIViewControllerContextTransitioning {
+  
+  typealias ContainerTransitionCompletion = () -> Void
+  
+  // MARK: Properties
+  
+  var animationType: TransitionAnimationType?
+  var transitionDuration: Duration = defaultTransitionDuration
+  
+  // MARK: Private
+  
+  private let container: UIView?
+  private let parentViewController: UIViewController?
+  private var viewControllers: [String: UIViewController]? = nil
+  private var views: [String: UIView]? = nil
+  private let completion: ContainerTransitionCompletion?
+  
+  // MARK: Life cycle
+  
+  init(animationType: TransitionAnimationType,
+       container: UIView, parentViewController: UIViewController,
+       fromViewController: UIViewController?, toViewController: UIViewController,
+       completion: ContainerTransitionCompletion? = nil) {
+    
+    self.completion = completion
+    self.animationType = animationType
+    self.container = container
+    self.parentViewController = parentViewController
+    
+    container.translatesAutoresizingMaskIntoConstraints = false
+    toViewController.view.translatesAutoresizingMaskIntoConstraints = true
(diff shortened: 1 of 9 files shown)


IBAnimatable/Container.swift
```diff
@@ -0,0 +1,110 @@
...
+    
+    parentViewController?.view.userInteractionEnabled = false
+    let animator = AnimatorFactory.generateAnimator(unwrappedAnimationType)
+    animator.animateTransition(self)
+  }
+}
+
+// MARK: - UIViewControllerContextTransitioning
+
+extension ContainerTransition {
```
JakeLin: Can we move `UIViewControllerContextTransitioning` in this `extension`?

IBAnimatable/Container.swift
```diff
@@ -0,0 +1,110 @@
+//
+//  Created by Tom Baranes on 06/05/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+class ContainerTransition: NSObject, UIViewControllerContextTransitioning {
```
JakeLin: Does it need to be `public`?

JakeLin:
@tbaranes thanks for your PR, I don't have better idea to call it, may call `ContainerControllerAnimatedTransition` or just `ContainerTransition`? it is your call to name it. 

IB doesn't support ContainerView either, and we have to use code to support / switch multiple viewControllers, I actually like this PR because it can enable all `***Animator`s to use in sub controllers, not only for Push/Pop and Present/Dismiss transitions. That one principle when we architectured our Transition Animators / Controllers, they are all standard transition controllers conforming to `UIViewControllerAnimatedTransitioning`, we can use them in Apple's standard way for transitions.

tbaranes:
Happy to see you like it! For the name, I was thinking about `ContainerTransition` too, so it will probably be the final name. 

l will add an example, and update the docs this weekend 🎉

JakeLin:
cool, thanks @tbaranes  🎉

tbaranes:
Update of this PR:
- Add a `UITabBar` using `ContainerTransition` example, it's accessible from the transitions tableView. Each `UITabBarItem` will be show with different animation: Explode, Fold, Portal, and NatGeo
- Update the CHANGELOG
- Didn't update the README since we are going to change its architecture

Ready to go 🎉

JakeLin:
Cool, please merge it. 👍

PR #195, opened 2016-05-07
Fix segue transitions

### LINKED ISSUE #194: Present***Segue are not working anymore
I don't know if I'm missing something, or if that is a regression, but I can't get working the `Present***Segue` in storyboard. The animation is now a classic modal transition instead of an Explode...

Few days ago, it was still working in my project, but I realise yesterday that wasn't working anymore. I started to look deeper, it seems that the Presenter and the Animator are well initialised, but the `UIViewControllerTransitioningDelegate` is never called.

Step to reproduce it: just change the animation of one of the segue in the demo project.

Anyway, if that a real bug, we have to fix it before the next release!

tbaranes:
Finally, the clean code from yesterday wasn't a good idea, didn't notice that was a method from the delegate. My bad 😆 
Hopefully, that wasn't released.

Close #194

### DIFF
--- a/IBAnimatable/Presenter.swift
+++ b/IBAnimatable/Presenter.swift
@@ -74,7 +74,7 @@ public class Presenter: NSObject {
 
 extension Presenter: UIViewControllerTransitioningDelegate {
   // MARK: - animation controller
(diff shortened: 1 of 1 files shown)


JakeLin:
Cool, get it in master please. 👍

PR #196, opened 2016-05-07
Pinch interactive animator/controller

### LINKED ISSUE #125: More interactive animators/controllers

JakeLin:
- Add Pinch Close/Open gestures for Pop/Dismiss.
- Set Pinch gesture as default `interactiveGestureType` for `SystemCameraIrisAnimator` and `PortalAnimator` 
- Fix a bug for `PortalAnimator` when use it with interactive gesture `fromView.removeFromSuperview()` will cancel the `InteractiveAnimator`, so hide `fromView` during the transition and make it visible when the transition finishes. Have to use `print()` logs to find the problem, it took me days to find out and fix it 😓 . Will PR back to `VCTransitionsLibrary` later
- Set `Backward` as default for `PortalAnimator` since `Portal(Forward)` will crash tab bar animation. It didn't crash because in master, it calls `executeBackwardAnimations` if we don't set the direction like `"Portal"` only.

@tbaranes please have a look, thanks. I think we are ready for 2.3 after this PR. 🎊

should close #125


IBAnimatable/SystemCameraIrisAnimator.swift
```diff
@@ -26,12 +23,15 @@ public class SystemCameraIrisAnimator: NSObject, AnimatedTransitioning {
     case .Open:
       self.transitionAnimationType = .SystemCameraIris(hollowState: .Open)
```
tbaranes, line 24: Should we use the new directions `Open` and `Close` instead of hollowState ?

tbaranes:
Just one question about the `HollowState(Open/Close)` versus `Direction(Open/Close)`, otherwise, seems perfect, and nice catch for the Portal.

Let's move forward by releasing 2.3 👍

IBAnimatable/SystemCameraIrisAnimator.swift
JakeLin, line 24: @tbaranes yes, I will change it to make it more consistent, thanks for picking up this.

JakeLin:
@tbaranes I merge it first, I have looked at `SystemCameraIrisAnimator` and think it should not have any knowledge about  `GestureDirection` because it is not a gesture animator/controller. 

Since 2.3 is a big release, I am going to write some documents before the release. Will start with `Transitions.md` based on your great suggestions. 😄

PR #197, opened 2016-05-08
Add SwiftLint

tbaranes:
Since I started to port the transition from `VCTransitionsLibrary` to swift, I noticed that it's really easy to miss a semicolon when copy / paste, or let a `CGRectOffset` instead of using swift geometry initialiser... That's why I think it could be worth to use [SwiftLint](https://github.com/realm/SwiftLint) in `IBAnimatable` in order to have a more consistent coding style.

In this PR:
- New script phase to run [SwiftLint](https://github.com/realm/SwiftLint) -> if SwiftLint is not installed, it will do nothing, otherwise, you may have some warnings / errors if you doesn't respect all the rules. The script has been added only for the framework, not the demo app
- Fixed all the warnings / errors generated by SwiftLint (I never noticed the `.offsetBy` before...)
- Add a [.swiftlint.yml](https://github.com/JakeLin/IBAnimatable/blob/feature/swiftlint/.swiftlint.yml) to disable some rules. In a long term process, I think we could enable most of them to have a cleaner code, but it will ask some refactors. Also, if we enable them, we will probably need to arrange the rules following our needs

It may looks a bit annoying, but I'm already using it in all my projects, and it's really useful to be sure that we don't miss a semicolon, a geomtry swift initialiser, and so many other cases.

@JakeLin I let you the final choice to merge or not this PR depending of if you want to use that library in IBanimatable!


JakeLin:
@tbaranes please merget it once the build is green, I was doing the rebase and saw your last two commits 👍. SwiftLint is a good tool to improve the code quality, I will try to remove some items from `.swiftlint.yml` to make the code better.

PR #199, opened 2016-05-08
Slide transition

tbaranes:
We did complex and advanced transition, but forgot to do the most basic one: Slide!
- New animator: SlideAnimator
- Support for the both directions: Left, Right, Top, Bottom
- Take one custom parameter: "fade" in order to add a fade animation to the slide transition
- Support interactive gesture, the default one is Pan
- Refactored a bit `TransitionFromDirection` to detect directly if the direction is horizontal or not

About the interactive gesture, the Pan is reversed. Is there a reason about this?

Related to #155

### DIFF
--- a/IBAnimatable/SlideAnimator.swift
+++ b/IBAnimatable/SlideAnimator.swift
@@ -28,22 +28,22 @@ public class SlideAnimator: NSObject, AnimatedTransitioning {
     case .Right:
       self.transitionAnimationType = .Slide(direction: .Right, params: params)
       self.reverseAnimationType = .Slide(direction: .Left, params: params)
-      self.interactiveGestureType = .Pan(direction: .Right)
+      self.interactiveGestureType = .Pan(fromDirection: .Right)
       reverse = true
     case .Top:
       self.transitionAnimationType = .Slide(direction: .Top, params: params)
       self.reverseAnimationType = .Slide(direction: .Bottom, params: params)
-      self.interactiveGestureType = .Pan(direction: .Top)
+      self.interactiveGestureType = .Pan(fromDirection: .Top)
       reverse = false
     case .Bottom:
       self.transitionAnimationType = .Slide(direction: .Bottom, params: params)
       self.reverseAnimationType = .Slide(direction: .Top, params: params)
-      self.interactiveGestureType = .Pan(direction: .Bottom)
+      self.interactiveGestureType = .Pan(fromDirection: .Bottom)
       reverse = true
     default:
       self.transitionAnimationType = .Slide(direction: .Left, params: params)
       self.reverseAnimationType = .Slide(direction: .Right, params: params)
-      self.interactiveGestureType = .Pan(direction: .Left)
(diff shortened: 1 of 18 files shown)


tbaranes:
My bad about the pan gesture, it seems to be have the right behaviour following its setup, but didn't we reverse all the interactive gesture in each transition? I mean, shouldn't we set `Left` when the direction is from `Left`. Currently, we are setting it to Right for Left

EDIT: In fact, whichever direction I set to the pan, I get the same gesture. Definitely something weird

IBAnimatable/SlideAnimator.swift
```diff
@@ -0,0 +1,98 @@
...
+  
+  func animateSlideTransition(toView toView: UIView, fromView: UIView, travel: CGAffineTransform, completion: AnimatableCompletion) {
+    UIView.animateWithDuration(transitionDuration, animations: {
+      fromView.transform = travel
+      toView.transform = CGAffineTransformIdentity
+      if self.fade {
+        fromView.alpha = 0
+        toView.alpha = 1
+      }
+    }) { _ in
```
JakeLin: I try to follow the code style from https://github.com/raywenderlich/swift-style-guide#closure-expressions , it help me understand the closure better. How do you think @tbaranes ?

JakeLin:
@tbaranes thanks for your PR. I am OK except for the gesture, I will take a deep look at the gesture tomorrow.

IBAnimatable/SlideAnimator.swift
tbaranes: I'm not used about this syntax (I'm generally following apple / github code style) but it's your call 👍

JakeLin:
@tbaranes I think the Pan gesture is OK, just change the gesture direction in fb788f0 , to make `Pan` and `ScreenEdgePan` more specific, I have refactored the input parameter from `direction` to `fromDirection`. Because they were new to 2.3, no breaking changes 😁 Please have a look and merge it if you feel ready.

I am working on the improvement of the demo app  for transitions. e.g. adding title for the transition. It will help us write the document.

tbaranes:
Excellent idea to change the name, it will be clearer like this.
Shouldn't we change the gesture direction in the others Animators (SystemMove, Fade...)?

JakeLin:
@tbaranes I am not sure which gesture direction are you talking about? do you mean to swap `.Left` with `.Right`?

tbaranes:
Yes, right now, the `SlideAnimator` doesn't have the same pan gesture direction as the other animators using that same interactive gesture (Fade...) which, I think, doesn't make sense.

JakeLin:
@tbaranes because all of system built-in transition animations are `From` direction (e.g. `UIViewAnimationTransitionFlipFromLeft`, `UIViewAnimationTransitionFlipFromRight`), I am thinking change their input parameter to `fromDiretion` instead of `direction`. How do you think?

But for your PR, please merge it when you think it is ready, thanks 👍

tbaranes:
Agree for the naming, that will improve the readability.

PR #202, opened 2016-05-10
Makes AnimatableNavigationController configurable after init

tbaranes:
The `Navigator` will be create again with the new parameters if we change the animation type, the duration, or the interactive gesture. That can be useful in case that we are using only one navigation controller but want to customise the animation.. between two push / pop. If the animation is set to nil, then the custom navigation is completely reset and back to the default behaviour.

It wasn't what I was looking at first, but I'm pretty sure that can be pretty useful in a lot of cases. I'm just wondering, if there any reason that we create custom segues for Present / Dismiss but not Push / pop?

### DIFF
--- a/IBAnimatable/AnimatableNavigationController.swift
+++ b/IBAnimatable/AnimatableNavigationController.swift
@@ -8,22 +8,36 @@ import UIKit
 public class AnimatableNavigationController: UINavigationController, TransitionAnimatable {
   
   // MARK: - TransitionAnimatable
-  @IBInspectable public var transitionAnimationType: String?
-  @IBInspectable public var transitionDuration: Double = .NaN
-  @IBInspectable public var interactiveGestureType: String?
-
+  @IBInspectable public var transitionAnimationType: String? {
+    didSet {
+      configureNavigationControllerDelegate()
+    }
+  }
+  @IBInspectable public var transitionDuration: Double = .NaN {
+    didSet {
+      configureNavigationControllerDelegate()
+    }
+  }
+  @IBInspectable public var interactiveGestureType: String? {
+    didSet {
+      configureNavigationControllerDelegate()
+    }
+  }
+  
   // MARK: - Lifecylce
   public override func viewDidLoad() {
     super.viewDidLoad()
     configureNavigationControllerDelegate()
   }
-
+  
   // MARK: - Private
   // Must have a property to keep the reference alive because `UINavigationController.delegate` is `weak`
   private var navigator: Navigator?
-
+  
   private func configureNavigationControllerDelegate() {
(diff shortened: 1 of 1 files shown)


JakeLin:
@tbaranes Good idea to make it configurable after `init` 👍 , The reason why we don't have Push / Pop custom segues is in most of cases, all `Push`s within the same one `UINavigationController` should have the same transition animation. If we change it, it will impact on the others, it maybe not what the user wants. 

For example, VC1 pushs to VC2 using `Fade` animation, and if VC2 pushs to VC3 using `Slide` animation, then VC3 pops to VC2 is `Slide` (what the user expected), but VC2 pops to VC1 is also `Slide` not 'Fade`because`UINavigationController.delegate`has been changed to`SlideAnimator`not`FadeAnimator`. I think that's not the user wants.

That maybe an Apple's architure issue, or intentionally, they want to have consistent animations within the same `UINavigationController`. If we support `***PushSegue`s, I don't know how to fix this scenario, but if the user uses code, they can fix it by using Apple default way like overriding `func navigationController(navigationController: UINavigationController, animationControllerForOperation operation: UINavigationControllerOperation, fromViewController fromVC: UIViewController, toViewController toVC: UIViewController) -> UIViewControllerAnimatedTransitioning?` and returns the corresponding animator based on the scenario. 

Therefore, to not confuse the user when they configure Push / Pop transition animations in Interface Builder, we don't support `***PushSegue`, if the user wants to change the animation in code, they can still use the transition animators in `IBAnimatable` and override `func navigationController(navigationController: UINavigationController, animationControllerForOperation operation: UINavigationControllerOperation, fromViewController fromVC: UIViewController, toViewController toVC: UIViewController) -> UIViewControllerAnimatedTransitioning?`

tbaranes:
hmmm... yeah, I understand your point of view. In fact, I was exactly in that weird case: My first push was a `Fade`, some normal one, then an explode, then normals. That's why a push segue would have make sense for me since I needed a classic transition in most of case beside two of them. Nevertheless, you are right that can make give a weird behaviour.

I merged that PR which helps to solve that issue, the segue was easier to do since we just had to override and doesn't ask any code, but by configure again each animation before pushing, I can handle them as I want: make it Fade, make it Explode, make it nothing special before each transition.

Anyway, thanks for your explanation 👍

JakeLin:
@tbaranes thanks for your PR. I hope it can still fit your requirements. I found out it is very challenging to design APIs for framework but I am enjoying doing that because we can learn from that process.

tbaranes:
Agree, that a nice job to design the API! I can get my needs with that PR, it's more code, but I'm ok with it 👯

PR #208, opened 2016-05-22
Correct default parallelogram angle

lastMove:
The default Parallelogram angle is now effectively 60 degrees (as the doc says)


JakeLin:
👍  thanks @lastMove 

BTW: I have updated the [Roadmap.md](https://github.com/JakeLin/IBAnimatable/blob/master/Documentation/Roadmap.md) to include `Parallelogram` mask.

PR #204, opened 2016-05-17
IBAnimatable Playground - Transition Animations

JakeLin:
We have created a new playground to demonstrate the IBAnimatable features, currently, we can demo Transitions and Interactions, other features (UI and Animations) are coming.

Transition Animations  

In Transitions playground, we can see
1. all supported transitions
2. how they look like in Push/Pop and Present/Dismiss transitions
3. how supported segues with/without dismiss interaction works

In Interactions playground, we can see custom container transitions.

@tbaranes please have a look when you have time.

PR #213, opened 2016-05-31
Segue Completion Handlers

TheDarkCode:
I understand that you can select the animation segue via `transitionAnimationType` inspectable on `AnimatableViewController`, but there is currently no way to deal with an overriding `perform()` function on a custom `UIStoryboardSegue` that allows you to access the completion handler on `presentViewController` calls from the `sourceViewController`.

This pull request adds `OptionalAnimatableCompletion` as `(() -> Void)?` that is used in the changes. 

There is redundancy in the edits and so the final changes would obviously be better using the override on `perform()` with a call to the public method that has the completion handler parameter and just passing nil.

This comes into play if you want to do things using the callback after the `viewDidAppear: method is called on the presented view controller`.


IBAnimatable/DismissSegue.swift
```diff
@@ -1,5 +1,6 @@
 //
 //  Created by Jake Lin on 12/14/15.
+//  Updated by Mark Hamilton on 5/30/16
```
JakeLin, line 3: We don't usually because git keeps all history.

TheDarkCode, line 3: That's fine, I can remove it from the files.

IBAnimatable/PresentExplodeSegue.swift
```diff
@@ -10,4 +11,8 @@ public class PresentExplodeSegue: UIStoryboardSegue {
     destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Explode(params: []))
     sourceViewController.presentViewController(destinationViewController, animated: true, completion: nil)
   }
+  public func perform(completion: OptionalAnimatableCompletion) {
```
JakeLin, line 14: I am not sure will we use Segue programmatically like this? Because Apple has default way to use Transition Controllers / Animators which doesn't require a Segue object. `IBAnimtable` provides Segues because we want to provide a convenience way for the user to configure present transition in Interface Builder. I don't think we should use the segue class  programmatically.

TheDarkCode, line 14: We could do it in `AnimatableViewController` as something like a convenience extension on `performSegueWithIdentifier` method. Just kicking the idea around since the completion handler there is essentially inaccessible as-is.

JakeLin:
@TheDarkCode thanks for your PR, I finally understand your approach for adding callback completion. From my point of view, I don't think Apple recommend us to use `UIStoryboardSegue` programmatically other than `override func perform()` which is called by Interface Builder. 

I am sorry I am not going to merge it to master now since I haven't found a good use case for  `func perform(completion: OptionalAnimatableCompletion)`, Can you please tell what kind of use case we will call  `func perform(completion: OptionalAnimatableCompletion)` programmatically?

TheDarkCode:
So to show as an example, but one which refers to code that was used prior to #204 in TransitionViewController.swift as an implementation:

``` swift
    let segue = segueClass.init(identifier: segueName, source: self, destination: toViewController)
    prepareForSegue(segue, sender: self)
    segue.perform {
         //print("I'm inside the callback")
    }
```

To do this you would then need to add:

``` swift
public protocol SegueCompletionHandler {
    func perform(callback: (() -> Void)?)
}

extension UIStoryboardSegue: SegueCompletionHandler {
    public func perform(callback: (() -> Void)?) {
          self.perform() // not using callback unless you add override
    }
}
```

In reference to adding to something like PresentFoldSegue:

``` swift
  public override func perform(completion: OptionalAnimatableCompletion) {
    destinationViewController.transitioningDelegate = PresenterManager.sharedManager().retrievePresenter(.Fold(direction: .Left, params: []))
    sourceViewController.presentViewController(destinationViewController, animated: true, completion: completion)
  }
```

tbaranes:
I'm joining @JakeLin on this, you shouldn't use programatically a segue for that goal since you can already reach it by using `AnimatableNavigationController`, or `AnimatableViewController`, can't you?

TheDarkCode:
@tbaranes We are talking about calling a completion handler on the `sourceViewController` when performing a custom Animatable segue that is initiated by it. All that is setup in the code now is being able to pass data using a sender, as normal. However, by implementing a `presentViewController` from the `sourceViewController` there is a completion handler that is encapsulated with no means of accessing it. This is due to the subclassing that went on. Apple actually proposes a different method: [Creating a Custom Segue](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIStoryboardSegue_Class/#//apple_ref/doc/uid/TP40010911-CH1-SW11)

tbaranes:
I really didn't get what you were looking for! Thanks for the explanation. I don't see a nice use case in my project, but I can understand why you ask for it.
Anyway, it's up to @JakeLin 👍

TheDarkCode:
The question resolves down to this issue where you can't edit the segue properties in storyboard like IBDesignable yet - yell at Apple for that one. Thus, you can say you want X animation, you just can't say it should do Y on completion.

It is all limited via inceptors you have to program yourself on the destination, and senders you pass from the source. But there is no callback to the source VC to do something special when performing that specific segue (ie: maybe it wants to release data it shouldn't keep a copy of because it would be stale). At least when combining animatable types and non-animatable types. This is due to the choice of subclassing versus making a custom init.

JakeLin:
@TheDarkCode thanks for your update, I maybe not right, but I think we should use [presentViewController(_:animated:completion:)](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewController_Class/#//apple_ref/occ/instm/UIViewController/presentViewController:animated:completion:) to handle completion. And then use [animationControllerForPresentedController(_:presentingController:sourceController:)](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewControllerTransitioningDelegate_protocol/#//apple_ref/occ/intfm/UIViewControllerTransitioningDelegate/animationControllerForPresentedController:presentingController:sourceController:) to specify the animation for the transition. We can use any `***Animator` within `IBAnimatable` for the purpose. Is the use case you want to achieve? If yes, we don't need `public override func perform(completion: OptionalAnimatableCompletion)` 

The reason we subclass `UIStoryboardSegue` is we can not configure different transition animation in Interface Builder. When we subclass `UIStoryboardSegue`, we only override [perform()](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIStoryboardSegue_Class/#//apple_ref/occ/instm/UIStoryboardSegue/perform), which is using Template Pattern if I remember correctly. It is used by the framework like UIKit only.

> You can subclass UIStoryboardSegue in situations where you want to provide a custom transition between view controllers in your application. To use your custom segue, create a segue line between the appropriate view controllers in **Interface Builder** and set its type to Custom in the inspector; you must also specify the class name of the segue to use in the inspector.
> 
> When the **storyboard runtime** detects a custom segue, it creates a new instance of your class, configures it with the view controller objects, asks the view controller source to prepare for the segue, and then performs the segue.

Please notice [the document from Apple](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIStoryboardSegue_Class/), I have highlighted some key words. If I understand correctly, subclass of `UIStoryboardSegue` should be only used in **Interface Builder** and executed by **storyboard runtime**

The example you mentioned in `TransitionViewController` is an exception. Because I want to demonstrate all transition animations in one set of Storyboard. In a real app, for consistency, we should only use one animation for the same transition via one segue. We should not create Segue object by  ourself like `let segue = segueClass.init(identifier: segueName, source: self, destination: toViewController)` unless we work for Apple and develop UIKit.

@TheDarkCode If you still have any question, or your use case is different from what I mentioned, please let me know. Thank for participating in this project. As I mentioned in our discussion in the project before, designing APIs are a very challenging process. We need to balance a lot of aspects and constraints. One of our simple principles is to expose as less and enough as possible. In this case, we only add new API because the current APIs can't support the particular use case.

@tbaranes Thanks for discussing this PR.

JakeLin:
@TheDarkCode As we discussed off Github, I close this PR now. Please feel free to leave comments if you have any questions on this PR. Thanks.

PR #215, opened 2016-06-03
Minor improvements in playgrounds

tbaranes:
Added two labels to guide how to come back after a `tap to present` and `tap to present via segue` from Transitions -> IB. I have to say I struggle a bit the first time I get there!

Also, I was wondering if `Interactions -> Container` is really the right place for this? I mean, by reading `Interactions` I'm expecting gesture, not a sample of the result of `ContainerTransition`. Container shouldn't it be at the root of the lists instead of in a sub one?

### DIFF
--- a/IBAnimatableApp/README.md
+++ b/IBAnimatableApp/README.md
@@ -3,4 +3,4 @@
 
 The `code` folder is not mandatory for the demo App. **The app was made in Interface Builder with `IBAnimatable` without a single line of code**. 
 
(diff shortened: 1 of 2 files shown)


JakeLin:
👍 , thanks for fixing the typos and making the example clearer for the user to dismiss.

For `ContainerTransition`, I didn't think carefully before, you are right, I think it should be in "Transitions". In "Interactions", we can demonstrate Gestures. 

Since it is in the demo app, I am going to fix after 2.3 release. I want to release 2.3 before WWDC. And we can work on some new features if required.

Please merge it when you free ready.

Issue #193, opened 2016-05-06
Improving README


tbaranes:
I'm creating this issue to track the discussion started in #186 

First thing, I think we have to decide what we want to keep in the main README:
- What is it?
- Install
- where to find the docs?
- Documentation (just a summary with the links)

As you suggested, we can move out all the API documentation, I think we have two solutions where we can move it:
- Creating a directory `Documentation` which will contains several markdown files, examples:
  - [Fastlane](https://github.com/fastlane/fastlane): 
  - [Quick](https://github.com/Quick/Quick)
  - [Moya](https://github.com/Moya/Moya)
  - [Alamofire](https://github.com/Alamofire/Alamofire) -> just the migration guide
- Put the documentation directly in the wiki

=> I prefer the first one since we already using the wiki with technical stuff. Moreover, it seems to be simpler to use

Whatever our final choice, here a first suggestion of how we can divide the API documentation:
- TransitionAnimatable.md
- Animatable.md
- Designable.md
- Extensions.md
- Segue.md -> should it be merged in TransitionAnimatable?

JakeLin:
@tbaranes thanks for starting this, Do you like `Documentation` folder more than wiki? I don't mind either way.

I like your groups, may have some more friendly to the user thinking, they match the project's goal.
- Animations: Animatable
- UI: Designable
- Transitions: TransitionAnimatable and Segue
- Misc: Extensions  

Another important thing I am going to do after release 2.3 is blogging. I am going to write some blog posts about "How to prototype and App Store ready Apps in IB with IBAnimatable", "How to prototype and design animations in Swift Playground", "How to prototype and design transition animations in IB" etc. `IBAnimatable` is the first library to support designing transition animations in IB. I think it is a very useful tool for prototyping and design since we have a lot of comprehensive features.

tbaranes:
Yes, I prefer the `Documentation` folder, I find it more usable since we can edit it them locally.

Excellent idea for your group names, it's user friendly and really understandable. 
I'm looking forward to read your articles, moreover it will be an excellent idea since `IBAnimatable` is more and more popular, but is clearly missing some tutorials / helpers to get started about advanced usage.

tbaranes:
Closing it now since the new structure is already there. We can still open a new issues just to list the documentation missing.

PR #216, opened 2016-06-05
Improve README.md

JakeLin:
It is the first step for #193 , off cause, there are a lot of spaces we can improve, but moving forward is better than doing nothing. 

In this PR, we move APIs section in a separate file called "APIs.md". I think in the feature we can split it into different more specific documents like "Transitions.md".

@tbaranes please have a look when you have time, we are about to release 2.3. 😉


tbaranes:
Seems good to me. Two suggestions:
- Documentation to API link isn't working, is that normal?
- Since the documentation _is not really big_ and at the bottom of the README, we can easily miss it. Maybe it would be better to put it at the top of the README, just after `Use cases` since the others field will (I suppose) be moved in the `Documentation` directory. If we do that, we have to add a note "all the documentation hasn't be migrated yet, check out the end of the README" or whatever

JakeLin:
@tbaranes Thanks your suggestions.
1. I am using absolute URL, which contains branch name, I need to merge into to see it. I don't know how to use relative URL on GitHub.
2. Have moved the "Documentations" section up.

PR #220, opened 2016-06-15
Migrate syntax to Swift 3 so we can build with Xcode 8

YoniTsafir:
Only ran the converter and manually fixed some issues it didn't find automatically. warning: Didn't heavily test the changes.

Should be merged to a new branch of course and not to master for now (don't know how to do this in github).

### DIFF
--- a/IBAnimatable/PortalAnimator.swift
+++ b/IBAnimatable/PortalAnimator.swift
@@ -27,14 +27,14 @@ public class PortalAnimator: NSObject, AnimatedTransitioning {
     }
     
     switch fromDirection {
-    case .Forward:
-      self.transitionAnimationType = .Portal(direction: .Forward, params: params)
-      self.reverseAnimationType = .Portal(direction: .Backward, params: params)
-      self.interactiveGestureType = .Pinch(direction: .Close)
+    case .forward:
+      self.transitionAnimationType = .portal(direction: .forward, params: params)
+      self.reverseAnimationType = .portal(direction: .backward, params: params)
+      self.interactiveGestureType = .pinch(direction: .Close)
     default:
-      self.transitionAnimationType = .Portal(direction: .Backward, params: params)
-      self.reverseAnimationType = .Portal(direction: .Forward, params: params)
-      self.interactiveGestureType = .Pinch(direction: .Open)
+      self.transitionAnimationType = .portal(direction: .backward, params: params)
+      self.reverseAnimationType = .portal(direction: .forward, params: params)
+      self.interactiveGestureType = .pinch(direction: .Open)
     }
 
     super.init()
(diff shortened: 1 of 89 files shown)


tbaranes:
Won't be better to support swift 2.3 until an official release of swift 3? This way we will be compatible with Xcode 7 and 8.

YoniTsafir:
That's why I suggest doing it in a branch. I use IBAnimatable via Carthage in my project and I CAN'T move to Swift 3 unless IBAnimatable is compiled with Swift 3 as well (it gives an error trying to import it).

IBAnimatable/PortalAnimator.swift
```diff
@@ -67,33 +67,33 @@ extension PortalAnimator: UIViewControllerAnimatedTransitioning {
...
 
     let leftSnapshotRegion = CGRect(x: 0, y: 0, width: fromView.frame.width / 2, height: fromView.bounds.height)
-    let leftHandView = fromView.resizableSnapshotViewFromRect(leftSnapshotRegion, afterScreenUpdates: false, withCapInsets: UIEdgeInsetsZero)
+    let leftHandView = fromView.resizableSnapshotView(from: leftSnapshotRegion, afterScreenUpdates: false, withCapInsets: UIEdgeInsetsZero)!
     leftHandView.frame = leftSnapshotRegion
     containerView.addSubview(leftHandView)
 
     let rightSnapshotRegion = CGRect(x: fromView.frame.width / 2, y: 0, width: fromView.frame.width / 2, height: fromView.frame.height)
-    let rightHandView = fromView.resizableSnapshotViewFromRect(rightSnapshotRegion, afterScreenUpdates: false, withCapInsets: UIEdgeInsetsZero)
+    let rightHandView = fromView.resizableSnapshotView(from: rightSnapshotRegion, afterScreenUpdates: false, withCapInsets: UIEdgeInsetsZero)!
```
JakeLin, line 82: I am not sure should we force unwrap it?

IBAnimatable/SystemCubeAnimator.swift
```diff
@@ -22,34 +22,34 @@ public class SystemCubeAnimator: NSObject, AnimatedTransitioning {
...
-      self.reverseAnimationType = .SystemCube(fromDirection: .Bottom)
-      self.interactiveGestureType = .Pan(fromDirection: .Bottom)
-    case .Bottom:
-      self.transitionAnimationType = .SystemCube(fromDirection: .Bottom)
-      self.reverseAnimationType = .SystemCube(fromDirection: .Top)
-      self.interactiveGestureType = .Pan(fromDirection: .Top)
+    case .right:
+      self.transitionAnimationType = .systemCube(fromDirection: .right)
+      self.reverseAnimationType = .systemCube(fromDirection: .left)
+      self.interactiveGestureType = .pan(fromDirection: .Left)
```
JakeLin, line 28: This enum also needs to change.

JakeLin:
@theyonibomber Thank you very much for doing that, I was working on the same upgrade and saw your nice PR. I stopped my thing and created a branch called swift3, can you pull request to https://github.com/JakeLin/IBAnimatable/tree/swift3 ?

I think we need to fix more things for Swift 3. e.g. changing `public func transitionDuration(_ transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval` to `public func transitionDuration(transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval`, but your one is a very good starting point.

Related to #221

JakeLin:
@tbaranes I am not sure should we upgrade to Swift 2.3? As far as I know, Swift 2.3 = Swift 2.2 + new APIs. All apps can't use those APIs untile Xcode 8 RTM ships. Once Xcode 8 ships, we upgrade to Swift 3. Please let me know if I missed something.

If we don't upgrade to 2.3 then we will keep current master with 2.2 and have swift3 branch for Swift 3.

PR #227, opened 2016-06-24
TextView placeholder

### LINKED ISSUE #181: Hello Sir, how about adding a placeholder in UITextview ..!
This library got everything to get a push in development so sir if you like can you add a placeholder in UITextview, it can be added programmatically but if it is available in IBAnimatable which is a great library it would be great.

tbaranes:
Close #181 
This PR is a first try to make `AnimatableTextView` conforms to `PlaceholderDesignable`. Current result:
- Designable placeholder text and color 
- Supports multilines
- The placeholder uses all the properties set in the textview: textAlignment, font... any of them are currently customisable for the placeholder.

I think `AnimatableTextView` has too much knowledge about the placeholder, I'd like to abstract more in `PlaceholderDesignable` but since it's fully related to the instance itself (override properties, notification...) I didn't find a working output. The main issue about this implementation is that makes a lot of work to conform `PlaceholderDesignable` from an `UITextView`.  Any suggestions welcomed! 

Otherwise, it's ready to go 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 None
 
 #### Enhancements
-None
+- New designable properties for `UITextView`: placeholder text and color [#227](https://github.com/JakeLin/IBAnimatable/issues/227)
 
 #### Bugfixes
 None
--- a/Documentation/APIs.md
+++ b/Documentation/APIs.md
@@ -102,7 +102,8 @@ It is used in `AnimatableTextField` to add padding on either or both sides.
 #### `PlaceholderDesignable`
 | Property name | Data type | Description |
 | ------------- |:-------------:| ----- |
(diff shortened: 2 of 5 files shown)


JakeLin:
@tbaranes Good work 👍, I don't find a better way to move `placeholderLabel` and `placeholderLabelConstraints` out of `AnimatableTextView`, if we put them in `PlaceholderDesignable`,  then `AnimatableTextField` has to override them.  Anther ideas I can think of is **Associated Objects**, but I don't like to use Objective-C Runtime magic if possible.

I think your solution is fine because we are hitting some limitations. And I add an example in the demo app, it works great.  Please merge it once you think it is ready. Thanks again for your great work.

simulator screen shot jun 25 2016 10 51 30 pm

tbaranes:
Good idea the demo update. Thanks!

TheDarkCode:
Nice PR. Would be interesting to allow textviews to have speech and word tracking options via the builder too.

PR #232, opened 2016-06-29
Add AnimatableScrollView

tbaranes:


IBAnimatable/AnimatableScrollView.swift
```diff
@@ -0,0 +1,154 @@
+//
+//  Created by Tom Baranes on 28/06/16.
+//  Copyright © 2015 Jake Lin. All rights reserved.
+//
+
+
+import UIKit
+import IBAnimatable
```
TheDarkCode: Why are you importing IBAnimatable?

tbaranes: My bad. I did that in my project then copy in IBAnimatable, forgot to remove it.

JakeLin:
👍 thanks, @tbaranes , I think we can release 2.4 once I fix some mask issues.

tbaranes:
Agree for the release 👌

PR #228, opened 2016-06-25
Animatable slider

tbaranes:
Ok, that was an easy one 😁

Basic `AnimatableSlider` conforming some protocols that I think it could be used with a slider: `Animatable`, `Border`, `Shadow`, `Rotation`

I didn't update the demo project, should we add an example even if it's really basic? 

Close #70

Issue #218, opened 2016-06-06
API Improvement proposal - Supporting enum for Inspectable properties in code and Interface Builder 


lastMove:
**enum checklist to improve**
- [x] MaskType
- [x] AnimationType
- [x] GradientStartPoint
- [x] BorderSide
- [x] BlurEffectStyle (Add new options `regular` and `prominent` for `UIBlurEffectStyle`)
- [x] ColorType - FlatColorsTypeScript.swift
- [x] GradientType - GradientsTypeScript.swift
- [x] InteractiveGestureType
- [x] TransitionAnimationType
- [x] TransitionFromDirection
- [x] TransitionHollowState
- [x] TransitionPageType
- [x] PresentationAnimationType
- [x] PresentationModalSize
- [x] PresentationModalPosition
- [x] PresentationKeyboardTranslation

Other tasks
- [x] Migration Guide 

-------- Original Message --------
Hi All, I really love using this library. However i think that the API can be improved.

Presently a lot of properties are String because of InterfaceBuilder. Some property would be more usable, self-documented and less error-prone if they were enums with associated type.

Enums with associatedType are powerful and fit perfectly with IBAnimatable Purpose.

Before pursuing one example with maskType.using enums it will be possible to make this:

``` Swift

animatableView.maskType = MaskType.Parrallelogram(angle:45, direction: .Top)

animatableView.maskType = .Polygon(sides:8)
animatableView.maskType = .Circle

animatableView.maskType = “Polygon(8)“ // yes this is possible since the MaskType Enum adopts StringLiteralConvertible protocol

BUT (sadly) this will NOT be possible 

let string = “Polygon(8);
animatableView.maskType = string // Since there is no StringConvertible protocol in swift (yet) , only String Lterals are convertible 

```

I tried a lot of thing to be able to do this AND to stay compatible with Interface Builder But No way :(

The most convenient way IMO is to make 2 properties. 

``` Swift

private var maskType : String
var eMaskType : MaskType

```

Since the string maskType is made for Interface Builder it can be private. 
The advantage is that when accessing from code they will be just one way exposed. 
and since MaskType is StringLiteralConvertible, the existant code that looks like : 
view.maskType = “Circle”

will justNeed to be transformed to 

`view.eMaskType = “Circle”
`
And we could use`@avalable(iOS, renamed=“eMaskType”)` in order to simplify the migration

However, Since WWDC is soon, we can just wait to see Interface Builder improvements :) 
[...]

PR #233, opened 2016-06-29
Update Enums to camelCase

TheDarkCode:
This PR also adds #232 and #228 to the Swift3 branch.


IBAnimatable/InteractiveGestureType.swift
```diff
@@ -19,18 +19,18 @@ public enum InteractiveGestureType {
...
+      interactiveGestureType.hasPrefix("pinch") {
       return fromStringWithDirection(interactiveGestureType)
     }
     return nil
   }
   
   // Return the `String` without qualification
   public func toString() -> String {
-    let namespace = "IBAnimatable." + String(GestureDirection) + "."
+    let namespace = "Ahtau." + String(GestureDirection) + "."
```
JakeLin: That should not change.

IBAnimatable/SystemTransitionType.swift
```diff
@@ -10,18 +10,18 @@ import Foundation
...
+  case flip = "flip"
+  case cube = "cube"
+  case pageCurl = "pageCurl"
+  case pageUnCurl = "pageUnCurl"
+  case rippleEffect = "rippleEffect"
+  case suckEffect = "suckEffect"
+  case cameraIris = "cameraIris"
+  case cameraIrisHollowOpen = "cameraIrisHollowOpen"
+  case cameraIrisHollowClose = "cameraIrisHollowClose"
+  case rotate = "rotate"
```
JakeLin: We can remove `= "fade"` part now since they are same.

IBAnimatable/AnimatableScrollView.swift
```diff
@@ -0,0 +1,153 @@
...
+    // MARK: - Animatable
+    @IBInspectable public var animationType: String?
+    @IBInspectable public var autoRun: Bool = true
+    @IBInspectable public var duration: Double = Double.nan
+    @IBInspectable public var delay: Double = Double.nan
+    @IBInspectable public var damping: CGFloat = CGFloat.nan
+    @IBInspectable public var velocity: CGFloat = CGFloat.nan
+    @IBInspectable public var force: CGFloat = CGFloat.nan
+    @IBInspectable public var repeatCount: Float = Float.nan
+//    @IBInspectable public var x: CGFloat = CGFloat.nan
```
JakeLin: Why do we remove it? it is using in `MoveTo` and `MoveBy` animations.

JakeLin:
@TheDarkCode Thanks for your PR, I have made some comments, please have a look.

And why do you change all file mode from 100644 to 100755? Refer to http://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha, it may all the files include the +x (executable) flag. I think the old one were fine.

IBAnimatable/InteractiveGestureType.swift
TheDarkCode: Opps. Nice catch.

IBAnimatable/AnimatableScrollView.swift
TheDarkCode: This is a change from Ahtau, not intended for IBAnimatable. I have written other extensions for UIView that make use of x and y values without the Animatable protocol in place.

IBAnimatable/SystemTransitionType.swift
TheDarkCode: Fixed it. Thanks for the catch.

JakeLin:
Thanks @TheDarkCode 👍

Hi, @tbaranes if you have time, can you have a quick look at this PR as well?

Related to #218

tbaranes:
@TheDarkCode why the files mode changed: changelog, .travis... has been changed? There is a lot!

Also, I think we should rebase master in swift3 instead of add the last PRs directly here, what do you think about it?

EDIT: Didn't see that @JakeLin already reported the file mode changed, my bad!

IBAnimatable/ColorType.swift
```diff
@@ -6,24 +6,24 @@
...
-  case FlatGreenSea
-  case FlatBelizeHole
-  case FlatNephritis
-  case FlatMidnightBlue
-  case FlatClouds
-  case FlatWisteria
-  case FlatCarrot
-  case FlatPumpkin
+  case flatEmerland
+  case flatPomegranate
```
tbaranes: For this enum, we have to update the script since it's automatically generated

IBAnimatable/FillDesignable.swift
```diff
@@ -52,45 +52,45 @@ private extension FillDesignable {
     }
     
     switch colorType {
-    case .FlatEmerland:
+    case .flatEmerland:
```
tbaranes: Same here, we have to update the script since it's automatically generated.

IBAnimatable/GradientDesignable.swift
```diff
@@ -77,283 +77,283 @@ public extension GradientDesignable where Self: UIView {
     }
     
     switch gradientType {
-    case .SunnyDays:
+    case .sunnyDays:
```
tbaranes: Same here, we have to update the script since it's automatically generated.

IBAnimatable/GradientType.swift
```diff
@@ -6,143 +6,143 @@
 import Foundation
 
 public enum GradientType: String {
-  case SunnyDays
```
tbaranes, line 9: Same here, we have to update the script since it's automatically generated.

tbaranes:
I'm having a second though about this update: I know it's swift 3 guidelines to have camelCase enums, but most of the users are using in interface builder, we won't have errors compilation even if the string doesn't match the enum... I think about two solutions:
- Makes all the string check insensitive (some of them are already insensitive, but I don't think we did it for all them)
- Break all the layouts and let everyone take care of their issues (I think they will have some warnings) 😆 

Anyway, if we merged this now, we definitely have to handle this before merging swift3 in master, otherwise...😱

IBAnimatable/GradientType.swift
JakeLin, line 9: @tbaranes I ask @TheDarkCode to leave these scripts now because we are still working on #218, and we may need to have `IBEnum` for `ColorType` and `GradientType`. Once we have a pattern for enum, we can update those scripts to use the new pattern.

tbaranes:
Beside some feedbacks, it's a nice first step forward to swift 3! 👍

JakeLin:
We may need to "Makes all the string check insensitive (some of them are already insensitive, but I don't think we did it for all them)" to have backward compatibility. But since we will change a lot, there will be some breaking changes 😫

tbaranes:
The most important part, I think,  is to minimise interface builder breaking change: it's really annoying to fix them all when `IBAnimatable` is highly use in some projects. Code breaking change are really easier to handle since we have compilation errors.

IBAnimatable/GradientType.swift
tbaranes, line 9: 👍

TheDarkCode:
@tbaranes Why not use a script to process storyboard files and change any instances of the old type to the new one?

tbaranes:
That could be a good idea, but I'm pretty sure that at least half of the users won't read the CHANGELOG or get the knowledge about this, and will just fix everything manually.

TheDarkCode:
@tbaranes This is where one might say "well, let them squirm". I think there is only so much you can do to push the project forward and adopt the Swift 3 API guidelines while also not throwing a wrench into everyone's existing projects. At a certain point people will just have to adopt the breaking changes or not get the updates. But I think that if it was clear with compiler warnings, it might be more obvious.

IBAnimatable/GradientType.swift
TheDarkCode, line 9: One thing I was pointing out to @JakeLin was that you have some instances where things are directly convertible to "proper" camel case with a first character modification. Ie: BacktotheFuture I changed to backToTheFuture. As well as others like: GreenandBlue, BetweenNightandDay, Curiosityblue, etc.

tbaranes:
Of course, I'm not against breaking changes, otherwise I would be against swift 3 itself 😆  but in this case, we have an easy way to avoid it. Moreover, I think it's a nice move since it's easy to just forgot a capital and get the wrong layout... we don't have to be that strict when it's about interface builder.

tbaranes:
Another point: why all the images has been deleted / added? I suppose it's an issue similar to the file mode, but still...

TheDarkCode:
@tbaranes We can do either a build script, an external script - like the current color and gradient scripts, or put compiler warnings that notify users. Alternatively we make an online tool where people paste their storyboard code into and we just run a simple parser to remove all the old instances. Lastly, its case insensitivity as an option.

TheDarkCode:
@tbaranes This PR might just be a scrap and re-do since my fork was out of date and must have gotten corrupted when I was changing things around.

TheDarkCode:
@tbaranes The images are from the file permissions changing. I'm going to fix the file permission errors in a new commit to the branch.

TheDarkCode:
@JakeLin @tbaranes Switched the file mode back to 644.

TheDarkCode:
I added #234 into this PR also.

JakeLin:
@TheDarkCode Thanks for your PR, we need improve some enums to support existing Interface Builder configurations.

PR #234, opened 2016-06-29
Fix some small bugs for Masks

JakeLin:
Fix some small bugs for Circle, Polygon, and Wave masks. Also, add examples for demonstrating all masks. 

@tbaranes Could you please have a look when you have time, thanks.

simulator screen shot jun 29 2016 11 21 33 pm
simulator screen shot jun 29 2016 11 21 36 pm

### DIFF
--- a/IBAnimatable/MaskDesignable.swift
+++ b/IBAnimatable/MaskDesignable.swift
@@ -72,7 +72,8 @@ public extension MaskDesignable where Self: UIView {
     let center = CGPoint(x: bounds.width / 2.0, y: bounds.height / 2.0)
     var angle: CGFloat = -CGFloat(M_PI / 2.0)
     let angleIncrement = CGFloat(M_PI * 2.0 / Double(sides))
-    let radius = bounds.width / 2.0
+    let length = min(bounds.width, bounds.height)
+    let radius = length / 2.0
     
     path.moveToPoint(pointFrom(angle, radius: radius, offset: center))
     for _ in 1...sides - 1 {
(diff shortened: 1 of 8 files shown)


IBAnimatable/MaskDesignable.swift
```diff
@@ -191,7 +196,13 @@ public extension MaskDesignable where Self: UIView {
   
   private func maskWaveFromString(mask: String) {
     let params = retrieveMaskParameters(mask, maskName: MaskType.Wave.rawValue).componentsSeparatedByString(",")
-    if let unwrappedWidth = Float(params[1]), unwrappedOffset = Float(params[2]) where params.count == 3 {
+    
+    guard params.count == 3 else {
```
tbaranes, line 200: Not sure that is necessary, but shouldn't we able to use the wave with 1 or 2 parameters if we only need to set the direction... ? Anyway, that can be another PR.
That's comment is extendable to any masks using parameters.

tbaranes:
Awesome, that's a really nice fix 👍

IBAnimatable/MaskDesignable.swift
JakeLin, line 200: Right, will fix it, I thought `if let unwrappedWidth = Float(params[1]), unwrappedOffset = Float(params[2]) where params.count == 3` should guard it, But it didn't, and crash the app.

JakeLin, line 200: @tbaranes I have tested the other masks, they can support wrong number of parameters.

PR #237, opened 2016-07-06
Feature/swift3 masktype

lastMove:
MaskType Reworked. 
Related to #218

### DIFF
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -566,8 +566,8 @@ public extension Animatable where Self: UIView {
   public func rotate(_ clockwise: Bool = true, completion: AnimatableCompletion? = nil) {
     CALayer.animate({
       let animation = CABasicAnimation(keyPath: "transform.rotation")
-      animation.fromValue = clockwise ? 0 : ((360 * M_PI) / 180)
-      animation.toValue = clockwise ? ((360 * M_PI) / 180) : 0
+      animation.fromValue = clockwise ? 0 : CGFloat.pi * 2
+      animation.toValue = clockwise ? CGFloat.pi * 2 : 0
       animation.duration = CFTimeInterval(self.duration)
       animation.repeatCount = self.repeatCount
       animation.autoreverses = false
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -566,8 +566,9 @@ public extension Animatable where Self: UIView {
   public func rotate(_ clockwise: Bool = true, completion: AnimatableCompletion? = nil) {
     CALayer.animate({
       let animation = CABasicAnimation(keyPath: "transform.rotation")
-      animation.fromValue = clockwise ? 0 : CGFloat.pi * 2
-      animation.toValue = clockwise ? CGFloat.pi * 2 : 0
+      let pi: CGFloat = .pi
+      animation.fromValue = clockwise ? 0 : pi * 2
+      animation.toValue = clockwise ? pi * 2 : 0
       animation.duration = CFTimeInterval(self.duration)
       animation.repeatCount = self.repeatCount
       animation.autoreverses = false
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -103,12 +103,21 @@ import UIKit
   @IBInspectable public var startPoint: String?
   
   // MARK: - MaskDesignable
-  @IBInspectable public var maskType: String? {
+  public var eMaskType:MaskType = .none {
     didSet {
       configMask()
-      configBorder()      
+      configBorder()
     }
   }
+  
+  @IBInspectable private var maskType: String? {
+    didSet {
+      if let maskType = maskType {
+        eMaskType = MaskType(string:maskType);
+      }
+    }
+  }
+  
   
   // MARK: - Animatable
   @IBInspectable public var animationType: String?
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -112,13 +112,10 @@ import UIKit
   
   @IBInspectable private var maskType: String? {
     didSet {
-      if let maskType = maskType {
-        eMaskType = MaskType(string:maskType)
-      }
+      eMaskType = MaskType(string: maskType)
     }
   }
-  
-  
+
   // MARK: - Animatable
   @IBInspectable public var animationType: String?
   @IBInspectable public var autoRun: Bool = true
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -110,7 +110,7 @@ import UIKit
     }
   }
   
-  /// The mask type used in Interface Builder. **Should not** this property in code.
+  /// The mask type used in Interface Builder. **Should not** use this property in code.
   @IBInspectable var _maskType: String? {
     didSet {
       maskType = MaskType(string: _maskType)
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -103,7 +103,7 @@ import UIKit
   @IBInspectable public var startPoint: String?
   
   // MARK: - MaskDesignable
-  public var eMaskType:MaskType = .none {
+  public var eMaskType: MaskType = .none {
     didSet {
       configMask()
       configBorder()
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -110,6 +110,7 @@ import UIKit
     }
   }
   
+  /// The mask type used in Interface Builder. **Should not** this property in code.
   @IBInspectable var _maskType: String? {
     didSet {
       maskType = MaskType(string: _maskType)
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -113,7 +113,7 @@ import UIKit
   @IBInspectable private var maskType: String? {
     didSet {
       if let maskType = maskType {
-        eMaskType = MaskType(string:maskType);
+        eMaskType = MaskType(string:maskType)
       }
     }
   }
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -103,16 +103,16 @@ import UIKit
   @IBInspectable public var startPoint: String?
   
   // MARK: - MaskDesignable
-  public var eMaskType: MaskType = .none {
+  public var maskType: MaskType = .none {
     didSet {
       configMask()
       configBorder()
     }
   }
   
-  @IBInspectable internal var maskType: String? {
+  @IBInspectable var _maskType: String? {
     didSet {
-      eMaskType = MaskType(string: maskType)
+      maskType = MaskType(string: _maskType)
     }
   }
 
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -110,7 +110,7 @@ import UIKit
     }
   }
   
-  @IBInspectable private var maskType: String? {
+  @IBInspectable internal var maskType: String? {
     didSet {
       eMaskType = MaskType(string: maskType)
     }
--- a/IBAnimatable/CardsAnimator.swift
+++ b/IBAnimatable/CardsAnimator.swift
@@ -140,7 +140,7 @@ private extension CardsAnimator {
       var t1 = CATransform3DIdentity
       t1.m34 = 1.0 / -900
       t1 = CATransform3DScale(t1, 0.95, 0.95, 1)
-      t1 = CATransform3DRotate(t1, 15.0 * CGFloat(M_PI) / 180.0, 1, 0, 0)
+      t1 = CATransform3DRotate(t1, 15.0 * .pi / 180.0, 1, 0, 0)
       return t1
   }
 
--- a/IBAnimatable/IBEnum.swift
+++ b/IBAnimatable/IBEnum.swift
@@ -1,0 +1,34 @@
+//
+//  IBEnum.swift
+//  IBAnimatableApp3
+//
+//  Created by jason akakpo on 27/06/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import Foundation
+
+
+public protocol IBEnum  {
+  init(string:String)
+}
+
+public extension IBEnum {
+  
+  
+  /// Helper function that returns a tuple containing the name and params
+  ///
+  ///
+  /// - Discussion
+  /// Since "name(param1,param2,param3)" is the classic IBAnimatable format
+  ///
+  /// this method will return "wave"
+  static func extractNameAndParams(_ string:String) -> (name:String, params:[String]) {
+    let tokens = string.components(separatedBy: CharacterSet(charactersIn: "()")).filter({!$0.isEmpty}) // extract params list   self.enumName
+    let maskName = tokens.first ?? "";
+    let paramsString = tokens.count >= 2 ? tokens[1] : "";
+    let params = paramsString.components(separatedBy: ","); // params list
+    return (name:maskName, params:params);
+  }
+  
+}
--- a/IBAnimatable/IBEnum.swift
+++ b/IBAnimatable/IBEnum.swift
@@ -5,43 +5,47 @@
 
 import Foundation
 
-
+/**
+ A protocol provides extension method for converting `String` into `enum`.
+ Because `@IBInspectable` property can not support `enum` directly. To provide both programmatical `enum` API and `@IBInspectable` supported type `String`, we use `IBEnum` to bridge Swift `enum` and `String`
+ */
 public protocol IBEnum: StringLiteralConvertible {
+  /**
+   Initializes a swift `enum` with provided optional string
+   
+   - Parameter string: The optional string to be converted into `enum`.
+   */
   init(string: String?)
 }
 
 public extension IBEnum {
-  
-  /// Helper function that returns a tuple containing the name and params
-  ///
-  ///
-  /// - Discussion
-  /// Since "name(param1,param2,param3)" is the classic IBAnimatable format
-  ///
-  /// this method will return "wave"
-  static func extractNameAndParams(from string: String) -> (name: String, params: [String]) {
-    let tokens = string.components(separatedBy: CharacterSet(charactersIn: "()")).filter({!$0.isEmpty}) // extract params list   self.enumName
-    let maskName = tokens.first ?? ""
+  /**
+   Helper function that returns a tuple containing the name and params from a string `string`
+   
+   - Parameter from string: The string to be converted into `enum`.
+    - Discussion: the string format is like "enumName(param1,param2,param3)"
+   - Returns: A tuple containing the name and an array of parameter string
+   */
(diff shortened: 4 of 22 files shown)


IBAnimatable/Animatable.swift
```diff
@@ -566,8 +566,8 @@ public extension Animatable where Self: UIView {
   public func rotate(_ clockwise: Bool = true, completion: AnimatableCompletion? = nil) {
     CALayer.animate({
       let animation = CABasicAnimation(keyPath: "transform.rotation")
-      animation.fromValue = clockwise ? 0 : ((360 * M_PI) / 180)
-      animation.toValue = clockwise ? ((360 * M_PI) / 180) : 0
+      animation.fromValue = clockwise ? 0 : CGFloat.pi * 2
+      animation.toValue = clockwise ? CGFloat.pi * 2 : 0
```
tbaranes: Just a detail, everywhere else you are using `.pi` directly, can we uniforms it?

IBAnimatable/IBEnum.swift
```diff
@@ -0,0 +1,47 @@
...
+
+public extension IBEnum {
+  
+  /// Helper function that returns a tuple containing the name and params
+  ///
+  ///
+  /// - Discussion
+  /// Since "name(param1,param2,param3)" is the classic IBAnimatable format
+  ///
+  /// this method will return "wave"
```
tbaranes: It won't be always be "wave", is it?

IBAnimatable/MaskDesignable.swift
```diff
@@ -141,35 +110,27 @@ public extension MaskDesignable where Self: UIView {
...
   private func maskParallelogramBezierPath(_ topLeftAngle: Double) -> UIBezierPath {
-    let topLeftAngleRad = Double(topLeftAngle) * M_PI / 180
+    let topLeftAngleRad = topLeftAngle * .pi / 180
     let path = UIBezierPath()
-    let offset = abs(CGFloat(tan(topLeftAngleRad - M_PI / 2)) * bounds.height)
+    let offset = abs(CGFloat(tan(topLeftAngleRad - .pi / 2)) * bounds.height)
     
     if topLeftAngle <= 90 {
-      path.move(to: CGPoint(x: 0, y: 0))
+      path.move(to: CGPoint(x:0, y:0))
```
tbaranes, line 181: Just curious, SwiftLint doesn't report any warnings here?

IBAnimatable/MaskType.swift
```diff
@@ -4,12 +4,68 @@
...
+  /// Isocele triangle shape. The triangle's height is equal to the view's frame height. If the view is a square, Triangle is equilateral.
   case triangle
-  case wave
-  case parallelogram
+  
+  /// Wave shape, with
+  /// direction (up or down, default : up)
+  /// width (default 40)
+  /// offset (default : 0)
+  case wave(direction:WaveDirection?, width:Float?, offset:Float?)
```
tbaranes, line 26: If we document, we should use swift documentation guidelines in order to have the quick view documentation... otherwise, don't document at all (just in the github docs)

IBAnimatable/MaskType.swift
```diff
@@ -4,12 +4,68 @@
...
+    
+    let paramsAndName = MaskType.extractNameAndParams(from: string)
+    let name = paramsAndName.name
+    let params = paramsAndName.params
+    
+    switch name.lowercased() {
+    case "circle":
+      self = .circle
+    case "polygon":
+      self = .polygon(sides: Int(params[safe:0] ?? ""))
```
tbaranes, line 57: Shouldn't be the default params instead of empty string?

IBAnimatable/Utils.swift
```diff
@@ -0,0 +1,27 @@
...
+//  IBAnimatableApp
+//
+//  Created by jason akakpo on 03/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import Foundation
+
+internal extension Array {
+  /// Returns the element at the specified index iff it is within bounds, otherwise nil.
```
tbaranes, line 12: Typo: `iff`-> `if`

IBAnimatable/Utils.swift
```diff
@@ -0,0 +1,27 @@
...
+  subscript(safe index: Int ) -> Element? {
+    return indices.contains(index) ? self[index] : nil  /// Returns the element at the specified index iff it is within bounds, otherwise nil.
```
tbaranes: Typo: `iff` -> `if`
Since it's a duplication of the same comment, is it really useful?

tbaranes:
Just did a first review, it looks really great for a first version 👍 
I let you some feedbacks / questions, can you give it a look?

Two general points:
- Do you have (SwiftLint)[https://github.com/realm/SwiftLint] installed? I'm surprised that it didn't report some space warnings for `variable:CGFloat`.
- We don't have any rules about comments but I think if we want to document code (params...), we should follow swift documentation in order to have Xcode quick look compatibility. Otherwise, I don't think it's useful to document that much, that will just make the code harder to read. Any opinions though this @JakeLin ?

IBAnimatable/MaskDesignable.swift
```diff
@@ -6,47 +6,34 @@
...
+    case .circle:
+      maskCircle()
+    case .parallelogram(let angle):
+      maskParallelogram(angle ?? 60)
+    case .polygon(let sides):
+      maskPolygon(sides ?? 6)
+    case .star(let points):
+      maskStar(points ?? 5)
+    case .wave(let direction, let width, let offset):
+      maskWave(direction ?? .up, waveWidth: width ?? 40.0, waveOffset:offset ?? 0.0)
```
tbaranes: Already comment this somewhere else, but I definitely think it will be worth to set the default value in the enums instead of here. It will be easier to get what is the default!

IBAnimatable/MaskDesignable.swift
lastMove, line 181: No, I had installed SwiftLint too.

tbaranes, line 181: There should be an issue normal, SwiftLint syntax is `x: 0, y: 0` instead of `x:0,y:0`. It would be worth it to find out why there's no warnings about this

IBAnimatable/MaskType.swift
lastMove, line 26: Yes, All the MaskType documentation is shown in the Xcode 8 quick view documentation ;)

IBAnimatable/MaskType.swift
lastMove, line 57: I wanted the default params to be on only one place. 
To avoid redundant code and prevent errors. 
Since this method is Used only when the MaskType is initialised from String (not when we do `maskType = MaskType.polygon(sides:8)` )

IBAnimatable/Utils.swift
lastMove, line 12: iff is here for "If and only if" but if this naming is not enough widespread I can change it.

IBAnimatable/MaskDesignable.swift
lastMove: Ideally I would set the default here 

```Swift
public enum MaskType: IBEnum {
  
  /// Circle shape with diameter equals to min(width, height)
  case circle
  
  ///  Polygon shape with n sides. (min 3, default)
  case polygon(sides:Int? = 3)
```
But it's not allowed by Swift. 
I couldn't put it in the init   `init(string: String?)` since it's used only when MaskType is init from String. 
That's why I put it her. 
I also don't really like it but where could we put it elsewhere ?

IBAnimatable/IBEnum.swift
lastMove: Yep, My bad, will change it.

lastMove:
Ideally I would set the default here 

``` Swift
public enum MaskType: IBEnum {

  /// Circle shape with diameter equals to min(width, height)
  case circle

  ///  Polygon shape with n sides. (min 3, default)
  case polygon(sides:Int? = 3)
```

But it's not allowed by Swift. 
I couldn't put it in the init   `init(string: String?)` since it's used only when MaskType is init from String. 
That's why I put it there. 
I also don't really like it but where could we put it elsewhere ?

JakeLin:
@tbaranes @lastMove hey guys. Based on @lastMove 's comments, I am going to update them to follow the Swift documentation standard. Will ask you guys to review it later.

IBAnimatable/AnimatableStackView.swift
```diff
@@ -110,6 +110,7 @@ import UIKit
     }
   }
   
+  /// The mask type used in Interface Builder. **Should not** this property in code.
```
tbaranes: `**Should not** this` -> Not sure about this, but it sounds weird, isn't missing a verb?

JakeLin:
@lastMove and @tbaranes, Can you guys have a look at the new commit https://github.com/JakeLin/IBAnimatable/pull/237/commits/f7c6a0e25454ba6af0d00ccbe0995f22c74331d8 I have updated all Mask related enum, protocol and methods with Swift documentation comments. Please feel free to give me some feedback since I am still learning how to write good documentation comments. 

Also, I have improved some API to make them more Swifty.

@lexrus Can you have a look at this PR if you have time. We discussed how to support enum in Interface Builder before. That is the first attempt. It looks very promising.

IBAnimatable/CardsAnimator.swift
```diff
@@ -140,7 +140,7 @@ private extension CardsAnimator {
       var t1 = CATransform3DIdentity
       t1.m34 = 1.0 / -900
       t1 = CATransform3DScale(t1, 0.95, 0.95, 1)
-      t1 = CATransform3DRotate(t1, 15.0 * CGFloat(M_PI) / 180.0, 1, 0, 0)
+      t1 = CATransform3DRotate(t1, 15.0 * .pi / 180.0, 1, 0, 0)
```
tbaranes, line 143: You used `pi` directly above, let's do this everywhere! :)

IBAnimatable/MaskDesignable.swift
```diff
@@ -5,71 +5,112 @@
...
+      path.addLine(to: point(from: angle, radius: radius, offset: center))
     }
     path.close()
     return path
   }
   
   // MARK: - Star
   
-  // See https://www.weheartswift.com/bezier-paths-gesture-recognizers/
-  private func maskStar(_ points: Int) {
```
tbaranes: Didn't look all the code, but is there a reason to remove the `private` access?
(same question for the others)

IBAnimatable/MaskDesignable.swift
```diff
@@ -78,26 +119,30 @@ public extension MaskDesignable where Self: UIView {
...
   }
   
-  private func starPath(_ points: Int, borderWidth: CGFloat = 0) -> UIBezierPath {
+  /**
+   Get a Bezier path for a star shape with provided points.
+   
+   - Parameter sides: The number of the star points.
+   - Returns: A Bezier path for a star shape.
+   */
+  func getStarPath(with points: Int, borderWidth: CGFloat = 0) -> UIBezierPath {
```
tbaranes: I don't know if we should document that. It should be private and not usable outside of `IBAnimatable`, so...

IBAnimatable/CardsAnimator.swift
JakeLin, line 143: I don't get this one. Do you mean we should use `pi` instead of `.pi`? We need to have the dot.

tbaranes, line 143: We should use the same syntax everywhere: `pi` or `.pi`. If we can do directly `pi`, that's even better! I notice that you used both syntax in `MaskDesignable`

IBAnimatable/AnimatableStackView.swift
JakeLin: Fixed, thanks.

IBAnimatable/CardsAnimator.swift
JakeLin, line 143: I prefer to use `.pi` since it is from Apple. Only one place I didn't as below because I need to avoid a compilation error.

```
let pi: CGFloat = .pi
animation.fromValue = clockwise ? 0 : pi * 2
animation.toValue = clockwise ? pi * 2 : 0
```

IBAnimatable/MaskDesignable.swift
JakeLin: I just grouped them within a private extension. We don't have to document `private` functions, I just tried out and practised the documentation comments.

IBAnimatable/CardsAnimator.swift
tbaranes, line 143: The syntax `pi` without dot isn't a Swift 3 new feature? I don't have any preference about this, so up to you 👌

IBAnimatable/MaskDesignable.swift
tbaranes: Oh, github changes put them in a public extension... my bad

IBAnimatable/MaskDesignable.swift
lastMove, line 181: Yes, strange.

IBAnimatable/MaskType.swift
```diff
@@ -6,33 +6,33 @@
...
+ 
+ - circle: For circle shape with diameter equals to min(width, height).
+ - polygon: For polygon shape with `n` sides. (min: 3, the default: 6).
+ - star: For star shape with n points (min: 3, default:6)
+ - triangle: For isosceles triangle shape. The triangle's height is equal to the view's frame height. If the view is a square, the triangle is equilateral.
+ - wave: For wave shape with `direction` (up or down, default: up), width (default: 40) and offset (default: 0)
+ - parallelogram: For parallelogram shape with an angle (default: 60). If `angle == 90` then it is a rectangular mask. If `angle < 90` then is a left-oriented parallelogram \\-\\
+ - none: For no mask, if the input `string` of `init` is nil, the value of the neum is `.none`. If the `string` can not be mapped to supported mask type, it will set to default value `.none`.
+ */
+
```
lastMove, line 20: @tbaranes @JakeLin 
This documentation style doesn't work well on my side : 
For MaskType I have just the first line : 
screen shot 2016-07-06 at 17 41 03

And for the cases nothing : 
screen shot 2016-07-06 at 17 41 14

Does it work on your side ? 

That's why I decided to use the "///" style for each case which work relatively well : 

``` Swift
 /// Circle shape with diameter equals to min(width, height)
  case circle
```

screen shot 2016-07-06 at 17 42 07

IBAnimatable/CardsAnimator.swift
JakeLin, line 143: I think `.pi` with a dot is default from Apple, let's use it if possible.

IBAnimatable/MaskType.swift
JakeLin, line 20: @lastMove It works for me when I hover on `MaskType`. It displays the title and all enum options. But doesn't work when hover on `.circle`. I will check Swift Standard Library to see how do they do it for the `enums. We can adopt what they do.

lastMove, line 20: Ok, nice.

lastMove:
Except of the  Documentation point ; It looks very good for me ;)

IBAnimatable/MaskType.swift
lastMove, line 20: We could simply duplicate it like that :

```
/**
 Mask type for masking an IBAnimatable UI element.

 - circle: For circle shape with diameter equals to min(width, height).
 - polygon: For polygon shape with `n` sides. (min: 3, the default: 6).
 - star: For star shape with n points (min: 3, default:6)
 - triangle: For isosceles triangle shape. The triangle's height is equal to the view's frame height. If the view is a square, the triangle is equilateral.
 - wave: For wave shape with `direction` (up or down, default: up), width (default: 40) and offset (default: 0)
 - parallelogram: For parallelogram shape with an angle (default: 60). If `angle == 90` then it is a rectangular mask. If `angle < 90` then is a left-oriented parallelogram \\-\\
 - none: For no mask, if the input `string` of `init` is nil, the value of the neum is `.none`. If the `string` can not be mapped to supported mask type, it will set to default value `.none`.
 */

public enum MaskType: IBEnum {
  /// For circle shape with diameter equals to min(width, height)
  case circle
  /// For polygon shape with `n` sides. (min: 3, the default: 6).
  case polygon(sides:Int?)
  case star(points:Int?)
  case triangle
  case wave(direction:WaveDirection?, width:Float?, offset:Float?)
  case parallelogram(angle:Double?)
  case none
´´´´
```

JakeLin:
@lastMove I haven't found a good example from the swift-corelibs-foundation project. They don't have many comments. I leave that for now and we can update the comments any time once we find some good examples.

IBAnimatable/MaskType.swift
JakeLin, line 20: @lastMove duplicated comments are good for the user, but hard for us to maintain. I still want to see how Apple document their code and follow their standard. Please let me know if you find one.

lastMove:
I finally found one useful exemple not from The Swift Open-Source project but from the auto-generated UIKit header. All enums docs from UIKit (That I found) looks like this : 

``` Swift

/* Distribution—the layout along the stacking axis.

 All UIStackViewDistribution enum values fit first and last arranged subviews tightly to the container,
 and except for UIStackViewDistributionFillEqually, fit all items to intrinsicContentSize when possible.
 */
@available(iOS 9.0, *)
public enum UIStackViewDistribution : Int {


    /* When items do not fit (overflow) or fill (underflow) the space available
     adjustments occur according to compressionResistance or hugging
     priorities of items, or when that is ambiguous, according to arrangement
     order.
     */
    case fill


    /* Items are all the same size.
     When space allows, this will be the size of the item with the largest
     intrinsicContentSize (along the axis of the stack).
     Overflow or underflow adjustments are distributed equally among the items.
     */
    case fillEqually


    /* Overflow or underflow adjustments are distributed among the items proportional
     to their intrinsicContentSizes.
     */
    case fillProportionally


    /* Additional underflow spacing is divided equally in the spaces between the items.
     Overflow squeezing is controlled by compressionResistance priorities followed by
     arrangement order.
     */
    case equalSpacing


    /* Equal center-to-center spacing of the items is maintained as much
     as possible while still maintaining a minimum edge-to-edge spacing within the
     allowed area.
        Additional underflow spacing is divided equally in the spacing. Overflow 
     squeezing is distributed first according to compressionResistance priorities 
     of items, then according to subview order while maintaining the configured 
     (edge-to-edge) spacing as a minimum.
     */
    case equalCentering
}

```

So if we apply this to our purpose, it will be something like that : 

``` Swift
/**
 Mask type for masking an IBAnimatable UI element.
 */
public enum MaskType: IBEnum {
  /// For circle shape with diameter equals to min(width, height)
  case circle
  /// For polygon shape with `n` sides. (min: 3, the default: 6).
  case polygon(sides:Int?)
  ///  For star shape with n points (min: 3, default:6)
  case star(points:Int?)
  /// For isosceles triangle shape. The triangle's height is equal to the view's frame height. If the view is a square, the triangle is equilateral.
 case triangle
  /// For wave shape with `direction` (up or down, default: up), width (default: 40) and offset (default: 0)
  case wave(direction:WaveDirection?, width:Float?, offset:Float?)
  /// For parallelogram shape with an angle (default: 60). If `angle == 90` then it is a rectangular mask. If `angle < 90` then is a left-oriented parallelogram \\-\\
  case parallelogram(angle:Double?)
 /// For no mask, if the input `string` of `init` is nil, the value of the neum is `.none`. If the `string` can not be mapped to supported mask type, it will set to default value `.none`.
  case none
}
```

@tbaranes @JakeLin 
What do you think about using this for comment enum ?

JakeLin:
@lastMove Cool, I think we should follow that.

Issue #221, opened 2016-06-15
Swift 3 support


JakeLin:
Discuss the topics for Swift 3 support. We are working on [swift3 branch](https://github.com/JakeLin/IBAnimatable/tree/swift3) to support Swift 3. Once Xcode 8 RTM, we merge it to master.
- [x] Fix all build errors #222
- [x] Improve all enums #218 
- [x] More Swifty APIs
- [x] Update playground

Some remain works need to be done
- [x] BorderSide in the demo doesn't work
- [x] Blur effect in playground app doesn't work - It doesn't change when we change "Blur"
- [x] Predefined gradients in playground app doesn't work 
- [x] Gradients in playground app doesn't work 
- [x] Mask option "None" in playground app doesn't work 
- [x] Transition in playground app crashes
- [x] Container Transition in playground app doesn't work 
- [x] Fix all purple colors
- [x] Presentation transition doesn't work
- [x] Upgrade activity indicators
- [x] Fix all factor methods with `make***`
- [x] Fix all methods from `config***` to `configure***`
- [x] Fix all `unwrapped***` variable.
- [x] Remove `repeatCount` in `Animatable`

cemolcay:
can you update xcode 8beta6 please?

JakeLin:
@lastmove has taken the initiative on this, please have a look at https://github.com/IBAnimatable/IBAnimatable/commit/22475b3911ef6b11938ecac58fb21f19442a957b

If you have time, could you please help us to check the changes? we also discuss a big change about access scope on #285

vincentlong889:
Swift 3.0 Released!

learnwithgabbar:
Not getting updated code on 2.8.1 
I updated code through pod but not getting the updated code, It giving error in every file

tbaranes:
v2.8.1 is swift 2.3 support, if you want swift 3, you have to target the master directly, the swift 3 release will arrive this week or next one.

learnwithgabbar:
Okay

JakeLin:
Finally, we close this now, well done guys.

PR #239, opened 2016-07-08
Upgrade to Xcode 8 beta2

JakeLin:
In this PR, we upgrade the project to support Xcode 8 beta 2. And also make some APIs more Swifty.

Related to #221


tbaranes:
👍

JakeLin:
@tbaranes thanks for your review. We are on Xcode 8 beta 2 now.

Issue #198, opened 2016-05-08
AnimatablePresentationController


JakeLin:
`AnimatablePresentationController` can support the user in prototyping and designing Modal View ( `UIPresentationController`) in Interface Build. 

It may have the following properties.
- [x] `cornerRadius`, `shadowColor`, `shadowRadius`, `shadowOpacity`, etc. for the Modal View, most of properties are supported by `AnimatableView` 
- [x] `dimmingViewBackgroundColor`, `dimmingViewOpacity`, `dimmingViewBlurEffect`, etc for background dimming view
- [x] `animationType`, e.g. `Pop(direction)` and all supported transition animators
- [x] `dismissGestureType`, e.g. `Swipe`, `Tap`
- [x] Custom size / position
- [x] `observeKeyboard`, up the modal when keyboard is opening
- [ ] `useDynamics`, use UIKit Dynamics when panning the view
- [x] `OverCurrentContext`  / fullscreen

Just some initial idea, feedback is welcome. Don't know which Milestone will work on it, so put "help wanted" 😉

Relative to #172, #281 and #20

tbaranes:
There's definitely some ideas to take from [MZFormSheetPresentationController](https://github.com/m1entus/MZFormSheetPresentationController).

For the milestone, why not add it to the 2.4, or 4.0 if we don't have a 2.4?

JakeLin:
@tbaranes I tag it to 4.0, but can be done earlier if needed.

TheDarkCode:
@JakeLin I might take this on - in some part - during the upcoming week, depending on time. Would be a really cool feature to have.

JakeLin:
Thanks guys, should we swap 4.0 with 3.0? I don't have many ideas on the goal of Sketch plugin. If yes, we can just change the number in the milestone.

tbaranes:
I think so, we can still move on the plugins once we determine a clear roadmap for them.

JakeLin:
@tbaranes I have swapped 3.0 and 4.0. Let's don't worry about Sketch plugin first.

tbaranes:
I'd like to add more examples to cover all the features supported by our presentation controller. Which one should be the best:
- A list containing different ready use cases (like the current transition examples): animations, custom dimming view, ... we can easily create at least 20 examples using a list
- Creating a list of textfield for each customisation available, and creating the popup based on the values entered by the user

Any preferences?

JakeLin:
[...]

PR #247, opened 2016-07-16
Animatable presentation controller

tbaranes:
Related to #198 
This PR won't be merged in master but in `feature/presenter` since it's not ready to be released. There's still mandatory features missing, and that first PR implementing mainly the basics. Once merged, it will be easier to move forward.

**New in this PR:**
- Introducing `PresentationDesignable` which handle all the designable part of the `presentedVC`. Currently supporting: `cornerRadius`, `backgroundColor`, `opacity`, `blur`, `shadow` and `dismissOnTap`. Useful to customise the `dimmingView` and `presentedView`
- Introducing `PresentationAnimatable` which inherits of `UIPresentationController` and will be in charge of managing the present / dismiss animations, dimming view... and all the setup around presentation
- Introducing `AnimatablePresentatedViewController`, as the others, it just the `Animatable` type to present a controller
- Example project updates in order to test this feature

**Current state:**
- You can present an `AnimatablePresentatedViewController` programatically or using segues
- Every variables are customisable programatically as well as using IB
- Support 6 present / dismiss transitions: `Cover(Left, Right, Top, Bottom)`, `CrossDissolve`, `Flip`, `Invert`
- The `presentViewController` will always be presented in fullscreen, no support for custom size / position (next step)

**Questions / though:**
- Why `TransitionAnimatable` is an `Animatable` instead of `Designable`?
- I'm not really satisfy with the naming. Since the transitions and presentation view controller are together, it's complicated to understand which is what (when respecting apple's naming). There still some improvements to do on this
- How filtering animations for transitions and for presentations? For example, `Fade(.In)` must be available for the both, but in the current master implementation, it will work only for transitions because a presentation doesn't always have a `fromView`. So, two solutions:
  - Fix all the animations to support transitions as well as presentation. [Implementation of this solution for Fade](https://github.com/JakeLin/IBAnimatable/pull/247/commits/604f87421fcb413edc6e18342bf76d00c1695b59)
  - Create a new enums dedicated for presentation (which means a lot of duplication if we want the same animations), but it will make the code cleaner and the feature easier to understand.

Anyway, that's a first version, and there still a lot of work to do on it. Any feedbacks welcome!
Once that PR merged, it will be really easier to work all together on this :)


lastMove:
Very useful feature !
Concerning your interrogations, I think the best is to adapt all animations to support both transitions and  Presentations.
It will be simpler and more maintainable

tbaranes:
> Concerning your interrogations, I think the best is to adapt all animations to support both transitions and Presentations.

That's what I though first, but I changed my mind about this. Sadly, most of the current animators won't be reusable for a presentation animation because they are based on the fact that we have a `fromVC` and a `toVC` which is not the case when presenting...

So, if we keep solution 1, it will be hard for the user to know which enum they can use or not following if they are using a custom transition or presentation.

tbaranes:
In the last commit, I implemented the second option (duplicate and differentiate a transition and a presentation instead of abstract them). If you want to see the main files diff, you should compare the following ones:
- `PresentationPresenter` vs `TransitionPresenter`
- `PresentationPresenterManager`vs `TransitionPresenterManager`
- `PresentationAnimationType` `TransitionAnimationType`
- `AnimatedPresenting` `AnimatedTransitioning`
  Each versus are pretty similar, but has subtil difference. As said above, most of the animators `TransitionAnimationType` can't be reused as a `PresentationAnimationType`. Presentation animations are really simpler, and there's not a lot of animation possible. 

Whichever the solution we choose in the end, we will need a context: are we doing a transition or a presentation? We can't escape from that, some part of the interfaces are completely related.

tbaranes:
Closing in favour of #248

PR #248, opened 2016-07-17
Animatable presentation controller

tbaranes:
Related to #198, previous PR #247 (for the history)
This PR is the first step of custom presentation controller support in IBAnimatable. In the current state, the feature is ready to go since the main features are implemented.

**New in this PR:**
- Introducing `PresentationDesignable` which handle all the designable part of the `presentedVC`
- Introducing `PresentationAnimatable` which inherits of `UIPresentationController` and will be in charge of managing the present / dismiss animations, dimming view... and all the setup around presentation
- Introducing `AnimatablePresentatedViewController`, as the others, it just the `Animatable` type to present a controller
- Introducing `PresentationModalSize` and `PresentationModalPosition` which are used to customise the modal size and position once presented
- A section "Presentation" has been added to the project example in order have the different animation in demonstration. 

**Features supported:**
- `AnimatablePresentatedViewController` presentation programatically or using segues
- `AnimatablePresentatedViewController` customisations (programatically or in IB):
  - `cornerRadius` (dimminView)
  - `dismissOnTap`
  - `backgroundColor`
  - `opacity`
  - `blurEffectStyle`
  - `blurOpacity`
  - `shadowColor`
  - `shadowRadius`
  - `shadowOpacity`
  - `shadowOffset`
-  6 present / dismiss animations
  - `Flip`
  - `CrossDissolve`
  - `Cover(Top, Right, Bottom, Left)`
  - `Invert`
- Custom position:
  -  `Center`
  - `TopCenter`
  - `BottomCenter`
  - `LeftCenter`
  - `RightCenter`
  - `CustomCenter(centerPoint: CGPoint)`
  - `CustomOrigin(origin: CGPoint)`
- Custom size (width and height are set separately)
  - `Half`
  - `Full`
  - `Custom(size: Float)`

**Missing:**
- More options: `keyboard observer`, `UIKit Dynamics`,  `context support`... but we can add them progressively by adding new options.
- More examples: currently showing only the transition, but it could be worth to have example playing with the dimming view customisation, as well as custom position / size modal
- Documentation in CHANGELOG and README 

**Questions / though:**
- Why `TransitionAnimatable` is an `Animatable` instead of `Designable`? (not really related with this PR, but just wondering)
- The global naming can be improved for sure, if you have any suggestions... !
- Discussion about how to differentiate `Presentation` and `Transition` code. Read #247 comments to have a full context

I know that is a big PRs with a lot of changes, but I'm counting on you all to take a look at the code and give me some feedbacks 😄


JakeLin:
@tbaranes I need more time to go through such a big PR, thanks.

JakeLin:
@tbaranes we may keep this great feature for 2.7 as the first go of IBAnimatable organization.

IBAnimatable/FadeAnimator.swift
```diff
@@ -64,9 +66,9 @@ extension FadeAnimator: UIViewControllerAnimatedTransitioning {
         case .In:
           toView.alpha = 1
         case .Out:
-          fromView.alpha = 0
+          tempfromView?.alpha = 0
```
JakeLin, line 69: Why don't we use unwrapped `fromView`?

tbaranes, line 69: That was an example if we want to use the same animators (in the final implementation, I don't the same), but in this case: when we present a controller, we don't always have a `fromView`, which will make it fails on the guard with the previous implementation, even for a `Fade` or `FadeIn`.

If we choose to keep going with the last implementation, I will rollback that code.

IBAnimatable/AnimatablePresentatedViewController.swift
```diff
@@ -0,0 +1,166 @@
+//
+//  Created by Tom Baranes on 16/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public class AnimatablePresentedViewController: UIViewController, PresentationDesignable {
```
JakeLin: Using `AnimatablePresentedViewController` as an API is a little bit weird for me. I may confuse you since I used `TransitionPresentedViewController` in the demo app.  Becuase when we use `Present` and `Push` transitions, there also have `PresentedViewController`. How about `AnimatableModalViewController` or other similar? On iPad, sometimes, we use Popover, So I am not sure `Modal` is a good name for both iPhone or iPad? How do you think?

IBAnimatable/AnimatedPresenting.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  Created by Tom Baranes on 16/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public typealias FrameTransformer = (finalFrame: CGRect, containerFrame: CGRect) -> CGRect
+
+public protocol AnimatedPresenting: UIViewControllerAnimatedTransitioning {
```
JakeLin: We may be able to have an abstract protocol like `ViewControllerAnimatedTransitioning` (name can be varied) to encapsulate `transitionDuration`, `func retrieveTransitionDuration(transitionContext: UIViewControllerContextTransitioning?) -> NSTimeInterval`, `retrieveViews` and `retrieveViewControllers`, then let `AnimatedPresenting` and `AnimatedTransitioning` conform to it.

IBAnimatable/InvertAnimator.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  Created by Tom Baranes on 16/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public class ZoomAnimator: NSObject, AnimatedPresenting {
```
JakeLin: The file name should be `ZoomAnimator.swift`.

IBAnimatable/PresentationAnimatable.swift
```diff
@@ -0,0 +1,166 @@
+//
+//  Created by Tom Baranes on 16/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public class PresentationAnimatable: UIPresentationController {
```
JakeLin: Because it is a subclass of `UIPresentationController`, and `PresentationAnimatable` sounds like a protocol. Shall we call it `AnimatablePresentationController`, I am thinking the user can use `AnimatablePresentationController` without using `AnimatablePresentedViewController`, for example, a normal `UIViewController` as a modal view controller. I like the idea to have `PresentedSetup`. The user can specify the present effect by providing this input parameter.

IBAnimatable/PresentationDesignable.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Created by Tom Baranes on 16/07/16.
+//  Copyright © 2016 Jake Lin. All rights reserved.
+//
+
+import UIKit
+
+public protocol PresentationDesignable: class {
+
+  var presentationAnimationType: String? { get set }
+  var dismissAnimationType: String? { get set }
```
JakeLin: May call `dismissalAnimationType` as the other Apple's API.

IBAnimatable/PresentationDesignable.swift
```diff
@@ -0,0 +1,45 @@
...
+  var blurEffectStyle: String? { get set }
+  var blurOpacity: CGFloat { get set }
+
+  var shadowColor: UIColor? { get set }
+  var shadowRadius: CGFloat { get set }
+  var shadowOpacity: CGFloat { get set }
+  var shadowOffset: CGPoint { get set }
+}
+
+public struct PresentedSetup {
```
JakeLin: I really like this struct. I am thinking the name of it. `Setup` is also a noun, but I feel `configuration` maybe better for a struct. How about `PresentionConfiguration`? It's very challenging for me to pick up a good name, 😂

JakeLin:
I like the PR very much, the presentation effects are great. Well done @tbaranes , I have some comments on naming. Please have a look at those comments and let me know you feedback, don't have to change them all. 

> Why TransitionAnimatable is an Animatable instead of Designable? (not really related with this PR, but just wondering)

The reason I picked this name because it has `transitionAnimationType` for animations. I was thinking something like `Transitionable` or `Transitable`, but they are not actual words.  How do you think? I don't mind to change it. 

> The global naming can be improved for sure, if you have any suggestions... !

Please see my comments. I am also not 100% sure though.

> Discussion about how to differentiate Presentation and Transition code. Read #247 comments to have a full context

I like your idea to separate them. Otherwise, all transition animators should support both Presentation and Transition, but some of the animators should not use in Presentation and vice versa. 

To better to reuse the code, I think we need to find a way to share the animation effects.

In summary, you have done a great job, 👍

IBAnimatable/AnimatablePresentatedViewController.swift
tbaranes: In fact, I used `Presented` because Apple uses `Presented`, but I'm joining you to say that it can be confusing. I'm not fan of `AnimatableModalViewController` since it's restrictive, but there definitely a better name to find.

IBAnimatable/InvertAnimator.swift
tbaranes: 😨 Wuuups

tbaranes:
I agree with you with almost every comments! Just one where I still think that we have to find a better name. Anyway, with all these changes, we are going in the right direction 👯 

> The reason I picked this name because it has transitionAnimationType for animations. I was thinking something like Transitionable or Transitable, but they are not actual words. How do you think? I don't mind to change it.

Make sense. We can keep it, I was just curious.

> To better to reuse the code, I think we need to find a way to share the animation effects.

I think we can think about this later following which transitions are we integrating. At the beginning, I thought that too, but there almost any transitioning transitions that can be reused for a presentation (only Fade, Slide, ...). In the current version, there no duplicate, so no issues! 
If we really want to reuse the animators, it won't be too complex, in fact, `FadeAnimator` is not used for presentation since we have the system animation, but it can already be used for the both cases.

JakeLin:
@tbaranes Well, you have addressed all comments 👏  Yes, we are heading in the right direction. Agree with you on "reusing code for both Presentation and Transition". There is no duplicated code now. We can think about it how to do it once we have.

BTW: `Modal Position`, `Modal Width` and `Modal Height` are not working for me, I will debug them and see.

I will spend more time on this PR and may make more comments 😬

IBAnimatable/AnimatablePresentatedViewController.swift
JakeLin: I think `AnimatableModalViewController` may be an option because Apple use  [UIModalPresentationStyle](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewController_Class/#//apple_ref/c/tdef/UIModalPresentationStyle), it consist of some options like `PageSheet`, `FullScreen` and `Popover`, another idea is `AnimatablePresenterViewController`. I see Apple use `Presented` as input parameters only but not as a class name.  `PresentedViewController` means a running status of a ViewController. Have you found any class name with `Presented`?

tbaranes: I'm now supporting that name because of this argument 😆

tbaranes: Renaming done!

tbaranes:
For your information, I just rebased this branch in order to fix git's conflicts!
About this PR, it still needs documentation and changelog updates, but I don't think we should add any more features in it, the next ones will be add in others PRs following the needs of everyone.

@JakeLin Won't have enough time to check the `Position` / `Size` bugs, but will do it next as soon as possible if you didn't take care about it before 👍

JakeLin:
@tbaranes cool, I don't think we need to add new features to the PR. 

I am looking at the details of implementation. For example, the shadow color is not working as expected. I also noticed some animations are not consistent. I am looking at them and see whether I can fix them. 

The whole design is stable and great. Once we fix all small issues, I think we can merge it to master.

IBAnimatable/AnimatableModalViewController.swift
```diff
@@ -29,79 +29,79 @@ public class AnimatableModalViewController: UIViewController, PresentationDesign
 
   @IBInspectable public var modalPosition: String = "Center" {
     didSet {
-      presenter?.presentedSetup?.modalPosition = PresentationModalPosition.fromString(modalPosition)
+      presenter?.presentionConfiguration?.modalPosition = PresentationModalPosition.fromString(modalPosition)
```
tbaranes: Typo? "presentionConfiguration" -> "presentationConfiguration"

JakeLin: @tbaranes thanks, will fix it.

JakeLin:
@tbaranes I have fixed the typo, thank for pointing it out. I am looking at the last bit of the PR which is the animation. I am still learning from your code 😊. Some animations are not consistent. For example, `Cover(Top)` animation is moving in from the bottom and moving out to **the top**. But  `Cover(Bottom)` animation is running in from the bottom and moving out to **the bottom**. I feel like the `Cover(Bottom)` is right since where it is from should be where it goes.

So I have a question for you, Do you think all dismissal animation direction should be opposite of presentation animation direction. e.g. Move in from left for presentation then move out to left for dismissal?

tbaranes:
I think we should follow apple's pattern: the default modal animation is move in from bottom, and the dismissal one is move out to bottom. So, I don't think the default dismissal animation should be the opposite, but the same just using the same `from` but out.

JakeLin:
@tbaranes I think I have fixed all `Cover` transitions. They are working as Apple's pattern, can you try it out and see?

But I am having some issues with system animations for `CrossDissolve` and `Flip`. If I don't set `dismissalAnimationType`, both will use `Cover(Bottom)` as the dismissal transition. I have spent hours on it and still I don't know why. But if I specify `dismissalAnimationType` as `CrossDissolve` or `Flip`, then their dismissal transitions are working as the preset value. Can you have a look when you have time? There is the last issue I found for this PR. Sorry for reviewing it for a long time, I am heading to bed now 😴

JakeLin:
@tbaranes I have played around this in the train this morning and made a dirty hack on it to make the demo app works properly. I found out if we use system built-in animation e.g. `.CrossDissolve`, `.FlipHorizontal` or `.CoverVertical`, it will set `UIModalTransitionStyle`, once we set it, we can't override it to the other system built-in animation. Looks like it is a system limitation. 

I think the PR is ready to go, can you merge it if you think is ready. Let me know if you have any questions, thanks.

tbaranes:
Awesome @JakeLin! Thanks for your fixed. 
I merged it without the CHANGELOG update, I prefer to update the CHANGELOG once the documentation is done. If we notify our users that is available but can't explain how to use it, it will just create more issues 😨 

I updated the todo list in #198 for the documentation update. Once that done, I think we can release a 2.7 :)

JakeLin:
Thanks, you did a great job on this, it's very challenging to design flexible and extendable APIs. You did that in a very nice way. 👍 we can easily add new transitions for Presentation Controller now 😘

tbaranes:
You did the hardest part by implementing the interface for the transitions 😬

PR #249, opened 2016-07-18
New animations: ZoomInvert{In,Out}

tbaranes:
This PRs add two new animations, they are the reverse of the current `Zoom{In,Out}`. I didn't know how to name this since `Zoom{In,Out}` works for the both result, it's just how we imagine the animation... Any suggestions about a better name? :)

@lastMove Once that PR merged, could you copy it in your branch and merged it in your new enums?


lastMove:
Ok Nice ;) 
Since it still a Zoom, we could maybe add another parameter to Zoom. 
Zoom(in, inverted)
Zoom(in, magnifying)
Zoom(in, enlarging) 

Don't know which one is best. 
and the opposite, maybe :
Zoom(in, shrinking)

JakeLin:
👍 good one, how about we do a release and then merge the release in swfit3?

tbaranes:
Let's do this. It's always nice to have new features!
I didn't change the name, the swift 3 release will break it all anyway, so we can always change it later.

PR #258, opened 2016-07-28
Fix contributors link

mortonfox:
A relative link would go to blob/master/graphs/contributors, which does not exist.


JakeLin:
Thanks, 👍

PR #261, opened 2016-07-30
Blur effect style

lastMove:
This PR is a little bit out of scope off the initial planned features. 

What 's New :
-  remove `BlurEffectStyle` and make `UIBlurEffectStyle` conforming to `IBEnum` 
- Add a Default Implementation for IBEnum where Self: RawRepresentable using iterateEnum. (This default Implementation is case-insensitive). 
- Improve Mask and BlurEffect Demo.
- update `BlurDesignable` : `blurEffectStyle` and `vibrancyEffectStyle` is now of type  `UIBlurEffectStyle?`. it's an optional because `UIBlurEffectStyle` don't have `none` case.


tbaranes:
Just give a quick review, it seems really cool 👍 
Will do another deepest review next week!

JakeLin:
@lastMove Nice one, I will have a deeper look at this soon.

IBAnimatable/BlurEffectStyle.swift
```diff
@@ -3,10 +3,41 @@
...
+import UIKit
 
-public enum BlurEffectStyle: String {
-  case extraLight
-  case light
-  case dark
+
+extension UIBlurEffectStyle : IBEnum {
+  /** initialize a UIBlurEffectStyle from string. `extraLight`, `light`, `dark`, `prominent` (iOS 10+), `extraLight` (iOS 10+)
+    default value is `extraLight`.
```
JakeLin: `default value is nil`?

IBAnimatable/BlurEffectStyle.swift
```diff
@@ -3,10 +3,41 @@
...
+   defaultValue is used when the string is nil or not within the above list
+  */
+  public init?(string: String?) {
```
JakeLin, line 12: since we have `public extension IBEnum where Self : protocol<RawRepresentable, Hashable>` do we still need to override this method?

IBAnimatableApp/BlurEffectViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+  
+  /*
+   // MARK: - Navigation
+   
+   // In a storyboard-based application, you will often want to do a little preparation before navigation
+   override func prepare(for segue: UIStoryboardSegue, sender: AnyObject?) {
+   // Get the new view controller using segue.destinationViewController.
+   // Pass the selected object to the new view controller.
+   }
+   */
```
JakeLin: I prefer to remove all unused code including `didReceiveMemoryWarning`

IBAnimatableApp/BlurEffectViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+  lazy var values: [String] = {
+    var values = ["none", "extraLight", "light", "dark"];
+    if #available(iOS 10.0, *) {
+      values.append(contentsOf: ["prominent", "regular"])
+    }
+    return values
+  }()
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    pickerView.delegate = self
```
JakeLin: I prefer to do all this in Interface Builder. But I leave it to you.

IBAnimatable/AnimatableButton.swift
```diff
@@ -88,7 +88,7 @@ import UIKit
   /// The mask type used in Interface Builder. **Should not** use this property in code.
   @IBInspectable var _maskType: String? {
     didSet {
-      maskType = MaskType(string: _maskType)
+      maskType = MaskType(string: _maskType) ?? .none
```
JakeLin: Do we need `?? .none` since `init(string: String?)` is not failable.

IBAnimatable/IBEnum.swift
```diff
@@ -5,17 +5,19 @@
...
  */
-public protocol IBEnum: StringLiteralConvertible {
+public protocol IBEnum {
   /**
    Initializes a swift `enum` with provided optional string
    
    - Parameter string: The optional string to be converted into `enum`.
    */
-  init(string: String?)
+  init?(string: String?)
```
JakeLin, line 20: Shall we have two `init`s 

```
init?(string: String?)
init(string: String?)
```

because some enum uses `init?(string: String?)` and some uses `init(string: String?)`. I am not sure this one.

JakeLin:
I like this feature to extend system's enums. Great idea and implementation. Apart from the multiple adding blur effect issue, I give this PR 👍 . I like the mask demo as well, well done.

IBAnimatable/BlurEffectStyle.swift
lastMove: Yes, you are right, I will update it.
 Thank you.

IBAnimatable/BlurEffectStyle.swift
lastMove, line 12: Unfortunately it's not applicable in this case. `UIBlurEffectStyle` has rawValue of `Int`, it's the counter part of extending the system `UIBlurEffectStyle`

IBAnimatable/IBEnum.swift
lastMove, line 20: Hum, 
Me too, I am not very happy with this. 
But the problematic is : 
Some Enums (For the moment, `UIBlurEffectStyle`) need to have a failable `init` since there is no `none` case.
But other Enums Like `Masktype` and `AnimationType` don't need failable `init`. 
I first attempted to create a second protocol (IBOptionalEnum) which require a failabile `init`.
But It seems that a class that provides a non-bailable init still conforms to a protocol requiring a failable init.
That's why I decided to let it like that.
What do you think to be the best solution ?

IBAnimatableApp/BlurEffectViewController.swift
lastMove: I will.

IBAnimatableApp/BlurEffectViewController.swift
lastMove: I don't understand, what would you like to do in IB ?

IBAnimatable/AnimatableButton.swift
lastMove: Nope, My bad.

IBAnimatableApp/BlurEffectViewController.swift
JakeLin: @lastMove Usually, I right click the view to a controller and select the delegate and data source
screen shot 2016-08-04 at 9 06 24 am

In this case, it is totally up to you to decide how to use it 😉

IBAnimatable/IBEnum.swift
JakeLin, line 20: @lastMove I like the idea of `IBOptionalEnum`, will play around again and see whether we can have separated protocol.

IBAnimatableApp/BlurEffectViewController.swift
lastMove: Oh I see. 
It's important to have some code consistency ; I will do it your way.

IBAnimatable/IBEnum.swift
JakeLin, line 20: @lastMove I have tried out `IBOptionalEnum`. Please have a look at the commit https://github.com/IBAnimatable/IBAnimatable/commit/0b341304064e2858726b4718670515783573cb36 I think your solution is better 👍

lastMove, line 20: I tried something similar too. One inconvenient is that it can lead to duplicate code (for instance if we want to have a default implementation of `IBEnum` and `IBOptionalEnum` for `RawRepresentable` enums). I prefer to have just one enum for this purpose.

lastMove:
I will wait #264 to be merged then rebase (or merge since I have some difficulties with rebasing 😳 ) and finalise this PR.

JakeLin:
@lastMove I have merged #264, I will revert the main.storyboard later but it won't impact on your PR. 

BTW: please feel free to merge it next time if you think it is ready, you are also a collaborator as well 😉

lastMove:
Thanks, 
I have made the changes according to your comments, and merged back swift3 (#264).

JakeLin:
@lastMove great, please merge it when you think it is ready 👍

lastMove:
Ok :)

PR #268, opened 2016-08-07
Feature/gradient designable

lastMove:
What's New : 
- Introduce `GradientColor` (`typealias GradientColor = (start: UIColor, end: UIColor)`) 
- Add a `color` property that returns the associated UIColor 
- Add a `colors` property that returns the two colors. 
- Update gradientDesignable to uses Enum for `predefinedGradient` and `startPoint`
- Add 2 demo cases for Gradients (one using predefinedGradients, and the other one using 2 `ColorType`)
- Remove useless code.

NB:  
Maybe a bug : 

``` Swift
  gView.predefinedGradient = GradientType.sunnyDays
    gView.startPoint = GradientStartPoint .top
```

This code doesn't apply the gradient. It is necessary to call `configGradient()`
since `configGradient()` is called only within `configAfterLayoutSubviews()` it works when the gradient is set from `viewDidLolad` for instance, but not after.
I don't know if it intended or not.


JakeLin:
I had a quick look, looks very good , I like the new type GradientColor a lot, will take a deeper look tonight.

JakeLin:
@lastMove sorry for the late reply, your PR is all good, 👍 I think we need to call `configGradient()` after we set `predefinedGradient` or `startPoint` after `viewDidLolad` . The reson why we did it in  `configAfterLayoutSubviews` because if we set it eariler, the view bounds may not right.

lastMove:
No problem ;) 
We maybe have to call it also when setting `startColor` and `endColor`

JakeLin:
@lastMove yes, I think we need to call it when setting `startColor` and `endColor`. We may be able to optimize it and don't remove and re-add `gradientView` again and again.

PR #270, opened 2016-08-07
Removes unnecessary force unwraps

gkye:
Removed forced unwrapped where not necessary and to be consistent with
the swift 3 branch


JakeLin:
Thanks guys for fixing this, I think we should fix **force unwrapped** in swift3 branch as well.

PR #275, opened 2016-08-13
Xcode 8 colorLiteral eums

### LINKED ISSUE #274: Update scripts to generate ColorType.swift and GradientType.swift
Update `FlatColorsTypeScript.swift` and `GradientsTypeScript.swift` to generate `ColorType.swift` and `GradientType.swift`. 
- [x] `GradientsTypeScript.swift`
- [x] `FlatColorsTypeScript.swift`

We can generate the code using `#colorLiteral` like `#colorLiteral(red: 0.08235294118, green: 0.6980392157, blue: 0.5411764706, alpha: 1)`, then Xcode 8 can display the color in code editor like

screen shot 2016-06-27 at 1 41 17 pm

gkye:
Updated scripts to swift3 and generates colorLiterals in Xcode 8
Close #274

### DIFF
--- a/IBAnimatable/GradientType.swift
+++ b/IBAnimatable/GradientType.swift
@@ -10,7 +10,7 @@ public typealias GradientColor = (start: UIColor, end: UIColor)
 public enum GradientType: String, IBEnum {
   case coolBrown
   case harmonicEnergy
-  case playingwithReds
+  case playingWithReds
   case sunnyDays
   case greenBeach
   case intuitivePurple
@@ -87,14 +87,14 @@ public enum GradientType: String, IBEnum {
   case redMist
   case tealLove
   case neonLife
-  case manofSteel
+  case manOfSteel
   case amethyst
   case cheerUpEmoKid
   case shore
   case facebookMessenger
-  case soundCloud
+  case soundcloud
   case behongo
-  case servQuick
+  case servquick
   case friday
   case martini
   case metallicToad
@@ -105,24 +105,24 @@ public enum GradientType: String, IBEnum {
   case purpleBliss
   case predawn
   case endlessRiver
-  case pastelOrangeattheSun
+  case pastelOrangeAtTheSun
   case twitch
   case instagram
   case flickr
   case vine
-  case turquoiseflow
+  case turquoiseFlow
   case portrait
   case virginAmerica
   case kokoCaramel
   case freshTurboscent
-  case greentodark
+  case greenToDark
   case ukraine
-  case curiosityblue
+  case curiosityBlue
   case darkKnight
   case piglet
   case lizard
   case sagePersuasion
-  case betweenNightandDay
+  case betweenNightAndDay
   case timber
   case passion
   case clearSky
@@ -132,9 +132,9 @@ public enum GradientType: String, IBEnum {
   case littleLeaf
   case netflix
   case lightOrange
-  case greenandBlue
+  case greenAndBlue
   case poncho
-  case backtotheFuture
+  case backToTheFuture
   case blush
   case inbox
   case purplin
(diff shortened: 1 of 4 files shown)


IBAnimatable/GradientType.swift
```diff
@@ -84,7 +87,7 @@ public enum GradientType: String, IBEnum {
   case redMist
   case tealLove
   case neonLife
-  case manOfSteel
+  case manofSteel
```
tbaranes: Is that change normal? Same question for `greenToDark` (l.118) and `curiosityBlue` (l.120)

Scripts/GradientsTypeScript.swift
```diff
@@ -2,45 +2,69 @@
 
 import Foundation
 
+func getColorLiteral(hexString: String)->String{
```
tbaranes: I think we can make the name more swifty, for example: `colorLiteral(for hexString: String) -> String`
Also, I think SwiftLint doesn't run the script, so don't forget the spaces: `)->String{` -> `) -> String {`

Scripts/GradientsTypeScript.swift
```diff
@@ -2,45 +2,69 @@
...
+
+
+//MARK: - Script
+
 let gradientTypeURL = "https://raw.githubusercontent.com/Ghosh/uiGradients/master/gradients.json"
 
-func JSONFromURL(urlToRequest: String) -> NSData {
-    return NSData(contentsOfURL: NSURL(string: urlToRequest)!)!
+func JSON(_ urlToRequest: String) -> Data {
+  return (try! Data(contentsOf: URL(string: urlToRequest)!))
```
tbaranes: Instead of force unwrap, won't it be better to use a `guard ... else { fatalError(...) }`? If the script fails, it that fails, it would be for the best that the script stop running

Scripts/GradientsTypeScript.swift
```diff
@@ -2,45 +2,69 @@
...
+    print("JSON serialization did fail")
+    return nil
+  }
 }
 
 // Generator constants
 let enumCase = "\tcase %@\n"
 let switchCase = "case .%@:\n"
-let colors =  "\treturn (UIColor(hexString: \"%@\"), UIColor(hexString: \"%@\"))\n"
+let colors =  "\treturn (%@,  %@)\n"
```
tbaranes: `(%@,  %@)`, is there a reason for the two spaces?

Scripts/GradientsTypeScript.swift
```diff
@@ -2,45 +2,69 @@
...
 let endEnumOrSwitch = "}"
 
+
+
```
tbaranes: Why so many new lines? 😆

tbaranes:
I let some feedbacks, mostly code formatting, otherwise, that's a good job @gkye 👍

IBAnimatable/GradientType.swift
gkye: Yup, straight from the [json](https://raw.githubusercontent.com/Ghosh/uiGradients/master/gradients.json), only thing that was changed was the first charatcer to lowercase

tbaranes: Maybe we can improve it then, uppercasing next character after a space? It just feel weird otherwise 😆 
It would be interested to have the others though, @JakeLin @lastMove any opinions?

gkye: You're right. We could use `capitalized` and change the first character later after removing all spaces and e.t.c?

JakeLin:
Agree with @tbaranes , I think `manOfSteel` is more readable than `manofSteel`. Apart from that, 👍

JakeLin:
Agree with @tbaranes . I think `manOfSteel` is more readable than `manofSteel`. Apart from that, 👍

IBAnimatable/GradientType.swift
gkye: Fixed :smile:

JakeLin: That looks better 😄

gkye:
@JakeLin is it okay to merge this? :sweat_smile:

JakeLin:
That's your call @gkye 😉

PR #276, opened 2016-08-13
Improving app main screen

tbaranes:
Since we add the (awesome) playground page in the demo app, I was frustrating by the access point.  Also, I think the playground part is too hard to find if you don't read the docs. This PR is a try to improve it:
- New main screen: when launching the app, instead of getting in the demo, we will get a dispatch one: choose between "Demo app" or "Playground". The "forget password" is used to go back into the main screen. The interface can be improved (animations, colors, images...), but it's a start & proof of concept. 

simulator screen shot 13 aug 2016 10 54 39

I also add few other improvements:
- Uniformize back buttons in the "Placeholder" and "Mask" demo
- Uniformize UITableView separator line in the "User interface" and "Mask" demo
- Set mask selected as title in `MaskViewController`

### DIFF
--- a/Documentation/APIs.md
+++ b/Documentation/APIs.md
@@ -200,7 +200,7 @@ With these methods, we can navigate back or dismiss current ViewController witho
 ### TransitionAnimatable protocol
 | Property name | Data type | Description |
 | ------------- |:-------------:| ----- |
-| transitionAnimationType | Optional&lt;String> | Supported transition animations. Tap on "Forgot Password" button to see all predefined transition animations, e.g. `Fade`, `SystemCube(Left)` and `SystemPageCurl(Bottom)`. The transition type starts with `System` can only use in Push/Pop transitions, not Present/Dismiss transitions. Note: For `SystemRotate` seems that only `SystemRotate(90)` is working. |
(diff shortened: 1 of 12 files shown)


lastMove:
I like the new Main screen look.
Just a remark:  There is a lot of changes within the swift3 branch in the playground part ; 
For instance, the maskViewController has totally been changed.

tbaranes:
I think you can keep your changes in case of conflicts. That PR mostly aims to add the new main screen, the others changes are just minor improvements. We can always apply them manually in the swift3 branch if they are still necessary 😀

JakeLin:
I like it very much, good on you @tbaranes 👍, just a few of suggestions.
1. Change the status bar to light (white)
2. Set the buttons' background color to white with opacity (0.6, the number need to adjust by eye) Text and border are pure white (opacity = 1). That may make the buttons stand out.

JakeLin:
@lastMove I think we need to manually merge the storyboards, I did them for a few of times already. Very time took me hours 😢, I was thinking we should freeze version 2 and make all those changes to swift3 😬

tbaranes:
I updated the buttons, that's a good idea!
For the status bar, it should already be light, for some reason, it is not. I don't know why yet.

Sorry for the storyboards conflicts 😢 As said before, you should definitely keep only the new main screen when you will rebase it, the others are minor changes, it's ok to loose them.

JakeLin:
@tbaranes Don't be afraid of the conflicts, we are making good features, we can always fix those conflicts later.

tbaranes:
Just fixed the status bar, @JakeLin can you take a last look into this PR? 😀

JakeLin:
Good, I should do it from day one since our app all uses white status bar.

PR #278, opened 2016-08-13
Modal translation when keyboard is opening

tbaranes:
Related to #198 
Move the modal when the keyboard is opening. Translation supported:
- `MoveUp`
- `AboveKeyboard`

### DIFF
--- a/IBAnimatable/AnimatablePresentationController.swift
+++ b/IBAnimatable/AnimatablePresentationController.swift
@@ -105,17 +105,20 @@ extension AnimatablePresentationController {
       let presentedFrame = frameOfPresentedViewInContainerView()
       let duration = notification.userInfo?[UIKeyboardAnimationDurationUserInfoKey]?.doubleValue ?? 0.5
       let translatedFrame = presentationConfiguration.keyboardTranslation.translationFrame(keyboardFrame, presentedFrame: presentedFrame)
-      UIView.animateWithDuration(duration) {
+
+      let curve = UIViewAnimationOptions(rawValue: UInt(duration))
+      UIView.animateWithDuration(duration, delay: 0, options: curve, animations: {
         self.presentedView()?.frame = translatedFrame
-      }
+      }, completion: nil)
     }
   }
 
   func keyboardWillHide (notification: NSNotification) {    
     let duration = notification.userInfo?[UIKeyboardAnimationDurationUserInfoKey]?.doubleValue ?? 0.5
-    UIView.animateWithDuration(duration) {
+    let curve = UIViewAnimationOptions(rawValue: UInt(duration))
+    UIView.animateWithDuration(duration, delay: 0, options: curve, animations: {
       self.presentedView()!.frame = self.frameOfPresentedViewInContainerView()
-    }
+    }, completion: nil)
   }
 
 }
--- a/IBAnimatable/AnimatablePresentationController.swift
+++ b/IBAnimatable/AnimatablePresentationController.swift
@@ -23,6 +23,7 @@ public class AnimatablePresentationController: UIPresentationController {
 
     setupDimmingView()
     setupPresentedView()
+    setupObservers()
   }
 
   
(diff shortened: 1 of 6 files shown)


IBAnimatable/AnimatablePresentationController.swift
```diff
@@ -76,12 +75,51 @@ private extension AnimatablePresentationController {
...
+    NSNotificationCenter.defaultCenter().addObserver(self, selector: #selector(AnimatablePresentationController.keyboardWasShown(_:)), name: UIKeyboardWillShowNotification, object: nil)
+    NSNotificationCenter.defaultCenter().addObserver(self, selector: #selector(AnimatablePresentationController.keyboardWillHide(_:)), name: UIKeyboardWillHideNotification, object: nil)
+  } 
+
+  func keyboardWasShown (notification: NSNotification) {
+    if let keyboardFrame = notification.userInfo?[UIKeyboardFrameEndUserInfoKey]?.CGRectValue() {
+      let presentedFrame = frameOfPresentedViewInContainerView()
+      let duration = notification.userInfo?[UIKeyboardAnimationDurationUserInfoKey]?.doubleValue ?? 0.5
+      let translatedFrame = presentationConfiguration.keyboardTranslation.translationFrame(keyboardFrame, presentedFrame: presentedFrame)
+      UIView.animateWithDuration(duration) {
```
JakeLin: I think it is better to get the animation curve and use it as well like 

```
curveInInt = keyboardUserInfo[UIKeyboardAnimationCurveUserInfoKey] as? UInt,
let curve = UIViewAnimationOptions(rawValue: curveInInt)
UIView.animateWithDuration(duration, delay: 0, options: curve, animations: {
       self.presentedView()?.frame = translatedFrame
    }, completion: nil)
```

IBAnimatable/AnimatablePresentationController.swift
```diff
@@ -76,12 +75,51 @@ private extension AnimatablePresentationController {
...
+    }
+  }
+
+  func keyboardWillHide (notification: NSNotification) {    
+    let duration = notification.userInfo?[UIKeyboardAnimationDurationUserInfoKey]?.doubleValue ?? 0.5
+    UIView.animateWithDuration(duration) {
+      self.presentedView()!.frame = self.frameOfPresentedViewInContainerView()
+    }
+  }
+
```
JakeLin, line 123: May add this to remove all notifications

```
deinit {
    if let observer = keyboardWillShowNotificationHandler {
      NSNotificationCenter.defaultCenter().removeObserver(observer)
    }
    if let observer = keyboardWillHideNotificationHandler {
      NSNotificationCenter.defaultCenter().removeObserver(observer)
    }
  }
```

IBAnimatable/PresentationKeyboardTranslation.swift
```diff
@@ -0,0 +1,34 @@
...
+
+import UIKit
+
+public enum ModalKeyboardTranslation {
+  case None
+  case MoveUp  
+  case AboveKeyboard
+
+  static public func fromString(string: String?) -> ModalKeyboardTranslation {
+    if string == "MoveUp" {
```
JakeLin, line 14: Case-insensitive? But all these will be fixed by `IBEnum` in `swift3`

JakeLin:
Very cool and useful feature, well done 👍

IBAnimatable/PresentationKeyboardTranslation.swift
tbaranes, line 14: Yeah I think of this, but since all the enums in master are case-insensitive, I prefer to let it that way since it will be fixed in swift3 with the new implementation :)

JakeLin, line 14: Yes, I am OK with your implementation, and I think the Swift 3 adoption will be very quick.

tbaranes:
The last commits should conforms to your comments @JakeLin, can you take a look?

JakeLin:
All good, great 👍

PR #279, opened 2016-08-13
New presentation transition: DropDown

tbaranes:
Related to #198

### DIFF
--- a/IBAnimatable/DropDownAnimator.swift
+++ b/IBAnimatable/DropDownAnimator.swift
@@ -1,0 +1,101 @@
+//
+//  Created by Tom Baranes on 13/08/16.
+//  Copyright © 2016 IBAnimatable. All rights reserved.
+//
+
+import UIKit
+import QuartzCore
+
+public class DropDownAnimator: NSObject, AnimatedPresenting {
+
+  // MARK: - AnimatedPresenting
+  public var transitionDuration: Duration = defaultTransitionDuration
+  private var completion: AnimatableCompletion?
+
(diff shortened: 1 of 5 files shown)


IBAnimatable/DropDownAnimator.swift
```diff
@@ -0,0 +1,101 @@
...
+    animatingView.layer.addAnimation(animation, forKey: "dropdown")
+  }
+
+  func animateDismissingDropDown(animatingView animatingView: UIView, completion: AnimatableCompletion) {
+    var point = animatingView.center
+    let angle = CGFloat(arc4random_uniform(100)) - 50
+    point.y += UIScreen.mainScreen().bounds.height
+    UIView.animateWithDuration(transitionDuration, animations: {
+      animatingView.center = point
+      animatingView.transform = CGAffineTransformMakeRotation(angle / 100)
```
JakeLin, line 82: `CGAffineTransformMakeRotation` is taking `radians` not `degrees`. Is it what you want here?

JakeLin:
@tbaranes well done 👍, I play around it and it works fine, my only concern is `CGAffineTransformMakeRotation` (please the comment), I don't quite understand your algorithm. In my mind, it should have something like `randomNubmer * .pi`. But the effect is working fine. So the PR is all good.

tbaranes:
I have to say I didn't look deeper since it was working as expecting 😬 
I'm gonna merge this now, if there any improvements necessary, we can still add them later!

JakeLin:
All good @tbaranes thanks

PR #280, opened 2016-08-13
Feature/ Improve Border Sides

lastMove:
`BorderSides` is now a public struct (defined in BorderSide.swift).
`borderSides` property of `BorderDesignable`is now of type `BorderSides` (defaultValue `.allSides`)
Add a `@IBInspectable _borderSides` property. 
NB: `borderSide` become `borderSides`

Add a demo of Border in playground app (User Interface => Border Sides)
screen shot 2016-08-14 at 01 20 03

I noticed using this demo there is a difference between our borders implementation add CALayer's one ; Our implementation use outer-border, system one seems inner. 
 Since we fallback to the system implementation when borderSides == .AllSides, you can saw the difference, by checking-unchecking one side. I Don't know if it intentional  or not ?

### DIFF
--- a/IBAnimatable/BorderSide.swift
+++ b/IBAnimatable/BorderSide.swift
@@ -11,3 +11,55 @@ public enum BorderSide: String {
   case bottom
   case left
 }
+
+
+public struct BorderSides: OptionSet {
+  public let rawValue: Int
+  
+  public static let unknown = BorderSides(rawValue: 0)
+  
+  public static let top = BorderSides(rawValue: 1)
+  public static let right = BorderSides(rawValue: 1 << 1)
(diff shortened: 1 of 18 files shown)


IBAnimatable/BorderSide.swift
```diff
@@ -11,3 +11,55 @@ public enum BorderSide: String {
   case bottom
   case left
 }
+
+
+public struct BorderSides: OptionSet {
```
tbaranes: The file should be renamed too

JakeLin:
@lastMove I noticed the difference as well, but I didn't know the reason. As you mentioned the system one uses inner. Shall use inner (not outer-border) as well to match Apple's implementation. Apart from that, all good, well done 👍

Issue #281, opened 2016-08-14
Releasing new presentation controller


tbaranes:
This issue is related to #198 but is mostly dedicated to list all the points missing in order to get a new release containing our presentation controller feature:
- [x] Documentation: how to use it, which features are supported...
- [x] Changelog
- [x] Impoving demo apps: 
  - [x] animationType (picker view) 
  - [x] dismissalAnimationType (picker view) 
  - [x] modalPosition (picker view)
  - [x] modalSize (picker view)
  - [x] keyboardTranslation (picker view)
  - [x] cornerRadius (slider)
  - [x] dismissOnTap (toggle)
  - [x] backgroundColor (don't know how to present this, color picker or hex code)
  - [x] opacity (slider)
  - [x] blurEffectStyle  (picker view)
  - [x] blurOpacity  (slider)
  - [x] shadowColor (the UI is same as backgroundColor)
  - [x] shadowOpacity (slider)
  - [x] shadowRadius (slider)
  - [x] shadowOffset (two silders?)

_The above list is a copy of @JakeLin comment in the other thread, should be updated following what is necessary_

PR #282, opened 2016-08-15
PresentationController demo

tbaranes:
Related to #281 
No screenshots (the view is too big), check out the branch! It's as we said: sliders, pickers... everything is customisable before looking at our result!
Next step: documentation (😨)

**Note:** I didn't let the `Custom` size / position, I think it's overkill to have a `Custom` in a picker.


JakeLin:
@tbaranes thanks, I like the demo especially the color slider 👍.

I have a few comments
1. Have changed the back button and the segue to make it consistent as the other demo.
2. Shall we change the "Done" to "Present"? You can make the call 😉
3. The enums like `animations`, `keyboardTranslations` and `blurEffectStyles`. Shall we change the empty ("") to something more obviously like "none", "not set" or "N/A"? Because it is hard for me to find "" in the picker view. In swfit3, we will have `.none` for all custom enums. Maybe we can use "none".
4. `Opacity` and `Shadow` seem not working 😢

tbaranes:
Thanks for the feedback @JakeLin! I changed to reflect 2 and 3, I agree with you, that's better.
For the `opacity`, and `shadow`, I have to take time to look into it, will do it later, but nice catch 👍

JakeLin:
@tbaranes cool.

lastMove:
Very nice Demo !! 
Well done, It 's fun to use :) 
I was a little bit confused because of the "Present" button. It can be useful to add another present button at the end of the scroll (it's totally subjective). 
Great job.

tbaranes:
That's a good idea @lastMove! I just added it, a new button is waiting for you at the bottom of the scrollview. I didn't have that much ideas for the design, so it's an orange button 💩 (in fact, it's not that bad... but if you have preferences, let me know)

tbaranes:
@JakeLin The opacity bug was just an error in the setup (the value wasn't set), but I have a doubt about the implementation of that one, in order to make it work, I apply the `opacity` directly on the `backgroundColor` when setting `fillColor` instead of using `opacity`. When using `opacity`, the alpha is always 0.

Should it be that way? 🤔

About the shadow, I take a look into it, I have no idea why is not working. All the values are well-settled, and the layer seems good, so I will go back on it later, but if anyone has an idea... 😨

IBAnimatableApp/PresentingViewController.swift
```diff
@@ -0,0 +1,279 @@
...
+    presentedViewController.shadowRadius = CGFloat(sliderShadowRadius.value)
+    presentedViewController.shadowOffset = CGPoint(x: CGFloat(sliderShadowOffsetX.value), y: CGFloat(sliderShadowOffsetY.value))
+
+    setDismissalAnimationTypeIfNeeded(presentedViewController)
+  }
+
+  private func setDismissalAnimationTypeIfNeeded(viewController: AnimatableModalViewController) {
+    // FIXME: Dirty hack to make `Flip` and `CrossDissolve` work properly for dismissal transition.
+    // If we don't apply this hack, both the dismissal transitions of `Flip` and `CrossDissolve` will slide down the modal not flip or crossDissolve(fade).
+    if viewController.presentationAnimationType == "Flip" {
```
JakeLin, line 96: Shall we check dismissalAnimationType is nil here?

tbaranes, line 96: I don't think, if it's nil, the condition won't be fulfilled

JakeLin:
I think it's ok to set the opacity in background color because the dimming view either have a color or blur effect view. And the blur effect has its own opacity.

IBAnimatableApp/PresentingViewController.swift
JakeLin, line 96: If it's not nil, shall we use the specified value for dismissalAnimationType?

tbaranes, line 96: Aaaah, sorry I misunderstood, you are right, we have to do the setup fix only if it's nil

JakeLin, line 96: @tbaranes No worries, I should make it more specific, I was typing on my lovely 📱

tbaranes:
Here some news of the shadow:
-  Updating the default values to have a shadow by default if we set a color
- Updating the shadow implementation in order to make it work (and avoid useless if)

I have one last issue: the color from the slider seems to be invalid, if you hardcode the color, the shadow will appear as expected, but not when using the custom one. Well, almost there 👯

tbaranes:
I'm a bit disappointing about this hack, in these lasts commits, the shadow has been fixed. I also change the default value of the presentation in order to have a default shadow just by setting a color.

The main bug was due to the picker color, I don't know why, but the layer didn't process the hex color from the picker. So I replaced each hex by an `UIColor`, that make the shadow work as expected but the sliders vs the result can be weird (their not enough `UIColor`). Ideally, we should make a new sliders and respect the `UIColor`, maybe later?

Anyway, you can review it one more time, I think that time, it's ready to go :)

IBAnimatable/AnimatableModalViewController.swift
```diff
@@ -104,7 +104,7 @@ public class AnimatableModalViewController: UIViewController, PresentationDesign
     }
   }
 
-  @IBInspectable public var shadowOffset: CGPoint = CGPoint(x: CGFloat.NaN, y: CGFloat.NaN) {
+  @IBInspectable public var shadowOffset: CGPoint = CGPoint.zero {
```
JakeLin: We can remove `CGPoint` and just use `.zero`

JakeLin:
👍  well done @tbaranes

PR #283, opened 2016-08-16
Feature/transition animatable

lastMove:
TransitionAnimatable Updated to use enums. 
for instance : 
`var transitionAnimationType: String? { get set }`
becomes
`var transitionAnimationType: TransitionAnimationType? { get set }`

I have one interrogation : 
Should it be an optional or not (adding a none clause) ? 
I don't know enough the TransitionAnimatable system to decide about this.

PS: Seems that I didn't made the last Xcode beta update. 
There is maybe some errors, I will update it tomorrow

### DIFF
--- a/IBAnimatable/AnimatableNavigationController.swift
+++ b/IBAnimatable/AnimatableNavigationController.swift
@@ -8,7 +8,14 @@ import UIKit
 public class AnimatableNavigationController: UINavigationController, TransitionAnimatable {
   
   // MARK: - TransitionAnimatable
-  @IBInspectable public var transitionAnimationType: String? {
+  @IBInspectable  var _transitionAnimationType: String? {
+    didSet {
+      if let _transitionAnimationType = _transitionAnimationType {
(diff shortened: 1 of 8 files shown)


tbaranes:
cool 👍 

Make `transitionAnimationType` optional is be the easiest way (I think). If you want to add a `none` case to like the others enums, you have to change:
-  the `guard` condition in the `prepareForSegue` of `AnimatableViewController`
- the `configureNavigationControllerDelegate` of `AnimatableNavigationController`, 

I think that's all 👯

JakeLin:
COOL, 👍 I have a question for you guys @lastMove @tbaranes , shall we make all enums conform to `IBEnum` have an option `.none` except for system built-in one like 'UIBlurEffect' ? I remember having `.none` can help us apply default value in the protocol.

tbaranes:
That would be nice, moreover it's better to uniformise them, even for `UIBlurEffectStyle`, we can add a none case but just ignore it when we received that value.

JakeLin:
I don't think we can add .none to UIBlurEffectStyle , it will be great if we can

Issue #286, opened 2016-08-19
Xcode 8 Beta 6 IterateEnum


lastMove:
Hi, 
Within the project there is a function that allows us to iterate over all values of an Enum (by returning an iterator). 
There is some change with UnsafePointer in the last beta release of Xcode 8 ; so the function don't compile anymore. 

``` Swift

public func iterateEnum<T: Hashable>(from: T.Type) -> AnyIterator<T>? {
var x = 0
  return AnyIterator {
    let next = withUnsafePointer(&x) {
      UnsafePointer<T>($0).pointee
    }
    defer {
      x += 1
    }
    return next.hashValue == x ? next : nil
  }
```

Does someone have an idea of how to migrate this function ? 
Except this function (I commented it) the code is migrated here : #287

dylan:
Give my branch a shot, it seems to work well in playgrounds.

lastMove:
Oh thanks a lot ! I will try it out :+1:

lastMove:
It works well !

JakeLin:
Great, well done guys 👍👍👍

dylan:
![salute](http://gif.dyli.sh/bruceyes.gif)

PR #287, opened 2016-08-19
[WIP] Migrate To Beta 6

### LINKED ISSUE #285: [Swift 3] Open VS Public
The highly discussed [Proposal 0117](https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md) is now implemented : 

> Classes declared as public can no longer be subclassed outside of their defining module, and methods declared as public can no longer be overridden outside of their defining module. To allow a class to be externally subclassed or a method to be externally overridden, declare them as open, which is a new access level beyond public.  

We now have to choose
We have to explicitly decide which class are subclassable and which methods / properties are overridable. 
I am not a big fan of this decision to make classes non-subclassable by default, since we sometimes need to inherit from a class that was not designed for subclassing (for instance `UITabbar` etc). 
So I will not be very objective about this ;)

What do you think ?

lastMove:
Will be ready when #286 will be solved
Close #285

### DIFF
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -76,25 +76,25 @@ public extension Animatable where Self: UIView {
   public func animate(_ animation: AnimationType? = nil, completion: AnimatableCompletion? = nil) {
     switch animation ?? animationType {
     case let .slide(way, direction):
-      slide(way: way ?? .in, direction: direction ?? .right, completion: completion)
+      slide(way, direction: direction, completion: completion)
     case let .squeeze(way, direction):
-      squeeze(way: way ?? .in, direction: direction ?? .right, completion: completion)
+      squeeze(way, direction: direction, completion: completion)
     case let .squeezeFade(way, direction):
-      squeezeFade(way: way ?? .in, direction: direction ?? .right, completion: completion)
+      squeezeFade(way, direction: direction, completion: completion)
     case let .slideFade(way, direction):
-      slideFade(way: way  ?? .in, direction: direction ?? .right, completion: completion)
+      slideFade(way, direction: direction, completion: completion)
     case let .fade(way):
-      fade(way: way ?? .in, completion: completion)
+      fade(way, completion: completion)
     case let .zoom(way):
-      zoom(way: way ?? .in, completion: completion)
+      zoom(way, completion: completion)
     case let .zoomInvert(way):
-      zoom(way: way ?? .in, invert: true, completion: completion)
+      zoom(way, invert: true, completion: completion)
     case .shake:
       shake(completion)
     case .pop:
       pop(completion)
     case let .flip( axis):
-      flip(axis: axis ?? .x, completion: completion)
+      flip(axis, completion: completion)
     case .morph:
       morph(completion)
     case .flash:
@@ -104,11 +104,11 @@ public extension Animatable where Self: UIView {
     case .swing:
       swing(completion)
     case let .rotate(direction):
-      rotate(direction: direction ?? .cw, completion: completion)
+      rotate(direction, completion: completion)
     case let .moveBy(x, y):
-      moveBy(x: x ?? 0, y: y ?? 0, completion: completion)
+      moveBy(x, y: y, completion: completion)
     case let .moveTo(x, y):
-      moveTo(x: x ?? 0, y: y ?? 0, completion: completion)
+      moveTo(x, y: y, completion: completion)
     case .none:
       break
     }
@@ -125,7 +125,7 @@ public extension Animatable where Self: UIView {
   }
   
   // MARK: - Animation methods
-  private func computeValues(way: AnimationType.Way, direction: AnimationType.Direction, shouldScale: Bool) -> AnimationValues {
+  fileprivate func computeValues(_ way: AnimationType.Way, direction: AnimationType.Direction, shouldScale: Bool) -> AnimationValues {
     let yDistance = screenSize.height * force
     let xDistance = screenSize.width * force
     let scale = 3 * force
@@ -160,9 +160,9 @@ public extension Animatable where Self: UIView {
     
   }
   
-  public func slide(way: AnimationType.Way, direction: AnimationType.Direction, completion: AnimatableCompletion? = nil) {
+  public func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, completion: AnimatableCompletion? = nil) {
 
-    let values = computeValues(way: way, direction: direction, shouldScale: false)
+    let values = computeValues(way, direction: direction, shouldScale: false)
     switch way {
     case .in:
       animateIn(values.x, values.y, values.scaleX, values.scaleY, 1, completion)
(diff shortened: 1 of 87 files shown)


IBAnimatable/Animatable.swift
```diff
@@ -76,25 +76,25 @@ public extension Animatable where Self: UIView {
   public func animate(_ animation: AnimationType? = nil, completion: AnimatableCompletion? = nil) {
     switch animation ?? animationType {
     case let .slide(way, direction):
-      slide(way: way ?? .in, direction: direction ?? .right, completion: completion)
+      slide(way, direction: direction, completion: completion)
```
JakeLin: How can we apply default value?

IBAnimatable/Animatable.swift
```diff
@@ -198,7 +198,7 @@ public extension Animatable where Self: UIView {
 
   }
   
-  public func moveTo(x: Double, y: Double, completion: AnimatableCompletion? = nil) {
+  public func moveTo(_ x: Double, y: Double, completion: AnimatableCompletion? = nil) {
```
JakeLin: Shall we keep the "x:" part like `public func moveTo(x: Double, y: Double, completion: AnimatableCompletion? = nil)` or we can use `public func moveToX(_ x: Double, y: Double, completion: AnimatableCompletion? = nil)` ?

I think the first one is more Swift3ify.

IBAnimatable/AnimatableBarButtonItem.swift
```diff
@@ -5,17 +5,17 @@
 
 import UIKit
 
-@IBDesignable public class AnimatableBarButtonItem: UIBarButtonItem, BarButtonItemDesignable, Animatable {
+@IBDesignable open class AnimatableBarButtonItem: UIBarButtonItem, BarButtonItemDesignable, Animatable {
```
JakeLin, line 8: 👍 , I like `open` here.

IBAnimatable/ContainerTransition.swift
```diff
@@ -49,13 +49,13 @@ public class ContainerTransition: NSObject {
       return
     }
     
-    viewControllers = [UITransitionContextFromViewControllerKey: unwrappedFromViewController,
-                       UITransitionContextToViewControllerKey: toViewController]
-    views = [UITransitionContextFromViewKey: unwrappedFromViewController.view,
-             UITransitionContextToViewKey: toViewController.view]
+    viewControllers = [UITransitionContextViewControllerKey.from.rawValue: unwrappedFromViewController,
```
JakeLin: If we let `UITransitionContextViewControllerKey` conform to `Equatable` and implement `func ==(lhs: Self, rhs: Self) -> Bool`, can we use it as the key for the dictionary. Then we don't need to use `.rawValue`?

IBAnimatable/SlideAnimator.swift
```diff
@@ -5,18 +5,18 @@
 
 import UIKit
 
-public class SlideAnimator: NSObject, AnimatedTransitioning {
+open class SlideAnimator: NSObject, AnimatedTransitioning {
```
JakeLin: Do we have any requirements to subclass transition animator like `SlideAnimator`, `SystemCameraIrisAnimator` etc. ? I think we can keep them `public`. @tbaranes Do you have any idea about this?

tbaranes: Overriding an animator would be a bad idea. If we need to customise an animator, we either have to add more parameters, otherwise that means it should be a new one. So I'm joining to say that it should be `public` instead of `open`

JakeLin:
@lastMove very good first step. Please have a look at our comments. I think once we fix the `IterateEnum` issue, it is ready to go. But I know it is very difficult though 😢

IBAnimatable/Animatable.swift
lastMove: Actually those default value were useless. Beta 6 now put a warning when `??` operator is used with non-optional values. 
We decided some times ago to make `AnimationType` enum parameters non-optional. 
So we don't have anymore defaultValue when called from code, Just the documentation to indicate the default value. DefaultValues are still applied when called from IB (in `init(string: String?)`

IBAnimatable/Animatable.swift
lastMove: Yes I think too. I will make this change.

IBAnimatable/AnimatableBarButtonItem.swift
```diff
@@ -28,22 +28,22 @@ import UIKit
...
     }
   }
-  @IBInspectable public var autoRun: Bool = true
-  @IBInspectable public var duration: Double = Double.nan
-  @IBInspectable public var delay: Double = Double.nan
-  @IBInspectable public var damping: CGFloat = CGFloat.nan
-  @IBInspectable public var velocity: CGFloat = CGFloat.nan
-  @IBInspectable public var force: CGFloat = CGFloat.nan
-  @IBInspectable public var repeatCount: Float = Float.nan
+  @IBInspectable open var autoRun: Bool = true
```
tbaranes, line 37: Should it be `open` instead of `public`? I can't see why we would have to override them
(that question can be applied for most of `IBInspectable` variables)

IBAnimatable/AnimatableCollectionViewCell.swift
```diff
@@ -58,63 +58,63 @@ import UIKit
...
+  @IBInspectable open var removeSeparatorMargins: Bool = false
   
   // MARK: - GradientDesignable
-  @IBInspectable public var startColor: UIColor?
-  @IBInspectable public var endColor: UIColor?
-  public var predefinedGradient: GradientType?
+  @IBInspectable open var startColor: UIColor?
+  @IBInspectable open var endColor: UIColor?
+  open var predefinedGradient: GradientType?
   @IBInspectable var _predefinedGradient: String? {
```
tbaranes, line 67: Is it ok to not specify an access control?

IBAnimatable/ContainerTransition.swift
```diff
@@ -5,22 +5,22 @@
 
 import UIKit
 
-public class ContainerTransition: NSObject {
+open class ContainerTransition: NSObject {
```
tbaranes: I'm not sure that one should be `open` right now, as well as its properties and methods.

IBAnimatable/AnimatableCollectionViewCell.swift
JakeLin, line 67: It means `internal`, right? Is it still true for new Swift 3?

lastMove, line 67: Yes, The default access control is still `internal`

IBAnimatable/ContainerTransition.swift
lastMove: Good call ! Done. (`UITransitionContextViewControllerKey` & `UITransitionContextViewKey` were already `Equatable`)

IBAnimatable/SlideAnimator.swift
lastMove: Done :)

lastMove:
Since `IterateEnum` is fixed it's almost ready. 
Actually the Xcode migration tool made all `public` statement => `open` 
Personnaly I would prefer to let the biggest part of the Framework `open`. And use public only when the classes/methods/properties are dangerous to subclass or override. (Sometimes I found myself subclassing class that was not intended for. To fix some behaviours)

JakeLin:
@lastMove cool, great work. Is this PR ready to go now?

lastMove:
Yep

PR #298, opened 2016-09-02
Bring release 2.7 to Swift3

JakeLin:
This PR brings release 2.7 to Swift3. 

After that, all new features should be added to `swift3` branch unless we merge it to master. 

All activity indicators are working fine.

There are some issues when I tested the app, I would like to merge this big changes into `swift3` then we can fix them one by one.
- [x] BorderSide in the demo doesn't work
- [x] Blur effect in playground app doesn't work - It doesn't change when we change "Blur"
- [x] Predefined gradients in playground app doesn't work 
- [x] Gradients in playground app doesn't work 
- [x] Mask option "None" in playground app doesn't work 
- [ ] Transition in playground app crashes
- [ ] Container Transition in playground app doesn't work 

Related to #221


tbaranes:
That's too big to be reviewed, let's move forward step by step as you said.

JakeLin:
@tbaranes I did review it by myself and fixed some small issues. Let's merge it now.

lastMove:
@JakeLin I didn't had problems with borderSides demo. 
Which part didn't work ?

JakeLin:
@lastMove, the problem is in the demo app not IBAnimatable Playground, the border side in the playground works perfectly fine. Please leave this one to me. I think it may be an issue of the run time attribute of border side. Need to have an underscore like _borderSide , shouldn't be hard.

lastMove:
Ok. I didn't reach to fix the Blur Effects problem (Changing the vibrancy style don't have any effect, on the view).

PR #300, opened 2016-09-03
Fix demo app storyboard to support new enums.

JakeLin:
Interface Builder in Xcode 8 has improved a lot, I have fixed all layout warnings and supported all new enums. `borderSides` works fine now.


lastMove:
👍

JakeLin:
thanks @lastMove

PR #301, opened 2016-09-04
Fix all storyboards

JakeLin:
In this PR, we fixed all storyboards for Xcode 8 beta 6 and change all purple colors to `#BA77FF` using `sRGB` settings.

Related to #221 and #298 

@lastMove can you have a look when you have time?

### DIFF
--- a/IBAnimatable/ActivityIndicatorAnimating.swift
+++ b/IBAnimatable/ActivityIndicatorAnimating.swift
@@ -5,6 +5,14 @@
 
 import UIKit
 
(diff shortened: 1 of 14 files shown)


lastMove:
Nice work, lot of changes. 
Some colors were changed to white / transport. 
Is it intended  ?

JakeLin:
@lastMove, yes, I removed the color of the tableviewcell 's content view. Because now we have animatabletableviewcell , we don't need to set it twice.

lastMove:
Ok, It all goods.

PR #302, opened 2016-09-05
Fix blur effect view

JakeLin:
In this PR, we fix the blur effect view. Now we can re-set `blurEffectStyle` and it will apply the new blue effect to the view 😉. We can also support `vibrancyEffectStyle` but we can only set it once because  once we use `UIVisualEffectView`, Apple uses a private class like `_UIVisualEffectContentView` to contain subviews which will change the view hierarchy. After that, we are not able to restore the original view hierarchy. Also, the Auto Layout constraints will be broken when the system change the view hierarchy. We need to manually re-set up the constraints or use Autoresizing masks to layout the subviews.

@lastMove Can you have a look when you have time. 

Related to #221 and #298

### DIFF
--- a/IBAnimatableApp/ParamType.swift
+++ b/IBAnimatableApp/ParamType.swift
@@ -3,7 +3,7 @@
 //  IBAnimatableApp
 //
 //  Created by jason akakpo on 27/07/16.
-//  Copyright © 2016 Jake Lin. All rights reserved.
+//  Copyright © 2016 IBAnimatable. All rights reserved.
 //
 
 import Foundation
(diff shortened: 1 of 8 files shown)


JakeLin:
Here is the new GIF, it will look much better on a simulator 😂

blue effect view

lastMove:
@JakeLin Annoying bug, Maybe if we set up views programmatically we can re-create them each changes. I will try something (to fix it within the playground part of the app) after this PR

IBAnimatableApp/ParamType.swift
```diff
@@ -61,12 +61,12 @@ enum ParamType {
     }
   }
   
-  func titleAt(_ index: Int) -> String {
+  func title(at index: Int) -> String {
```
lastMove, line 64: 👍

lastMove:
Reviewed it all good. 
I will merge it.

JakeLin:
Thanks, @lastMove I will work on transition animations

PR #309, opened 2016-09-11
Fixes for animators when navigation bar non translucent. Issue #302

mikhailmulyar:
Fixes for Card, Fade, NatGeo, Slide and all System animations

### DIFF
--- a/IBAnimatable/AnimatedTransitioning.swift
+++ b/IBAnimatable/AnimatedTransitioning.swift
@@ -32,6 +32,11 @@ public extension AnimatedTransitioning {
       return
     }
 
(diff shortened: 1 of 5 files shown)


JakeLin:
@mikhailmulyar thanks for your PR. I have a question, do we need to set the `fromView.frame` as well?

mikhailmulyar:
No, the problem is that final frame is not correct, we do not care about final frame for fromView since it is removed

JakeLin:
@mikhailmulyar I think the problem is the (to/from) frames or snapshots are not right during the transition. I can see they dump and move back to the right position.

PR #310, opened 2016-09-11
Refator configure*** methods

JakeLin:
In this PR, we refactor all `config***` ethods to `configure***` according to **Clarity is more important than brevity** in  [API Design Guidelines](https://swift.org/documentation/api-design-guidelines/)

related to #221


tbaranes:
Good idea 👍

lastMove:
Nice.

JakeLin:
Thanks guys 👯

PR #311, opened 2016-09-11
Refactor presentation designable

lastMove:
Key changes : 
PresentationModalPosition`,`PresentationModalSize`,`PresentationAnimationType`now conforms to IBEnum.
`ModalKeyboardTranslation`now conforms to IBEnum.
replace some`static fromString(String?)`by`init?(String)`

Refactor PresentationDesignable && AnimatableModalViewController to use enums instead of strings.

modalWidth and modalHeight were removed from `PresentationDesignable` => instead we use directly `ModalSize`(`modalSize: ModalSize`)
There is a Lot of changes, to get it work. @tbaraness Since you wrote the orignal code, can you have a look please ? 

TO BE FIXED : 
Size of the modal seems to not work from the demo (both in this branch and the swift3 one).

related to #221


IBAnimatable/AnimatableModalViewController.swift
```diff
@@ -7,73 +7,98 @@ import UIKit
...
+  
+  public var modalPosition: PresentationModalPosition = .center {
     didSet {
-      presenter?.presentationConfiguration?.modalSize = (PresentationModalSize.fromString(size: modalWidth), PresentationModalSize.fromString(size: modalHeight))
+      presenter?.presentationConfiguration?.modalPosition = modalPosition
     }
   }
-
-  @IBInspectable public var modalHeight: String = "Half" {
+  @IBInspectable var _modalWidth: String? {
```
tbaranes, line 58: Should we keep a default value here?

IBAnimatable/AnimatableModalViewController.swift
```diff
@@ -91,62 +116,66 @@ open class AnimatableModalViewController: UIViewController, PresentationDesignab
...
+  
   public required init?(coder aDecoder: NSCoder) {
     super.init(coder: aDecoder)
     setupPresenter()
   }
-
+  
   // MARK: Life cycle
-
+
```
tbaranes: Did we disable swiftlint rule about the trailing spaces? There's a lot new ones here 😆

tbaranes:
Looks good 👍

IBAnimatable/AnimatableModalViewController.swift
JakeLin: I think we set it in line 60. Do we need to do it here as well?

IBAnimatable/PresentationAnimationType.swift
```diff
@@ -27,55 +27,27 @@ public enum PresentationAnimationType {
...
+    case "crossdissolve":
+      self = .crossDissolve
+    case "flip":
+      self = .flip
+    case "zoom":
+      self = .zoom
+    case "dropdown":
+      self = .dropDown
+    case "cover":
+      self = .cover(fromDirection: TransitionAnimationType.Direction.fromString(forParams: params)!)
```
JakeLin: Will the app crash because we use force unwrap here?

IBAnimatable/PresentationAnimationType.swift
```diff
@@ -27,55 +27,27 @@ public enum PresentationAnimationType {
...
-  static func params(forTransitionType transitionType: String) -> [String] {
-    let range = transitionType.range(of: "(")
-    let transitionType = transitionType.replacingOccurrences(of: " ", with: "").lowercased()
-      .substring(from: range?.lowerBound ?? transitionType.endIndex)
-      .replacingOccurrences(of: "(", with: "")
-      .replacingOccurrences(of: ")", with: "")
-      .capitalized
-    return transitionType.components(separatedBy: ",")
-  }
-
```
JakeLin, line 80: Good to remove a lot of code, 👍

IBAnimatable/PresentationAnimationType.swift
```diff
@@ -8,17 +8,17 @@ import UIKit
 /**
  Predefined Transition Animation Type
  */
-public enum PresentationAnimationType {
+public enum PresentationAnimationType: IBEnum {
   case flip
   case crossDissolve
   case cover(fromDirection: TransitionAnimationType.Direction)
```
JakeLin: We may be able to use `case cover(from: TransitionAnimationType.Direction)` based on **Choose parameter names to serve documentation** from **API Design Guidelines**, because the type can be found in `TransitionAnimationType.Direction`

JakeLin:
👍  🚢 🇮🇹

JakeLin:
@lastMove I have added the issue you mentioned to #312

IBAnimatable/PresentationAnimationType.swift
lastMove: Yes it better, I will do that .

IBAnimatable/AnimatableModalViewController.swift
lastMove: I checked, Swxftlint don't say anything here. Actually, it's because I made Xcode Auto re-indent ( Editor -> Structure -> Re-indent)

IBAnimatable/PresentationAnimationType.swift
lastMove: Oh ! my bad, good catch !

IBAnimatable/AnimatableModalViewController.swift
lastMove: Nope.

lastMove: I think it better to have default enums Value rather than String ones.

JakeLin:
Thanks @lastMove I merged it now.

PR #314, opened 2016-09-13
Refactor some names with present / dismiss

JakeLin:
In this tiny PR, we refactor some names with present / dismiss.

Related to #221


tbaranes:
👌

PR #316, opened 2016-09-15
disabled swiftlint variable_name_min_length for x, y, z, r, g, b

appoulsen:
disabling errors generated by swiftlint

### DIFF
--- a/IBAnimatable/ActivityIndicatorAnimationBallRotateChase.swift
+++ b/IBAnimatable/ActivityIndicatorAnimationBallRotateChase.swift
@@ -30,9 +30,8 @@ public class ActivityIndicatorAnimationBallRotateChase: ActivityIndicatorAnimati
 
 
 // MARK: - Setup
-
+// swiftlint:disable variable_name_min_length
 private extension ActivityIndicatorAnimationBallRotateChase {
-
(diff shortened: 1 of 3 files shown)


tbaranes (approved):
Won't it better

tbaranes (approved):
cool 👍

IBAnimatable/ActivityIndicatorAnimationBallRotateChase.swift
```diff
@@ -30,9 +30,8 @@ public class ActivityIndicatorAnimationBallRotateChase: ActivityIndicatorAnimati
 
 
 // MARK: - Setup
-
+// swiftlint:disable variable_name_min_length
```
tbaranes, line 33: Won't it be better to use `swifltint:disable:next variable_name_min_length` instead of `{disable,enable}`?

appoulsen: It is basically the same. Make a pull request if you think it is better. 
Cause right now it is the only thing preventing me from build on Xcode 8 with swiftlint

JakeLin:
@appoulsen thanks for your PR, I merge it now. Hope can unblock you for now. We are still busy on finalizing the release 3, will release once it is done.

PR #317, opened 2016-09-16
Upgrade to Swift 2.3

JakeLin:
Upgrade to Swift 2.3


JakeLin:
@gkye @loganblevins I have fixed all force unwrap code, can you have a look when you have time? Thanks.

gkye:
Looks good, perhaps we should add a link in the master README to let people know we also support `2.3`?

JakeLin:
@gkye you are right, added a card to the project, https://github.com/IBAnimatable/IBAnimatable/projects/1#card-124775 you can change it as well if you like 😉

PR #330, opened 2016-09-19
Potential BUGFIX InteractiveGestureType - ScreenEdgePan 

lastMove:
since `extractNameAndParams` lowercase name and params, the case should be lowerCased to match the case.


tbaranes (approved):

JakeLin (approved):
It is in another PR already.

PR #331, opened 2016-09-19
Refactor/remove useless code

lastMove:
Since `ColorType` has a `color: UIColor` property, we can just use it.
To be merged _after_ #330

### DIFF
--- a/IBAnimatable/InteractiveGestureType.swift
+++ b/IBAnimatable/InteractiveGestureType.swift
@@ -57,7 +57,7 @@ extension InteractiveGestureType: IBEnum {
     case "pan":
       let direction = GestureDirection(raw: params[safe: 0], defaultValue: .left)
       self = .pan(from: direction)
-    case "screenedgePan":
+    case "screenedgepan":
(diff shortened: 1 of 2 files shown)


tbaranes (approved):

IBAnimatable/InteractiveGestureType.swift
```diff
@@ -57,7 +57,7 @@ extension InteractiveGestureType: IBEnum {
     case "pan":
       let direction = GestureDirection(raw: params[safe: 0], defaultValue: .left)
       self = .pan(from: direction)
-    case "screenedgePan":
+    case "screenedgepan":
```
JakeLin, line 60: Good find.

JakeLin (approved):
Well done 👍👍👍

tbaranes:
I think you can merge even with travis failing, since travis-ci/push passed, it should be ok! Seems like it has some issues with the new xcodeproj renaming

JakeLin:
@tbaranes it is some weird issue with Travis, I can see buddybuild passes https://dashboard.buddybuild.com/apps/56abf6d42c882e010057b182/build/57df97086c21af010069aae3

PR #334, opened 2016-09-20
Fix `AnimationType`

JakeLin:
Fix `AnimationType` since `from` doesn't make sense for `.out` like `slide(.out, from: .left)`. It should be `.slide(.out, direction: .left)`.

It is a small PR, I made a mistake for changing `AnimationType`. Sometimes, `from` is wrong, it is better to use `direction`.


tbaranes (changes requested):
Shouldn't you change the demo app and playground as well? 

Otherwise 👍

JakeLin:
Thanks @tbaranes I have made a change to the playground 01e8ad0c27da66d8efc259d46a088b8788672c95  and no need to change the demo app.

tbaranes (approved):

PR #336, opened 2016-09-20
Fix travis

tbaranes:
I think I understand why travis is failing: it's trying to build the demo app but its scheme is not shared.
We'll tommorrow if that was the right fix or not 😬


JakeLin (approved):
Let's have a look

PR #348, opened 2016-10-01
Fix some inspectable properties in playground demo storyboard.

JakeLin:
Small PR to fix some inspectable properties in playground demo storyboard. For example, replace `predefinedColor` to `_ predefinedColor`.


tbaranes (approved):

JakeLin:
Thanks @tbaranes

PR #357, opened 2016-11-18
Get 3.1 release ready

JakeLin:
Thanks to @mmadjer , now we can configure corner radius sides in all `IBAnimatable` UI elements. We can ship it in 3.1 🎉

@tbaranes can you please have a look at this PR when you have time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ Add `pod 'IBAnimatable'` to your Podfile.
 
 Add `github "IBanimatable/IBAnimatable"` to your Cartfile.
 
-Please Notice, there is [a limitation of a built framework for `@IBDesignable` and `@IBInspectable`](https://github.com/Carthage/Carthage/issues/335), that will impact on `IBAnimatable` when you use Carthage.
+Please Notice, there is [a limitation of a built framework for `@IBDesignable` and `@IBInspectable`](https://github.com/Carthage/Carthage/issues/335), that will impact on `IBAnimatable` when you use Carthage. There is a workaround to use Carthage or Swift package manager with `IBAnimatable`, please have a look at [Carthage – no Animatable UI Classes appearing in Storyboard](https://github.com/IBAnimatable/IBAnimatable/issues/354)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -168,7 +169,7 @@ Add `pod 'IBAnimatable'` to your Podfile.
 
 Add `github "IBanimatable/IBAnimatable"` to your Cartfile.
 
-Please Notice, there is [a limitation of a built framework for `@IBDesignable` and `@IBInspectable`](https://github.com/Carthage/Carthage/issues/335), that will impact on `IBAnimatable` when you use Carthage.
+Please Notice, there is [a limitation of a built framework for `@IBDesignable` and `@IBInspectable`](https://github.com/Carthage/Carthage/issues/335), that will impact on `IBAnimatable` when you use Carthage. There is a workaround to use Carthage or Swift package manager with `IBAnimatable`, please have a look at [Carthage – no Animatable UI Classes appearing in Storyboard](https://github.com/IBAnimatable/IBAnimatable/issues/354)
```
tbaranes, line 172: Not sure we can talk about a workaround since the only solution is to use CocoaPods, so...

tbaranes (approved):
Just a feedback about the carthage workaround, not sure how we can improve the sentence, otherwise, it's all good! Let's make the 3.1 real 🎉

README.md
JakeLin: @tbaranes thanks for picking this up. @terenzeyuen mentioned they can use "a workaround of Carthage + Workspace". That's why I put it in README.md . Hopefully, it can help someone use `IBAnimatable` when they have to use Carthage.

tbaranes: Oh I miss that then, I'm keeping cocoapods just for IBAnimatable, doesn't seem to be working even with workspace. I will give it a second try

JakeLin: I am using CocoaPods for `IBAnimatable` too because of this issue. Also, I found out CocoaPods is easier to use than Carthage in most of the cases.

PR #359, opened 2016-11-18
Refactoring singleton

### LINKED ISSUE #358: Update Singleton syntax
According to this article [THE RIGHT WAY TO WRITE A SINGLETON](http://krakendev.io/blog/the-right-way-to-write-a-singleton), we can update the Singleton syntax like

```
class TheOneAndOnlyKraken {
    static let sharedInstance = TheOneAndOnlyKraken()
    private init() {} //This prevents others from using the default '()' initializer for this class.
}
```

- [x] `PresentationPresenterManager`
- [x] `TransitionPresenterManager`

tbaranes:
Close #358


JakeLin (approved):
You almost did it at the same time the issue created 😱 

👍👍👍

PR #361, opened 2016-11-20
Bump swiftlint 0.13.0

tbaranes:
Side though, should we setup [Danger](https://danger.systems) in order to be aware that we let some warnings before merging a PR?


JakeLin (approved):
Good one 👍

JakeLin:
@tbaranes I created an issue #362 for that.

PR #369, opened 2016-11-30
Bump SwiftLint 0.13.1

tbaranes:
SwiftLint 0.13.1 is out! I used that occasion to review a bit the rules we are using. Since, we are getting more and more to contribute in this project (that's awesome!), it's even more important now to have a strict coding style.

That's why I added a few `opt_in_rules` that have meanings for IBAnimatable:
 - closure_spacing
 - conditional_returns_on_newline
 - empty_count
 - explicit_init
 - overridden_super_call
 - redundant_nil_coalescing
 - nimble_operator (that one will be useful only if we write tests one day 😬)

I'd like also use do a reminder of the disabled rules that I think they can /  should be enabled:
  - **trailing_whitespace**: that one has been disabled because we have a lot of warnings. I will fix that soon
  - **line_length** (88 warnings): We have quiet long lines in IBAnimatable, I think it will be worth to reduce them
  - **cyclomatic_complexity** (3 warnings, 2 errors): That one is hard to fix without changing IBAnimatable core. I don't really mind since it's our core is build like that, but it's worth a reminder: _Do we really want that?_
  - **function_body_length** (7 warnings): mostly due to the huge switch. Same speech as `cyclomatic_complexity` rule, the project is build on enums so we have huge switch. Just a reminder.
  - **type_name** (14 warnings): mostly due to `AnimatableActivityIndicator*`. I already though that our names are long, very long (we basically have a long prefix (`Animatable*`) and suffix (`Designable`), but right now, I have no idea how to reduce them.

That last part can be a full issue, so I will open one if any of you wants to talk more about this. For now, that PR is ready to go 😬

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -7,3 +7,12 @@ disabled_rules:
   - function_body_length
   - cyclomatic_complexity
   - type_name
+
+opt_in_rules:
+ - closure_spacing
+ - conditional_returns_on_newline
+ - empty_count
+ - explicit_init
+ - overridden_super_call
+ - redundant_nil_coalescing
+ - nimble_operator
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,9 +1,11 @@
+line_length: 150
+file_length: 500
+
 disabled_rules:
   - trailing_whitespace
   - todo
   - variable_name
   - line_length
-  - file_length
   - function_body_length
   - cyclomatic_complexity
   - type_name
--- a/IBAnimatable/ColorType.swift
+++ b/IBAnimatable/ColorType.swift
@@ -31,49 +31,47 @@ public enum ColorType: String, IBEnum {
 
 extension ColorType {
   public var color: UIColor {
-    get {
-      switch self {
-      case .flatBelizeHole:
-        return #colorLiteral(red: 0.16078431372549, green: 0.501960784313725, blue: 0.725490196078431, alpha: 1.0)
-      case .flatAlizarin:
-        return #colorLiteral(red: 0.905882352941176, green: 0.298039215686275, blue: 0.235294117647059, alpha: 1.0)
-      case .flatEmerland:
-        return #colorLiteral(red: 0.180392156862745, green: 0.8, blue: 0.443137254901961, alpha: 1.0)
-      case .flatMidnightBlue:
-        return #colorLiteral(red: 0.172549019607843, green: 0.243137254901961, blue: 0.313725490196078, alpha: 1.0)
-      case .flatPumpkin:
-        return #colorLiteral(red: 0.827450980392157, green: 0.329411764705882, blue: 0.0, alpha: 1.0)
-      case .flatAmethyst:
-        return #colorLiteral(red: 0.607843137254902, green: 0.349019607843137, blue: 0.713725490196078, alpha: 1.0)
-      case .flatOrange:
-        return #colorLiteral(red: 0.952941176470588, green: 0.611764705882353, blue: 0.0705882352941176, alpha: 1.0)
-      case .flatNephritis:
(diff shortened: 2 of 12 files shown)


Panajev:
Only two points stick out for me: line length, depending on the value and type name length.

For the former, today's screens are far wider than they used to be, but a reasonable cap could be put in place (although line wrapping gets the job done) and clarity should still be king... which leads me to the length of type names. We do have fuzzy matches autocomplete, do we really have an issue there?

JakeLin:
@tbaranes thanks for maintaining the project and making it better and better 👍👍👍. 

It is great we can pass a big list of `opt_in_rules`. For the ones you mentioned. I think we should try to fix **trailing_whitespace** and ** line_length** as the starting point.  How about we create two tasks/issues and make them as `difficulty: easy`? Hopefully, we can fix them easily, also attract some contributors to start.

I agree with @Panajev "clarity should still be king". In [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/), Apple also suggests "Clarity is more important than brevity". We should keep the type names like `AnimatableActivityIndicator*` to keep the code more readable. 

@Panajev I know your concern about `line length`. If they are caused by the good readable long type name, I think we should keep them. In some case, they are caused by we put all parameters in one line, I think we should break them in multiple lines which can help readability.

Actually, we are all in the same boat. We want to improve the quality of the code. We should follow the rules benefit us as human beings. Lint rules are the methods help us achieve the goal. Sometimes, if the Lint rules are opposite of our goal, we should ignore them. But a lot of times, lint rules can help us avoid some small code style issues, especially for someone is new to the project. I can use it as a utility for checking small code style issues.

.swiftlint.yml
```diff
@@ -1,9 +1,20 @@
...
+  - type_name
+
+opt_in_rules:
+ - closure_spacing
+ - conditional_returns_on_newline
+ - empty_count
+ - explicit_init
+ - overridden_super_call
+ - redundant_nil_coalescing
+ - nimble_operator
```
JakeLin, line 20: Great to have a bunch of `opt_in_rules` 👍

IBAnimatable/ColorType.swift
```diff
@@ -31,49 +31,47 @@ public enum ColorType: String, IBEnum {
...
-      case .flatPomegranate:
-        return #colorLiteral(red: 0.752941176470588, green: 0.223529411764706, blue: 0.168627450980392, alpha: 1.0)
-      case .flatSilver:
-        return #colorLiteral(red: 0.741176470588235, green: 0.764705882352941, blue: 0.780392156862745, alpha: 1.0)
-      case .flatAsbestos:
-        return #colorLiteral(red: 0.498039215686275, green: 0.549019607843137, blue: 0.552941176470588, alpha: 1.0) 
-      case .flatCarrot:
-        return #colorLiteral(red: 0.901960784313726, green: 0.494117647058824, blue: 0.133333333333333, alpha: 1.0)
-      }
+    switch self {
```
JakeLin, line 34: good pick up.

IBAnimatableApp/MaskViewController.swift
```diff
@@ -16,7 +16,7 @@ class MaskViewController: UIViewController {
     let pointsParam = ParamType.number(min: 3, max: 10, interval: 1, ascending: true, unit: "points") // default 5
     let sidesParam = ParamType.number(min: 3, max: 10, interval: 1, ascending: true, unit: "sides") // default 6
     let angleParam = ParamType.number(min: 60, max: 120, interval: 2, ascending: true, unit: "°") // default 6
-    let waveParam = ParamType.init(fromEnum: MaskType.WaveDirection.self)
+    let waveParam = ParamType(fromEnum: MaskType.WaveDirection.self)
```
JakeLin, line 19: More Swifty 👏

JakeLin (approved):
👍

JakeLin:
@tbaranes great, thanks for adding tow tasks too 👍

Issue #360, opened 2016-11-19
Using Swift Struct as the key for dictionary


JakeLin:
In [WWDC 2016 - Understanding Swift Performance](https://developer.apple.com/videos/play/wwdc2016/416/)

Kyle mentioned we should use Swift Struct instead of `String` as the key in the dictionary. We should use that pattern in `IBAnimatable`.

> Well, in Swift we can represent this configuration space of color, orientation, and tail just using a struct.
> This is a much safer way to represent this configuration space than a String.
> And because structs are first class types in Swift, they can be used as the key in our dictionary.

- [x] `cache` in `PresentationPresenterManager`
- [x] `cache` in `TransitionPresenterManager`

JakeLin:
Thanks to@mmadjer, it is done by #373

PR #373, opened 2016-12-07
Use Swift Struct instead of String as the key in the cache.

mmadjer:
These changes are for #360.


IBAnimatable/PresentationPresenterChacheKey.swift
```diff
@@ -0,0 +1,22 @@
...
+//  PresentationChacheKey.swift
+//  IBAnimatable
+//
+//  Created by Miro on 8/12/16.
+//  Copyright © 2016 IBAnimatable. All rights reserved.
+//
+
+import Foundation
+
+struct PresentationPresenterChacheKey: Hashable {
```
JakeLin: Can we directly use `PresentationAnimationType` enum as the key, and let `PresentationAnimationType: Hashable`  and provide ` static func == (lhs: PresentationPresenterChacheKey, rhs: PresentationPresenterChacheKey)`? If we can, then we don't need the additional struct. enum is also a value type.

JakeLin (changes requested):
Can you check whether we can use the enums `PresentationAnimationType` and `TransitionAnimationType` as the key?

JakeLin:
@mmadjer thanks for your PR. Sorry, I have to disappoint you because I think we may be able to use the enums as the key. Can you double check this approach for me?  We also need to rebase since there is some conflict in `TransitionPresenterManager.swift`.

IBAnimatable/PresentationPresenterChacheKey.swift
mmadjer: I thought about that but could not find a proper way so far. I will have another look.
I'll give a try and apply the Hashable to the two enums. Why I haven't think about that before. ;o/
;o)

mmadjer:
Ok, I applied the Hashable to the two enums and using it as the key.

IBAnimatable/PresentationAnimationType.swift
```diff
@@ -49,5 +49,15 @@ public enum PresentationAnimationType: IBEnum {
       return nil
     }
   }
+
+}
+
+extension PresentationAnimationType: Hashable {
```
JakeLin, line 55: Perfect 👍

JakeLin (approved):
Perfect, looks better.

JakeLin:
@mmadjer Thanks for the super quick update, LGTM, can you do a rebase from the master and merge it when you free it is ready.

JakeLin (approved):
perfect, I always forget about it.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ All notable changes to this project will be documented in this file.
 None
 
 #### Enhancements
-None
+- Use Swift Enum as key in our internal presentation and transition cache.
```
tbaranes: Not sure that we should write a line about this since it's an internal change

tbaranes (approved):

tbaranes:
Not sure that we should add an entry in the CHANGELOG since it's an internal change. What do you think?

JakeLin:
@tbaranes, IMO we need that. It maybe a breaking change for someone uses the Singleton directly (not 💯 % sure since I am replying this on my mobile and didn't check the source code). That record can help the user find the issue.

Issue #375, opened 2016-12-16
Corner radius don't apply into AnimatableButton


nttra83:
Hello,
I set a background image for this button. I wanna set corner radius for button but It don't work.
Could you help me fix this issue?
Thanks,

![screen shot 2016-12-16 at 12 04 02 pm](https://cloud.githubusercontent.com/assets/24289191/21252247/e38ada24-c387-11e6-9ddd-05f8796273a7.png)
![screen shot 2016-12-16 at 12 01 26 pm](https://cloud.githubusercontent.com/assets/24289191/21252248/e56a11ca-c387-11e6-9467-fe4d6e003aab.png)
![screen shot 2016-12-16 at 12 01 21 pm](https://cloud.githubusercontent.com/assets/24289191/21252249/e73be1e0-c387-11e6-847e-b32d4b28ad31.png)

JakeLin:
Hi @nttra83, thanks for reporting that. I believe it is fixed by @mmadjer and has been merged in master, can you use master and try out? If it is fixed, I will make a release asap. Thanks.

JakeLin:
Hi @nttra83 , I have tried and found out you need to set **Clip to Bounds** to **true**
<img width="565" alt="screen shot 2016-12-18 at 10 29 32 pm" src="https://cloud.githubusercontent.com/assets/573856/21293201/54f1c610-c571-11e6-855e-51a9c600bad6.png">

When we use background image for `AnimatableButton`
<img width="487" alt="screen shot 2016-12-18 at 10 30 03 pm" src="https://cloud.githubusercontent.com/assets/573856/21293199/52434038-c571-11e6-8064-d251e02ce3ad.png">


Can you try and close the issue if it works? thanks.

nttra83:
Hello JakeLin,
Thanks for your support :).
It's work perfect :). But I wanna set a shadow color for AnimatableButton. It don't work when I set a shadow color :(. Could you fix it?
Thanks,

![screen shot 2016-12-19 at 10 44 29 am](https://cloud.githubusercontent.com/assets/24289191/21300674/a3c9a1a6-c5d8-11e6-9096-d82e8d322ac5.png)

mmadjer:
It is working for me using your shadow values and code of master.
Change the shadow color, i.e. to black and you should see the shadow rendering. Light Gray and these values are very hard to see on white background.

JakeLin:
[...]

PR #356, opened 2016-11-17
Add Corner sides

mmadjer:
Ability to apply individual corner sides to a view

Issue #267, opened 2016-08-04
Shadows not working with Mask type.


Abdul-Moiz:
I am trying to apply shadow on Mask type Circle seems like it's not working. Removing mask working properly.

JakeLin:
@Abdul-Moiz I introduced this bug when I fixed another bug https://github.com/IBAnimatable/IBAnimatable/pull/180 , I am looking at how can we fix it.

JakeLin:
@Abdul-Moiz  I have tried to fix this issue, please have a look at the commit https://github.com/IBAnimatable/IBAnimatable/commit/3c5bc85a9d6a7497ac58d64d9de2e8a75cbe810d

But the effect is not right, it is not easy to fix it. If you have time, can you have a look and let me know if you find anything? Thanks

![simulator screen shot aug 5 2016 11 45 48 pm](https://cloud.githubusercontent.com/assets/573856/17438717/aaa24f72-5b67-11e6-84ac-9a8a3917407f.png)

If you want to apply a shadow to a Circle, there is a workaround, you can set `cornerRadius = width / 2`.

Abdul-Moiz:
@JakeLin I tried to look into the issue. Somehow the shadow is clipping within the view bounds. 
The approach I can think of to fix it is adding sublayer below main layer. And avoid the masking completely. 

For time being going with the workaround 2.

JakeLin:
@Abdul-Moiz thanks for looking at it. As you said, I think we need to have two layers. And apply the mask to `shadowPath` of the second layer. Would you like to give it a go with a PR when you have time?  Please let me know if you are interested.

Abdul-Moiz:
@JakeLin This is the code that made the shadow render correctly on the view with masking. I didn't tested all the cases but working fine for AnimatableView, Button and image.

```
public func configMaskShadow() {
    commonSetup()
    // if a layer mask is specified, display the shadow to match the mask
    if let mask = layer.mask as? CAShapeLayer {
      // Clear default layer borders
      layer.shadowColor = nil
      layer.shadowRadius = 0
      layer.shadowOpacity = 0

      // Remove any previous layer
      layer.superlayer?.sublayers?.filter { $0.name == "shadowLayer-\(unsafeAddressOf(self))" }
        .forEach { $0.removeFromSuperlayer() }

      // Create new layer with object's memory reference to make this string unique. Otherwise common name will remove all the shadow layers as it's adding in layer's superview.
[...]

PR #377, opened 2016-12-20
WIP: corner radius shadow fix

mmadjer:
Fix corner side radius in conjunction with shadow mask.
See issue #375.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ None
 - Introducing `borderType`, find all the information in the documentation [#389](https://github.com/IBAnimatable/IBAnimatable/pull/380)
 
 #### Bugfixes
-
+- Fix `cornerSide` radius in conjunction with shadow mask.
+[#375](https://github.com/IBAnimatable/IBAnimatable/issues/375)
 - Fix `slide(.out, *)` and `squeeze(.out, *)` translation coordinates [#379](https://github.com/IBAnimatable/IBAnimatable/issues/379)
 
 ### [3.1.2](https://github.com/IBAnimatable/IBAnimatable/releases/tag/3.1.2)
@@ -339,5 +340,3 @@ None
 ### [1.0](https://github.com/IBAnimatable/IBAnimatable/releases/tag/1.0)
 
 - Initial release
-
-
--- a/IBAnimatable/AnimatableButton.swift
+++ b/IBAnimatable/AnimatableButton.swift
@@ -12,6 +12,7 @@ open class AnimatableButton: UIButton, CornerDesignable, FillDesignable, BorderD
   @IBInspectable open var cornerRadius: CGFloat = CGFloat.nan {
     didSet {
       configureCornerRadius()
+      configureBorder()
       configureMaskShadow()
     }
   }
@@ -19,6 +20,7 @@ open class AnimatableButton: UIButton, CornerDesignable, FillDesignable, BorderD
   open var cornerSides: CornerSides  = .AllSides {
     didSet {
       configureCornerRadius()
+      configureBorder()
       configureMaskShadow()
     }
   }
--- a/IBAnimatable/AnimatableButton.swift
+++ b/IBAnimatable/AnimatableButton.swift
@@ -12,12 +12,14 @@ open class AnimatableButton: UIButton, CornerDesignable, FillDesignable, BorderD
   @IBInspectable open var cornerRadius: CGFloat = CGFloat.nan {
     didSet {
       configureCornerRadius()
+      configureMaskShadow()
     }
   }
 
   open var cornerSides: CornerSides  = .AllSides {
     didSet {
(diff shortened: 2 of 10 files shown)


JakeLin:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Consider adding supporting documentation to this change. Documentation can be found in the <code>docs</code> directory.</td>
    </tr>
  </tbody>
</table>

### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
PresentationAnimationType.swift | 61 | Lines should not have trailing whitespace. 
TransitionAnimationType.swift | 249 | Lines should not have trailing whitespace. 
TransitionPresenterManager.swift | 20 | Lines should not have trailing whitespace. 


<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

JakeLin (changes requested):
Thanks for picking up this 👍, your changes can preview shadow in Interface Builder when we use `layer.mask` for corner radius. we also need to change `CornerDesignable` as well as below:

```
public func configureCornerRadius() {
    if !cornerRadius.isNaN && cornerRadius > 0 {
      // Use `layer.mask` instead of `layer.cornerRadius` to support shadow
      layer.cornerRadius = 0.0

      // if a layer mask is specified, remove it
      layer.mask?.removeFromSuperlayer()

      layer.mask = cornerSidesLayer()
    }
  }
```

and something similar in `configureCornerRadius` of `public extension CornerDesignable where Self: UICollectionViewCell`.

Then we use `layer.mask` instead of `layer.cornerRadius` to draw corner radius for all cases including 4 corners. Which can support `UIButton` with a background image with a shadow. In this case, `MaskDesignable` will draw another shadow `layer`, you can find it in `configureMaskShadow` of `MaskDesignable`. This change actually is very good, because it can improve the rendering performance without **Color blended layers**. Thanks for you original change for supporting difference corner of radius, which makes that improvement in place.


We may need to change `AnimatableTextView` and `AnimatableLabel` as well. Although I never apply `cornerRadius` to `AnimatableLabel` 😁

JakeLin:
simulator screen shot dec 20 2016 9 16 22 am
Before changing this, there is **Color blended layers** when use corner radius, it is OK since it is from UIKit.

simulator screen shot dec 20 2016 9 17 24 am
After changing this, there is no more  **Color blended layers** which can improve the performance a lot 🎉 , Great PR 👏👏👏

JakeLin:
@tbaranes Do you know why Danger was using my account? I am not a bot thought 😁

tbaranes:
Haha. I asked myself the same question when I saw it yesterday! Since it's using the right account in the other PR, let's see if that happen again.

JakeLin:
I tried to fix the issue when using the background image with `AnimatableButton`.

Before 2d7edee

screen shot 2016-12-28 at 9 27 56 am

After 2d7edee

screen shot 2016-12-28 at 9 57 02 am

I can see there are some issues with `border` now. It is very hard to make all corner radius, mask and border work well together.

Better to not merge this branch and I need more time to find a better solution.

tbaranes:
Just take a look at this PR, what's the status here? What's missing? Is there some regressions?

mmadjer:
Sorry guys, I was on holiday and did not find time since I'm back to have a look into that.
I can see that @JakeLin did something, not quite sure what the status is.

JakeLin:
This issue only happens in some very rare case. When we use the `backgroundImage` property of the `AnimatableButton` and set the `cornerRaidus` and `shadow` at the same time. It won't happen if we don't use the `backgroundImage` property.

The problem is caused by UIKit's *Clip to Bounds *, when we set it to `true` or `false`, it can only support some feature like `cornerRaidus` or `shadow`. So I think that's the problem of UIKit. 

To fix it, I tried to use `mask` for `cornerRaidus` not `layer.cornerRaidus`. @mmadjer has used `mask` for setting some corner of the `cornerRaidus` #356. I tried to apply the same method to all corners. It should work because we have used the method for `MaskDesignable`. Before @Abdul-Moiz find a way to use a separate layer for supporting shadow #267.  But after the commit 2d7edee, it ended up some issue like the border is miss( you can see it above).

We are not only fixing this issue; we try to find a good solution to support corner radius, mask and border, shadow work well together. I haven't found any other Open Source libraries can do that. I believe we are the best to support them together but still have some rare case like this issue. I think we are toward the right direction. But I still need more time to investigate and try out the solution. I also need some other eyes to find out the current problem (missing border issue).

JakeLin:
I close this now since it only happens in a rare case. I may have a look again when I have more time.

PR #379, opened 2016-12-22
Fix view translate {x,y} for {slide, squeeze} animation

tbaranes:
Since the final x, y were wrong, some animations was _failing_. In my case, on iPad, since my view was at the bottom of the screen, doing a `slide(.out, .down)` the view was going out of the screen way too fast (distance = iPad height). 

Less talk, more example:

Before the fix, after a `slide(.out, .down)`:

screen shot 2016-12-22 at 09 34 18

With the fix:

screen shot 2016-12-22 at 09 44 57

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ None
 None
 
 #### Bugfixes
-None
+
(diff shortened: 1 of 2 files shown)


JakeLin (approved):
Cool, let's merge it.

lastMove:
Nice One !

PR #380, opened 2016-12-22
Add custom gradient start point

### LINKED ISSUE #378: Custom gradient startPoint
Probably something like: `.custom(CGPoint, CGPoint)`, or `.custom(startX, startY, endX, endY)`

tbaranes:
Close #378 

Using `StartPoint.custom(0,0.5,0,0)`, it will render:

screen shot 2016-12-22 at 10 44 43

That may be useful! After a quick first look to update the example app, I have to say that I'm a bit lost with the `ParamType` 😨 @lastMove How am I supposed to create `ParamType` using custom parameters?

**Missing:**

- [ ] Example app
- [x] Changelog

### DIFF
--- a/IBAnimatable/GradientStartPoint.swift
+++ b/IBAnimatable/GradientStartPoint.swift
@@ -27,25 +27,23 @@ extension GradientStartPoint {
       return
     }
 
-    let nameAndParams = AnimationType.extractNameAndParams(from: string)
-    let name = nameAndParams.name
-    let params = nameAndParams.params
+    let (name, params) = AnimationType.extractNameAndParams(from: string)
     switch name {
     case "top":
       self = .top
-    case "topRight":
+    case "topright":
       self = .topRight
     case "right":
       self = .right
-    case "bottomRight":
+    case "bottomright":
       self = .bottomRight
     case "bottom":
       self = .bottom
-    case "bottomLeft":
+    case "bottomleft":
       self = .bottomLeft
     case "left":
       self = .left
-    case "topLeft":
+    case "topleft":
       self = .topLeft
     case "custom":
       self = .custom(start: CGPoint(x: params[safe: 0]?.toDouble() ?? 0,
--- a/IBAnimatable/GradientStartPoint.swift
+++ b/IBAnimatable/GradientStartPoint.swift
@@ -4,8 +4,9 @@
 //
 
 import Foundation
+import UIKit
 
-public enum GradientStartPoint: String, IBEnum {
+public enum GradientStartPoint: IBEnum {
   case top
   case topRight
   case right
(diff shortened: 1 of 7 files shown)


lastMove:
Hello, 
Nice feature ! 
How do you want to render it ? 
Presently in the Playground part of the app in "predefined gradients" section; 
there is a Picker with 2 components (one for the gradient, the other one for the startPoint).
So it will be difficult to had 4 more components to customise that. 
Maybe we should create another section.
ParamType allows To generate and manage arrays of possible values for a pickerComponent. 
And you can generate this array from enum; Sample : 
````Swift  
let gradientValues = ParamType(fromEnum: GradientStartPoint.self)
print(gradientValues.value(at: 0))
// "top"
print(gradientValues.value(at: 4))
// "bottomRight"
````
It's like an array generated from enum values. 

It works also to generate array of possible values from number interval : 

````Swift
private let numberParam = ParamType.number(min: -50, max: 50, interval: 5, ascending: true, unit: "")
// this will generate this array : [-50, -45, -40, -35, ... , 35, 40, 45, 50]
````
So numberParam.value(at: 0) will gives -50.

**Comment by tbaranes** [comment]
I was thinking to add directly a predefined startPoint, but you are right, it will be for the best to add a section to render it.
Thanks for the explanation! I will give it a try.

**Comment by tbaranes** [comment]
Working perfectly @lastMove! Thanks for the tips. 

simulator screen shot 22 dec 2016 14 22 29


One more question, any suggestion how we should update the current example (it won't compile since .custom isn't compatible with the other `startPoint`). I see that for the animation you used `PickerEntry`, should I do the same or is there a better existing solution?

File: IBAnimatable/GradientStartPoint.swift [hunk 1]
```diff
@@ -14,4 +15,45 @@ public enum GradientStartPoint: String, IBEnum {
...
+      self = .bottomLeft
+    case "left":
+      self = .left
+    case "topLeft":
+      self = .topLeft
+    case "custom":
+      self = .custom(start: CGPoint(x: params[safe: 0]?.toDouble() ?? 0,
+                                    y: params[safe: 1]?.toDouble() ?? 0),
+                     end: CGPoint(x: params[safe: 2]?.toDouble() ?? 0,
+                                  y: params[safe: 3]?.toDouble() ?? 0))
```
JakeLin, line 52: Great method from @lastMove for safe extracting data from params

JakeLin (approved):
I think it is ready to go👍

tbaranes:
Not really, the gradient example is still broken 😬 Open to any suggestion

IBAnimatable/GradientStartPoint.swift
```diff
@@ -27,25 +27,23 @@ extension GradientStartPoint {
...
 
-    let nameAndParams = AnimationType.extractNameAndParams(from: string)
-    let name = nameAndParams.name
-    let params = nameAndParams.params
+    let (name, params) = AnimationType.extractNameAndParams(from: string)
     switch name {
     case "top":
       self = .top
-    case "topRight":
+    case "topright":
```
JakeLin, line 34: haha, I always forgot that.

lastMove: :) It's error-prone. I saw it when testing with the exemple app.

lastMove:
Should be good know. 
@tbaranes the problem was that since GradientStartPoint does not anymore have a String RawValue  `ParamType.enumeration(values: ["top", "topLeft", "topRight", "left", "right", "bottom", "bottomRight", "bottomLeft"]) `

tbaranes:
@lastMove thanks for that!

Issue #382, opened 2016-12-26
Proposal : Improve Chain-ability


lastMove:
The goals of this proposal is to improve the Chainability based on the existing with the less possible breaking changes.  ( #14 ) 
 
### Introduce `AnimationConfiguration`  
  One of the problems of Animatable’s Chainability is Animation Parameters. 
 AnimationType is not enough to describe an Animation. 
The animation Parameters   (Damping, Velocity, duration, force, delay) are stored directly in the View.

````Swift
struct AnimationConfiguration { 	
       let damping: Double,
       let velocity: Double,  	
       let duration: Double,
       let delay: Double
}
````

Now when animating we can do 
:
`let conf = AnimationConfiguration(damping:0.9, velocity: 0.2, duration: 0.3, delay: 0)  view.animate(.Pop, configuration: conf)
`
It’s not a breaking change since configuration is an optional Parameters, and if not specified or nil the old method will be used.

## 2 - Base the callback system on promises.
 A good iOS promises library : [Facebook’s Bolts](https://github.com/BoltsFramework/Bolts-Swift)

Welcome to AnimationPromise : 
```Swift

struct AnimationPromise {
	func completion(() -> Void)
	func then(AnimationType, AnimationConfiguration) -> AnimationPromise 
	func and(AnimationType, AnimationConfiguration) -> AnimationPromise
}

```
The `animate` method will now return an `AnimationPromise`, and will not take anymore in parameter the completion closure.

### Before : 

```Swift
view.damping = 1
view.delay = 1  view.duration = 2 

view.animate(.slide(.in, .left)) { 
	view.duration = 1
	view.damping = 0.2  	
        view.animate(.pop) {
		view.duration = 2 
		view.damping = 1 
		view.delay = 0.5 		
               view.animate(.slide(.out, .left)) {
			self.label.text = "Done ;)" 
		}
	} 
}


```

### After : 
```Swift

view.animate(.slide(.in, .left), AnimationConfiguration(duration: 2, damping: 1, delay: 1))
	.then(.pop, AnimationConfiguration(duration: 1, damping: 0.2))
	.then(.slide(.out, .left), AnimationConfiguration(duration: 2, damping: 1, delay: 0.5))
	.completion { 
		self.label.text = "Done ;)"; 
       }
```

It's still a work in progress, what do you think about this ?

JakeLin:
@lastMove Thanks for starting the initiative. 

[...]

PR #385, opened 2017-01-06
Chainable animations

lastMove:
This PR is a Prototype of the Chainable feature; It's almost fully working and without breaking changes : 

```Swift
 self.blurEffectView.animate(animation: .pop(repeatCount: 1))
      .delay(1)
      .then( .slide(way: .out, direction: .left))
      .delay(1)
      .then(.slide(way: .in, direction: .left))
      .then(.pop(repeatCount: 1))
      .completion {
	print("Completed !");
    }
```

However I am not entirely satisfy by some points.  I need your help :

1 - Should The view property of AnimationPromise be `weak` (or maybe `unowned`)  ? 

2 - I wanted to say that the argument of AnimationPromise `initalizer` has to be an UIView implementing Animatable.  The only way I fond is by using Generics. 
Is there another solution ?  I need that because the animate Method was defined inside zextension Animatable where Self: UIView` 

3 - when doing view.animate(animation: .pop).delay(2).delay(3).then(.squash) 
the delay applied is 3s. What do you think about that, is it  okay ? 

4 - Coherence issue about   those methods : 

````Swift     public func zoom(_ way: AnimationType.Way, invert: @autoclosure @escaping Bool = default, completion: @autoclosure AnimatableCompletion? = default)

    public func flip(axis: AnimationType.Axis, completion: @autoclosure @escaping AnimatableCompletion? = default)
````
etc
Presently thes methods are publicand not chainable.  Should we put these method privates and encourage the usage of .animate().completion ? 

5 - To avoid editing a lot of methods i did : 

```Swift
    self.duration = configuration.duration
    self.delay = configuration.delay
    self.damping = configuration.damping
    self.velocity = configuration.velocity
    self.force = configuration.force
````
Before Each Animation i modify, theses properties from the view. It can be confusing, or break working code I guess.

### CODE REVIEW

**Comment by JakeLin** [comment]
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Any changes to library code need a summary in the Changelog.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is classed as Work in Progress</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Consider adding supporting documentation to this change. Documentation can be found in the <code>docs</code> directory.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

**Comment by JakeLin** [comment]
@lastMove great step for chainable animations 👍  Please have a look at my comments. 

> 1 - Should The view property of AnimationPromise be weak (or maybe unowned)?

I think it should be fine because the UI elements (e.g. `AnimatableView`) won't hold `AnimationPromise` as a property.

> 2 - I wanted to say that the argument of AnimationPromise initializer has to be a UIView implementing Animatable.  The only way I found is by using Generics.
Is there another solution?  I need that because the animate Method was defined inside `extension Animatable where Self: UIView'

I didn't find other solution either, looks like we have to use Generics.

> 3 - when doing view.animate(animation: .pop).delay(2).delay(3).then(.squash)
the delay applied is 3s. What do you think about that, is it okay ?

It is perfect because Apple uses `NSTimeInterval` (which is *second*) for `delay` as well. https://developer.apple.com/reference/uikit/uiview/1622451-animate

> 4 - Coherence issue about those methods:

I don't mind we make them all be `private` because we will have better syntax to handle chainable animations. And we need to upgrade the version number to major version 4 

> 5 - To avoid editing a lot of methods i did :

There are some side effects after calling `doAnimation` method to update the view's properties like `duration`, `delay` and so on. If we add `configuration: AnimationConfiguration? = nil` to the methods like `public func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, configuration: AnimationConfiguration? = nil, completion: AnimatableCompletion? = nil)` and `func animateIn(x: CGFloat, y: CGFloat, scaleX: CGFloat, scaleY: CGFloat, alpha: CGFloat, configuration: AnimationConfiguration? = nil, completion: AnimatableCompletion? = nil)` then, inside `animateIn` we can use either use `configuration`'s or `self`'s properties like `let duration = configuration?.duration ?? self.duration` . That will not introduce any breaking changes.

Just a few of ideas, any feedback is welcome 😄.

File: IBAnimatable/Chainable.swift [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+//  IBAnimatable
+//
+//  Created by jason akakpo on 01/01/2017.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+typealias AnimationTuple = (type: AnimationType, conf: AnimationConfiguration)
```
JakeLin: may use `configuration` here

IBAnimatable/Animatable.swift
```diff
@@ -66,7 +66,34 @@ public extension Animatable where Self: UIView {
...
+    return promise.then(animation)
+  }
+
+  public func animate(animation: AnimationType? = nil, completion: AnimatableCompletion? = nil) {// here for retro-compatibility
+   // self.animate(animation: animation, promise: nil)
+   // self.animate(animation)
+    self.animate(animation: animation ?? self.animationType).completion(completion)
+  }
+
+  public func delay(delay: Double) -> AnimationPromise<Self> {
```
JakeLin: Does it mean we can call `delay` first like `self.blurEffectView.delay(1).animate(animation: .pop(repeatCount: 1))`?

IBAnimatable/Chainable.swift
```diff
@@ -0,0 +1,64 @@
...
+  private var animationList: [AnimationTuple] = []
+  private var delayForNextAnim: Double = 0
+  private var completion: AnimatableCompletion?
+
+  internal init(view: T) {
+    self.view = view
+  }
+
+
+  internal func animCompleted() {
```
JakeLin: don't need to specify `internal` here, also may use the full name like `animationCompleted`

IBAnimatable/Chainable.swift
```diff
@@ -0,0 +1,64 @@
...
+
+
+  public func completion(_ completion: AnimatableCompletion?) {
+    self.completion = completion
+  }
+
+   @discardableResult public func  then(_ animation: AnimationType, duration: Double? = nil, damping: CGFloat? = nil, velocity: CGFloat? = nil, force: CGFloat? = nil) -> AnimationPromise {
+    let animTuple = AnimationTuple(type: animation, conf: AnimationConfiguration(damping: damping ?? view.damping, velocity: velocity ?? view.velocity, duration: duration ?? view.duration, delay: delayForNextAnim, force: force ?? view.force))
+    self.animationList.append(animTuple)
+    if animationList.count == 1 { // If it's the first animation, launtch it immediately
```
JakeLin: If there is only one animation?

IBAnimatableApp/BlurEffectViewController.swift
```diff
@@ -19,6 +19,18 @@ class BlurEffectViewController: UIViewController {
...
+    super.viewDidLoad()
+    self.blurEffectView.animate(animation: .pop(repeatCount: 1))
+      .delay(1)
+      .then( .slide(way: .out, direction: .left))
+      .delay(1)
+      .then(.slide(way: .in, direction: .left))
+      .then(.pop(repeatCount: 1))
+      .completion {
+        self.blurEffectView.alpha = 0
+    }
```
JakeLin: nice syntax 🎉

JakeLin (commented):
Please see my comments, thanks.

IBAnimatable/Animatable.swift
lastMove: Yes !

IBAnimatable/Chainable.swift
lastMove: ok

IBAnimatable/Chainable.swift
lastMove: yes

IBAnimatableApp/BlurEffectViewController.swift
lastMove: :)

IBAnimatable/Animatable.swift
```diff
@@ -66,7 +66,34 @@ public extension Animatable where Self: UIView {
...
+    let promise = AnimationPromise(view: self)
+    return promise.delay(delay)
+  }
+
+  internal func doAnimation(animation: AnimationType? = nil, configuration: AnimationConfiguration, promise: AnimationPromise<Self>) -> AnimationPromise<Self> {
```
tbaranes: You can drop the internal here

IBAnimatable/Animatable.swift
```diff
@@ -66,7 +66,34 @@ public extension Animatable where Self: UIView {
...
+
+    // Dirty code, need to be upgraded.
+    self.duration = configuration.duration
+    self.delay = configuration.delay
+    self.damping = configuration.damping
+    self.velocity = configuration.velocity
+    self.force = configuration.force
```
tbaranes: Can't we keep the configuration directly instead of store each property?

IBAnimatable/Animatable.swift
tbaranes: Shouldn't it be (file)private?

IBAnimatable/Animatable.swift
```diff
@@ -66,7 +66,34 @@ public extension Animatable where Self: UIView {
...
+
+    let completion = {() -> Void in
```
tbaranes: You can drop `() -> Void in`

IBAnimatable/Chainable.swift
```diff
@@ -0,0 +1,64 @@
+//
+//  Chainable.swift
```
tbaranes: Should we name it `AnimationChainable` to be more coherent?
You can also drop these lines to follow IBAnimatable convention!

IBAnimatable/Chainable.swift
```diff
@@ -0,0 +1,64 @@
...
+  let delay: Double
+  let force: CGFloat
+
+}
+
+
+public class AnimationPromise<T: UIView where T: Animatable> {
+
+  private var view: T
+  private var animationList: [AnimationTuple] = []
```
tbaranes: Don't know what do you think about that syntax, but I personally prefer: `private var animationList = [AnimationTuple]()` since it's more concise 
Any opinions on this @JakeLin ?

IBAnimatable/Chainable.swift
```diff
@@ -0,0 +1,64 @@
...
+  private var delayForNextAnim: Double = 0
```
tbaranes: As above, I'm in favor of the concise syntax: `var delayForNextAnim = 0.0`. Any thought @JakeLin ?

tbaranes: Also, in favor for a full name: `delayForNextAnimation`

IBAnimatableApp/BlurEffectViewController.swift
tbaranes: 👏

tbaranes (commented):
Good work @lastMove ! It's a pretty good first step 👍
I let you a few comments, but mostly about sugar syntax.

IBAnimatable/Animatable.swift
tbaranes: You should use `TimeInterval` instead of `Double`

tbaranes:
2 - I think the implementation is nice this way. Generics is a good way to handle it.
4 - Not sure about privatise them all, this may have improve the syntax for chainable animations, the sugar syntax is still a nice way to use single animations. Still, I'm not against that request, so, it's up to you! I think, we can do it in two releases: first, release with the new chainable syntax and a warning saying that we will privatise the single animation in a next release, then privatise them. 

For the others points, I globally agree with @JakeLin 

Very nice PR! The main point that's annoying me here is the usage of all these animations parameters. I think we can find a better way using `AnimationConfiguration` directly. Also, I read there's some code that was kept for retro-compatibility, but if we go with the point 4 and includes breaking changes, I think we can do it everywhere. It's a good moment to clean up the animations process and start again with a better one!

IBAnimatable/Chainable.swift
JakeLin: @tbaranes , yes Swift can figure out the type.

IBAnimatable/Chainable.swift
JakeLin: I don't have any preference for this. It is up to @lastMove then. or we can vote 😄

IBAnimatable/Chainable.swift
JakeLin: Good idea @tbaranes

JakeLin:
@tbaranes do you mean we should use `AnimationConfiguration` as a public parameter instead of `velocity`, `force` etc.? I made a comment on #382 to ask @lastMove to use separate parameters because most of the time the user will use the default values. In that case, we don't need to specify any additional parameters except for `AnimationType`. I am open to discussing because that was my initial idea of the API.

API using `AnimationConfigureation`
```
view.animate(.slide(.in, .left), AnimationConfiguration(duration: 2, damping: 1, delay: 1))
	.then(.pop, AnimationConfiguration(duration: 1, damping: 0.2))
	.then(.slide(.out, .left), AnimationConfiguration(duration: 2, damping: 1, delay: 0.5))
	.completion { 
		self.label.text = "Done ;)"; 
       }
```

API using separate parameters
```
view.animate(.slide(.in, .left), duration: 2, damping: 1)
        .delay(1)
	.then(.pop, duration: 1, damping: 0.2)
	.then(.slide(.out, .left), duration: 2, damping: 1)
        .delay(0.5)
	.completion { 
		self.label.text = "Done ;)"; 
       }
```

tbaranes:
I think both syntax are acceptable. `AnimationConfiguration` will give us a cleaner code for IBAnimatable, more flexible and scalable, but it's true that it seems more complex to use for the user.

So, after seeing both, I agree with both!

IBAnimatable/Animatable.swift
lastMove: Good call !

IBAnimatable/Animatable.swift
lastMove: Not private because it's called by AnimationPromise

IBAnimatable/Animatable.swift
lastMove: ok

IBAnimatable/Animatable.swift
lastMove: Seems that I have to keep internal here because it's in a public extension

lastMove:
I updated the PR following your comments. 
I privatise the old animation method and added AnimationConfiguration in parameter. Not optional to simplify the code.

IBAnimatable/Animatable.swift
```diff
@@ -66,75 +66,66 @@ public extension Animatable where Self: UIView {
...
   }
 
-  public func delay(delay: Double) -> AnimationPromise<Self> {
+  public func delay(delay: TimeInterval) -> AnimationPromise<Self> {
     let promise = AnimationPromise(view: self)
     return promise.delay(delay)
   }
 
-  internal func doAnimation(animation: AnimationType? = nil, configuration: AnimationConfiguration, promise: AnimationPromise<Self>) -> AnimationPromise<Self> {
+  internal func doAnimation(animation: AnimationType? = nil, configuration: AnimationConfiguration, promise: AnimationPromise<Self>) {
```
tbaranes: You can remove that internal 😬

IBAnimatable/Animatable.swift
```diff
@@ -148,42 +139,42 @@ public extension Animatable where Self: UIView {
   }
 
   // MARK: - Animation methods
-  public func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, completion: AnimatableCompletion? = nil) {
+  internal func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, configuration: AnimationConfiguration, completion: AnimatableCompletion? = nil) {
```
tbaranes: `internal`

IBAnimatable/Animatable.swift
```diff
@@ -148,42 +139,42 @@ public extension Animatable where Self: UIView {
...
-      animateIn(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, completion: completion)
+      animateIn(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, configuration: configuration, completion: completion)
     case .out:
-      animateOut(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, completion: completion)
+      animateOut(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, configuration: configuration, completion: completion)
     }
   }
 
-  public func squeeze(_ way: AnimationType.Way, direction: AnimationType.Direction, completion: AnimatableCompletion? = nil) {
+  internal func squeeze(_ way: AnimationType.Way, direction: AnimationType.Direction, configuration: AnimationConfiguration, completion: AnimatableCompletion? = nil) {
```
tbaranes: `internal`

IBAnimatable/Animatable.swift
```diff
@@ -148,42 +139,42 @@ public extension Animatable where Self: UIView {
...
-      animateIn(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, completion: completion)
+      animateIn(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, configuration: configuration, completion: completion)
     case .out:
-      animateOut(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, completion: completion)
+      animateOut(x: values.x, y: values.y, scaleX: values.scaleX, scaleY: values.scaleY, alpha: 1, configuration: configuration, completion: completion)
     }
   }
 
-  public func rotate(direction: AnimationType.RotationDirection, repeatCount: Int, completion: AnimatableCompletion? = nil) {
+  internal func rotate(direction: AnimationType.RotationDirection, repeatCount: Int, configuration: AnimationConfiguration, completion: AnimatableCompletion? = nil) {
```
tbaranes: `internal`
I'm stopping here, but there's a few left in that file!

tbaranes: Second though for all these internal, shouldn't they be `private`?

IBAnimatable/Animatable.swift
```diff
@@ -533,12 +525,13 @@ private extension Animatable where Self: UIView {
     })
   }
 
-  func animateOut(x: CGFloat, y: CGFloat, scaleX: CGFloat, scaleY: CGFloat, alpha: CGFloat, completion: AnimatableCompletion? = nil) {
+  // swiftlint:disable function_parameter_count
+  func animateOut(x: CGFloat, y: CGFloat, scaleX: CGFloat, scaleY: CGFloat, alpha: CGFloat, configuration: AnimationConfiguration, completion: AnimatableCompletion? = nil) {
```
tbaranes: Is it possible to embed `x`, `y`, `scaleX`, `scaleY`, `apha` in `AnimationConfiguration` or it doesn't make sense?

IBAnimatableApp/BlurEffectViewController.swift
```diff
@@ -28,7 +28,7 @@ class BlurEffectViewController: UIViewController {
       .then(.slide(way: .in, direction: .left))
       .then(.pop(repeatCount: 1))
       .completion {
-        self.blurEffectView.alpha = 0
+        print("Done ;)")
```
tbaranes: Is that just for debug, if not I think we can find a better feedback to show the completion

IBAnimatable/Animatable.swift
lastMove: I couldn't remove  'internal' because it's in a public extension. 
Yes we could make them private and embed them in a private extension.

IBAnimatable/Animatable.swift
JakeLin: Can you move it inside `private extension Animatable where Self: UIView`?

IBAnimatable/Animatable.swift
JakeLin: We can group them in a `private extension` or `internal extension`. I think `private` should be fine.

IBAnimatable/Animatable.swift
JakeLin: If `AnimationConfiguration` only be used internal, it is OK to group `x`, `y` etc. inside. Otherwise, we may expose too many details.

lastMove: I don't think that it makes sense, since AnimationConfiguration is used at a higher level than those variables.

JakeLin: I agree with @lastMove now since they are in different level of abstraction.

tbaranes: 👍

IBAnimatable/Animatable.swift
```diff
@@ -66,44 +66,64 @@ public extension Animatable where Self: UIView {
...
+    self.animate(animation: animation ?? self.animationType).completion(completion)
+  }
+
+  public func delay(delay: TimeInterval) -> AnimationPromise<Self> {
+    let promise = AnimationPromise(view: self)
+    return promise.delay(delay)
+
+  }
+
+  internal func doAnimation(animation: AnimationType? = nil, configuration: AnimationConfiguration, promise: AnimationPromise<Self>) {
```
lastMove: This `internal` is mandatory ;)

lastMove:
The code should be ready now. 
Could you review it a last time please. 
Still Miss Documentation. 
I don't think that we need to add a demo in IBAnimatableApp.

tbaranes:
If it's ready to go -> can you remove the `[WIP]` from the PR title? It's necessary to make Danger run again. Also, the CHANGELOG entry is missing
I will take one more look later today!

IBAnimatable/AnimationChainable.swift
```diff
@@ -0,0 +1,61 @@
...
+
+typealias AnimationTuple = (type: AnimationType, configuration: AnimationConfiguration)
+
+struct AnimationConfiguration {
+  let damping: CGFloat
+  let velocity: CGFloat
+  let duration: TimeInterval
+  let delay: TimeInterval
+  let force: CGFloat
+
```
tbaranes: One newline left here!

IBAnimatable/AnimationChainable.swift
```diff
@@ -0,0 +1,61 @@
...
+  private var view: T
+  private var animationList = [AnimationTuple]()
+  private var delayForNextAnimation = 0.0
+  private var completion: AnimatableCompletion?
+
+  init(view: T) {
+    self.view = view
+  }
+
+
```
tbaranes: Is your SwitLint up to date? Pretty two following newlines that will fire a warning

tbaranes (commented):
Can you check if your SwiftLint if up to date? (Danger will fire the warnings otherwise).

I will give a last review tomorrow to confirm, but I seem ready to go for me.

IBAnimatableApp/PresentingViewController.swift
```diff
@@ -192,16 +192,16 @@ extension PresentingViewController {
   func showPicker() {
     pickerView.reloadAllComponents()
     resetSelectedItemPicker()
-    dimmingPickerView.fade(.in)
-    containerPickerView.slide(.in, direction: .up)
+    dimmingPickerView.animate(animation: .fade(way: .in))
+    containerPickerView.animate(animation: .slide(way: .in, direction: .up))
```
tbaranes: A though outside the PR, should we keep the label name `animation:`?

JakeLin:
@lastMove One more thing, can please add a new Swift Playground example to use the new API, we may keep the old for now for reference until we remove them (Rename `Chaining Animations.xcplaygroundpage` to `Old Chaining Animations.xcplaygroundpage` ?). Then we can finally update the [README.md | Chaining Animations](https://github.com/IBAnimatable/IBAnimatable#chaining-animations) with nice syntax 🎉

tbaranes:
It would be great to merge that PR. @lastMove do you think you will have time to finalise it? Otherwise, do you mind if I (or any one volunteer) go ahead and fix the small last comments?

JakeLin:
@tbaranes thanks for following up this PR, do you think we should release 4 once this PR is done? I think that is a big improvement for the project.

tbaranes:
Fully agree, let's merge all the PRs running, making all the breaking change... then release!

tbaranes:
@JakeLin @lastMove I go ahead and do the last diffs:
- Rebased on master
- Fixing Swiftlint warnings
- Removing `animation` label name (I think its more swifty that way, let me know if you disagree)
- Updating `Chaining Animations` playground. The old one wasn't compiling anymore, so I think it's ok to update it directly with the new syntax
- Update README

Missing:
- CHANGELOG: not sure what breaking changes are we introducing here, so I prefer let someone else do it directly

Close #382

JakeLin:
@tbaranes please give some time to review this PR again. It is our biggest change for animation API. Very excited 🤗.

lastMove:
Sorry, I was very busy these days. 
Thanks @tbaranes for finishing this ! 👍

IBAnimatable.playground/Pages/Chaining Animations.xcplaygroundpage/Contents.swift
```diff
@@ -14,29 +14,19 @@ PlaygroundPage.current.liveView = iPhoneView
...
-}
+////: Chain your animations
+view.animate(.squeezeFade(way: .in, direction: .down))
+    .then(.pop(repeatCount: 1))
+    .then(.shake(repeatCount: 1))
+    .then(.squeeze(way: .in, direction: .down))
+    .then(.wobble(repeatCount: 1))
+    .then(.flip(along: .x))
+    .then(.flip(along: .y))
+    .then(.slideFade(way: .out, direction: .down))
```
lastMove, line 25: great

IBAnimatableApp/BlurEffectViewController.swift
```diff
@@ -19,6 +19,9 @@ class BlurEffectViewController: UIViewController {
     return values
   }()
 
+  override func viewDidLoad() {
+    super.viewDidLoad()
+  }
```
lastMove: Why Did I add this ? ^^

IBAnimatable.playground/contents.xcplayground
```diff
@@ -3,6 +3,6 @@
     <pages>
         <page name='Predefined Animations'/>
         <page name='Animation Properties'/>
-        <page name='Chaining Animations'/>
+        <page name='Old Chaining Animations'/>
```
JakeLin: Do we need to rename this playground?

IBAnimatable/Animatable.swift
```diff
@@ -22,12 +22,12 @@ public protocol Animatable: class {
   /**
    Animation duration (in seconds)
    */
-  var duration: Double { get set }
+  var duration: TimeInterval { get set }
```
JakeLin, line 25: Although `public typealias TimeInterval = Double`, It is great to change it to `TimeInterval` 👍. And shall we change all conforming classes to reflect this change to make the code more understandable? If we do, we need to apply to `var delay: TimeInterval { get set }` as well.

IBAnimatable/Animatable.swift
```diff
@@ -66,44 +66,63 @@ public extension Animatable where Self: UIView {
...
   }
 
-  public func animate(animation: AnimationType? = nil, completion: AnimatableCompletion? = nil) {
+  @discardableResult
+  public func animate(_ animation: AnimationType, duration: TimeInterval? = nil, damping: CGFloat? = nil, velocity: CGFloat? = nil, force: CGFloat? = nil) -> AnimationPromise<Self> {
+    return AnimationPromise(view: self).then(animation, duration: duration, damping: damping, velocity: velocity, force: force)
+  }
+
+  @available(*, deprecated, message: "use animate(AnimationType).then(...).completion instead. Will be removed in future versions")
+  public func animate(_ animation: AnimationType? = nil, completion: AnimatableCompletion? = nil) {// here for retro-compatibility
```
JakeLin: Since we have this method, we don't have breaking API for `func animate(_ animation: AnimationType? = nil, completion: AnimatableCompletion? = nil)`. But shall we remove it for release 4?

IBAnimatable/Animatable.swift
```diff
@@ -66,44 +66,63 @@ public extension Animatable where Self: UIView {
...
+    self.animate(animation ?? self.animationType).completion(completion)
+  }
+
+  public func delay(_ delay: TimeInterval) -> AnimationPromise<Self> {
+    let promise = AnimationPromise(view: self)
+    return promise.delay(delay)
+
+  }
+
+  internal func doAnimation(_ animation: AnimationType? = nil, configuration: AnimationConfiguration, promise: AnimationPromise<Self>) {
```
JakeLin, line 80: Remove `internal` here?

IBAnimatable/Animatable.swift
```diff
@@ -115,47 +134,50 @@ public extension Animatable where Self: UIView {
...
     }
   }
+}
+
+fileprivate extension Animatable where Self: UIView {
 
   // MARK: - Animation methods
-  public func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, completion: AnimatableCompletion? = nil) {
-    let values = computeValues(way: way, direction: direction, shouldScale: false)
+  func slide(_ way: AnimationType.Way, direction: AnimationType.Direction, configuration: AnimationConfiguration, completion: AnimatableCompletion? = nil) {
```
JakeLin, line 140: All of them are breaking APIs, right? I don't mind to change them in release 4 and let the user use the new and nice API.

IBAnimatable/AnimationChainable.swift
JakeLin: Remove it?

IBAnimatable/AnimationChainable.swift
```diff
@@ -0,0 +1,60 @@
...
+  public func completion(_ completion: AnimatableCompletion?) {
+    self.completion = completion
+  }
+
+  @discardableResult
+  public func  then(_ animation: AnimationType, duration: TimeInterval? = nil, damping: CGFloat? = nil, velocity: CGFloat? = nil, force: CGFloat? = nil) -> AnimationPromise {
+    let animTuple = AnimationTuple(type: animation, configuration: AnimationConfiguration(damping: damping ?? view.damping, velocity: velocity ?? view.velocity, duration: duration ?? view.duration, delay: delayForNextAnimation, force: force ?? view.force))
+    animationList.append(animTuple)
+    if animationList.count == 1 { // If it's the only animation, launch it immediately
+      view.doAnimation(animation, configuration: animTuple.configuration, promise: self)
```
JakeLin, line 48: Can we put `delay ` before `then`? Will it cause problem if we can?

IBAnimatable/AnimationChainable.swift
```diff
@@ -0,0 +1,60 @@
...
+  private var animationList = [AnimationTuple]()
+  private var delayForNextAnimation = 0.0
+  private var completion: AnimatableCompletion?
+
+  init(view: T) {
+    self.view = view
+  }
+
+   func animationCompleted() {
+      animationList.remove(at: 0)
```
JakeLin: May use `animationList.removeFirst` to make it more specify?

I got an idea here. Because `removeFirst` from an `Array` is an `O(N)` operation. It means anytime run an animation will be `O(N^2)` which is `O(N) + O(N-1) + ... + O(1)`. If we have an index for current running animations in `animationList` and then we increase the index every time. And get the `currentAnimation` via  `if let currentAnimation = animationList[currentAnimationIndex]` in line 33. 

But since we don't have a big list of animations. Currently, implementation is totally fine.

IBAnimatable/AnimationChainable.swift
```diff
@@ -0,0 +1,60 @@
...
+      if let anim = animationList.first {
```
JakeLin: Rename to `currentAnimation`?

IBAnimatable/AnimationType.swift
```diff
@@ -59,9 +59,7 @@ extension AnimationType: IBEnum {
       self = .none
       return
     }
-    let nameAndParams = AnimationType.extractNameAndParams(from: string)
-    let name = nameAndParams.name
-    let params = nameAndParams.params
+    let (name, params) = AnimationType.extractNameAndParams(from: string)
```
JakeLin, line 62: simpler, 👍

README.md
```diff
@@ -130,27 +130,16 @@ You can play around with all animations with different properties in the [Swift
...
-  }
-}
+view.animate(.squeezeFade(way: .in, direction: .down))
+    .then(.pop(repeatCount: 1))
+    .then(.shake(repeatCount: 1))
+    .then(.squeeze(way: .in, direction: .down))
+    .then(.wobble(repeatCount: 1))
+    .then(.flip(along: .x))
+    .then(.flip(along: .y))
+    .then(.slideFade(way: .out, direction: .down))
```
JakeLin, line 141: How beautiful it is 🎉🎉🎉

JakeLin (approved):
Well done guys @lastMove @tbaranes 👍👍👍 . I have leave some small comments, please have a look. I think is PR is ready to go 🎉🎉🎉.

IBAnimatable.playground/contents.xcplayground
tbaranes: That's weird, the playground hasn't been renamed 🤔

IBAnimatable/Animatable.swift
tbaranes: I tried, but that creates SwiftLint warnings since we have to use an explicit type (instead of typealias) for the inspectable properties

IBAnimatable/Animatable.swift
tbaranes: Not possible, that's a mandatory internal

IBAnimatable/AnimationChainable.swift
tbaranes: I'm not sure to see what are you asking for here 😓

tbaranes:
Thanks for your feedbacks @lastMove @JakeLin! All fixed 👍

IBAnimatable/Animatable.swift
tbaranes: Agreed, since we are releasing v4, let's break it.

IBAnimatable/Animatable.swift
JakeLin: @tbaranes OK then

IBAnimatable/Animatable.swift
JakeLin: Thanks Interface Builder.  I hope they can fix this in next version of Xcode to support `TimeInterval`

IBAnimatable/AnimationChainable.swift
JakeLin: I was thinking something like
```
view.delay(2)
.animate(.pop(repeatCount: 1))
```

Is it possible to delay the first animation?

JakeLin:
@tbaranes thanks for addressing all comments 💪 . Except for the last concern about 
```
view.delay(2)
.animate(.pop(repeatCount: 1))
```

This PR is ready to go. 🎉🎉🎉🎉🎉🎉

tbaranes:
@lastMove What do you think about this? Is it straightforward or does it imply something else?

lastMove:
I think that it's currently possible.
That's why there is a delay (which returns an AnimationPromise) method in the `Animatable` protocol extension
just have to do :
view.delay(0.5).then ...


On 10 mars 2017 à 13:35 +0100, Tom Baranes <notifications@github.com>, wrote:

@lastMove<https://github.com/lastMove> What do you think about this? Is it straightforward or does it imply something else?

JakeLin:
@lastMove @tbaranes I just tried out to call `.delay()` before `.animate` like

```
view.delay(2).animate(.squeezeFade(way: .in, direction: .down))
    .then(.pop(repeatCount: 1))
    .delay(2)
    .then(.shake(repeatCount: 1))
    .delay(2)
    .then(.squeeze(way: .in, direction: .down))
    .delay(2)
    .then(.wobble(repeatCount: 1))
    .delay(2)
    .then(.flip(along: .x))
    .delay(2)
    .then(.flip(along: .y))
    .delay(2)
    .then(.slideFade(way: .out, direction: .down))
```
It came up an error as below:
```
Playground execution failed: error: Chaining Animations.xcplaygroundpage:12:6: error: value of type 'AnimationPromise<CircleView>' has no member 'animate'
view.delay(2).animate(.squeezeFade(way: .in, direction: .down))
```
It means we don't support that. I think it is OK. I don't know how hard to support it. If the user wants to delay the first animation. they can use other API to do it.

tbaranes:
We can also add this feature later if we have some feature requests.

JakeLin:
I also found an issue in the playground file `Animation Properties`. Since we don't support `view.animate()` any more. Then I change it to `view.animate(animationType)` like 

```
let animationType: AnimationType = .squeeze(way: .in, direction: .left)
view.duration = 1
view.delay = 5
view.damping = 0.5
view.velocity = 2
view.force = 1

view.animate(animationType)
```

But no matter how do I change the values for `duration`, `delay`, `damping`, `velocity` or `force`. They don't apply to `view.animate(animationType)` because they all set to `nil` in
```
public func animate(_ animation: AnimationType, duration: TimeInterval? = nil, damping: CGFloat? = nil, velocity: CGFloat? = nil, force: CGFloat? = nil) -> AnimationPromise<Self> {
    return AnimationPromise(view: self).then(animation, duration: duration, damping: damping, velocity: velocity, force: force)
  }
```

I think we need to provide a way to the user to custom the animation factors in code like what we did before.

How do you think? @tbaranes @lastMove

JakeLin:
agree with you, @tbaranes , `delay` before `animate` is not a big problem for our API. We can add it later if needed.

JakeLin:
@lastMove @tbaranes I was silly, the new API does support custom animation factors like 

```
let animationType: AnimationType = .squeeze(way: .in, direction: .left)
let duration: TimeInterval = 1
let delay: TimeInterval = 1
let damping: CGFloat = 0.8
let velocity: CGFloat = 2
let force: CGFloat = 1

view.animate(animationType, duration: duration, damping: damping, velocity: velocity, force: force)
    .delay(delay)
    .then(.shake(repeatCount: 3), duration: duration, damping: damping, velocity: velocity, force: force)
```

It is much nicer than the old one. Well done guys 💪💪💪. 

I have updated the Playground files to use the new API, please have a look at the last a few commits, thanks. If there is not problem, I think this PR is ready 🎉

lastMove:
Nice ! 
To have initial delay working you have to do : 
````Swift
view.delay(2).then(.squeezeFade(way: .in, direction: .down))
````

tbaranes:
Only the changelog entry is still missing, otherwise it's all good 👍

JakeLin:
@lastMove perfect, we do support delay first. will add an example for it. i really like the work we guys have well done ꉂ ೭(˵¯̴͒ꇴ¯̴͒˵)౨”

JakeLin:
@tbaranes i may write a small migration doc like we had for version 3. a little bit busy this week. may do it next week. shall we merge the pr now? or leave it after the doc completed.

tbaranes:
Up to you, I think we can wait or not, both are ok 😬

JakeLin:
@tbaranes merged 🎉 🎉 🎉 🎉 Thanks again to @lastMove and @tbaranes 👍👍👍

PR #386, opened 2017-01-08
Bump SwiftLint 0.15.0

tbaranes:
New opt-in rules in addition with the _enabled by default rules_ enabled:
 - `attributes`
 - `closure_end_indentation`
 - `first_where`
 - `number_separator`
 - `operator_usage_whitespace`
 - `prohibited_super_call`

### DIFF
--- a/IBAnimatable/PanInteractiveAnimator.swift
+++ b/IBAnimatable/PanInteractiveAnimator.swift
@@ -60,7 +60,7 @@ public class PanInteractiveAnimator: InteractiveAnimator {
     progress = min(max(progress, 0), 0.99)
 
     // Finish the transition when pass the threathold
-    let shouldFinishInteractiveTransition =  progress > 0.5 || speed > 1000
+    let shouldFinishInteractiveTransition = progress > 0.5 || speed > 1_000
     return (progress, shouldFinishInteractiveTransition)
   }
 }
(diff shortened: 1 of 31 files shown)


IBAnimatable/PanInteractiveAnimator.swift
```diff
@@ -60,7 +60,7 @@ public class PanInteractiveAnimator: InteractiveAnimator {
     progress = min(max(progress, 0), 0.99)
 
     // Finish the transition when pass the threathold
-    let shouldFinishInteractiveTransition =  progress > 0.5 || speed > 1000
+    let shouldFinishInteractiveTransition = progress > 0.5 || speed > 1_000
```
JakeLin, line 63: I don't quite understand why it is `1_000`?

JakeLin (approved):
Just a small comment. otherwise, it is great 👍, thanks for updating that.

IBAnimatable/PanInteractiveAnimator.swift
tbaranes: That's due to `number_separator`: separate big numbers by "_" to make it easier to read. I'm not a huge fan for 1000, but starting with the next version, we can configure it to warn for number only > 10_000

JakeLin: @tbaranes OK, cool.

Issue #251, opened 2016-07-19
playground can'f find IBAnimatable


ghost:
playground can'f find IBAnimatable
open the 'IBAnimatable.playground',the Xcode prompt No such module 'IBAnimatable'
how can I break this problem???TKS!!!

tbaranes:
Here the steps to follow if you want to use IBAnimatable playground:
1. Open IBAnimatableApp.xcodeproj
2. Select IBAnimatable Framework scheme and build it with Command + b
3. Select IBAnimatable.playground, choose one page in Swift playground, then click on "Assistant editor" button to split the playground. After that, select "Timeline" on the top of right-hand side to preview the animation. We can use Xcode menu "Editor" -> "Execute" to re-run the playground.

JakeLin:
Related to #102

ghost:
TKS,but still error! @tbaranes

ghost:
before i put the issue,I searched the issue and try it like #102,But still error @JakeLin

JakeLin:
@ysfox can you run the demo app? which version of Xcode are you using? Just try to restart Xcode few times if the problem is still there.

ghost:
the Demo running is Ok,just the playground error!,the Xcode is Version 7.3.1 (7D1014) @JakeLin

tbaranes:
What error are you getting? Do you have any logs in the console... ?

tbaranes:
I'm closing this now since we can't help more right now. @ysfox Don't hesitate to reopen it if you still having an issue or any questions about that issue!

nico75005:
I have the same error, even after building the IBAnimatable Framework.

Here are the logs:

```
Playground execution failed: error: /var/folders/7z/jhrk21_sdaff773sc0000gn/T/./lldb/1356/playground25.swift:3:8: error: no such module 'IBAnimatable_Sources'
import IBAnimatable_Sources
       ^


* thread #1: tid = 0x11b012, 0x00000001089f13c0 Predefined Animations`executePlayground, queue = 'com.apple.main-thread', stop reason = breakpoint 1.2
  * frame #0: 0x00000001089f13c0 Predefined Animations`executePlayground
    frame #1: 0x00000001089f09c0 Predefined Animations`__37-[XCPAppDelegate enqueueRunLoopBlock]_block_invoke + 32
[...]

PR #391, opened 2017-01-18
Fix for 'No such module IBAnimatable' in playground

emcphersonburke:
Fix for issue #251 
Users may need to do a clean & build and/or close and open the project again after applying this fix in order for Playgrounds to behave itself.


tbaranes (approved):

JakeLin (approved):
Thank you, also updated CHANGLOG as well 👍

PR #398, opened 2017-02-01
Reset borders

### LINKED ISSUE #397: borderWidth set to 0 won't remove the current border
Steps to reproduce:
- Set a valid border
- Set borderWidth to 0 in order to remove that border
- Expected: no more border, current result: the border is staying there

That issue is due to the following guard in `BorderDesignable`:

```
    guard borderColor != nil, borderWidth > 0 else {
      return
    }
```

I'm not sure what is the right fix here: should we just remove the `borderWidth` check, or reseting the layers when we are in this case?

For now, a workaround is simply to set the `borderColor` to `.clear` instead of changing the `borderWidth`.

tbaranes:
Close #397


JakeLin (approved):
As quick as a thunder ⛈

tbaranes:
That's because that one was annoying me 😬

JakeLin:
@tbaranes here we go. It is gone 😅

Issue #401, opened 2017-02-21
PresentationConfiguration has no public initialisers.


mikhailmulyar:
At the moment it is not possible to use PresentationPresenter for custom UIViewController, not for AnimatableModalViewController. The problem is first in var presentationConfiguration of PresentationPresenter, which is not public and is not assignable, but required to be non nil if you want to use PresentationPresenter. Second thing that PresentationConfiguration has no public initialisers, so it is not possible to create it outside of the module and assign it to var presentationConfiguration of PresentationPresenter

tbaranes:
That's normal, if you want to use a `PresentationConfiguration` directly, you have to pass by `PresentationPresenterManager`'s `retrievePresenter` which will give you a valid since it will handle a cache for you, 

Side notes: as talked in #363, even if we let access to these classes, we don't encourage you to use them. If possible, you should directly use `PresentationPresentedViewController` to make your custom presentation.

mikhailmulyar:
As far as i can see `PresentationPresenterManager` creates `PresentationPresenter` with `presentationConfiguration = nil`. So it is not usable as `UIViewControllerTransitioningDelegate` until you assign some `presentationConfiguration`. `AnimatableModalViewController` assigns presentationConfiguration in its `setupPresenter` method which is internal.

tbaranes:
My bad, I misread your question. I will need more input about your context. Why don't you want to use `AnimatableModalViewController` directly? Can you give us a concrete example of how you will use this flow if we add a public initialiser to `PresentationConfiguration`?

mikhailmulyar:
I have view controller from another framework, so i can not subclass AnimatableModalViewController. But i want to use PresentationPresenter features to allow presentation features of it in my view controller.

tbaranes:
I understand now, that's make sense! I ended opening the project and realise that's an implementation error. I mixed up the usage of the presentation feature 😆

Just pushed #402, if you want to give it a try, it should solved your issue.

mikhailmulyar:
`var presentationConfiguration: PresentationConfiguration?` in `PresentationPresenter` should be also public to be able to set it

JakeLin:
@mikhailmulyar thanks for reporting that. <del>We</del> I always think more about how to use `IBAnimatable` in *Interface Builder*, not programmatically, especially how to customize UI elements. 

@tbaranes thanks for responding that quickly, you make the project better and better 👍.

PR #402, opened 2017-02-21
Add public intializer to PresentationDesignable

tbaranes:
Follow up of #401 
Since the philosophy of `IBAnimatable` is to use protocols for everything in order to reuse them in our own components, it's quite sad that we are blocking that usage for `PresentationDesignable`. 

I think we just missed that case, and  it's probably an implementation error.

### DIFF
--- a/IBAnimatable/PresentationDesignable.swift
+++ b/IBAnimatable/PresentationDesignable.swift
@@ -60,6 +60,8 @@ public protocol PresentationDesignable: class {
 
 }
 
+// MARK: - Configuration
+
 public extension PresentationDesignable where Self: UIViewController {
 
   public func configurePresenter() {
@@ -88,7 +90,16 @@ public extension PresentationDesignable where Self: UIViewController {
     presenter?.presentationConfiguration = presentationConfiguration
   }
 
+  public func configureDismissalTransition() {
+    let animationType = dismissalAnimationType
+    if let dismissalSystemTransition = animationType.systemTransition {
+      modalTransitionStyle = dismissalSystemTransition
+    }
+  }
+
 }
+
+// MARK: - PresentationConfiguration
 
 /// `PresentationConfiguration` a struct is used for specifying the dimming view and modal view for `AnimatablePresentationController`
 public struct PresentationConfiguration {
--- a/IBAnimatable/PresentationDesignable.swift
+++ b/IBAnimatable/PresentationDesignable.swift
@@ -9,6 +9,8 @@ public typealias ModalSize = (width: PresentationModalSize, height: Presentation
 /// PresentationDesignable is a protocol to define customised modal view controller which is used as the `presentedViewController` for `UIPresentationController`
 public protocol PresentationDesignable: class {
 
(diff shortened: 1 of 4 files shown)


tbaranes:
@mikhailmulyar Let's continue the discussion here.

I changed a bit my mind, instead of putting everything public... I prefer abstract the `setupPresenter` in the protocol, which means `PresentationDesignable` is now usable from anywhere. `AnimatableModalViewController` doesn't have anymore any private properties... used to do a presentation.

Can you give it a try and let me know if that's ok for you?

tbaranes:
I'm even wondering if it would be worth to add a `viewWillAppear` to the protocol in order to abstract the dismissal animation setup in order to avoid bug cases.

mikhailmulyar:
Looks like it will work. Gonna check it soon

mikhailmulyar:
Still not able to fully use it since those errors if i'm trying to adopt `PresentationDesignable` as in `AnimatableModalViewController`.
Error: 'transitionDuration' is inaccessible due to 'internal' protection level
Error: 'presentationConfiguration' is inaccessible due to 'internal' protection level

tbaranes:
This time, it compiles! I test it to make sure it was ok 😆

mikhailmulyar:
Yes, now it works. Thanks!

IBAnimatable/PresentationDesignable.swift
```diff
@@ -58,19 +60,62 @@ public protocol PresentationDesignable: class {
 
 }
 
+// MARK: - Configuration
+
+public extension PresentationDesignable where Self: UIViewController {
+
+  public func configurePresenter() {
```
JakeLin, line 67: Great idea for having an extension method 👍

IBAnimatable/PresentationDesignable.swift
```diff
@@ -58,19 +60,62 @@ public protocol PresentationDesignable: class {
...
+  public var blurOpacity: CGFloat = .nan
+  public var shadowColor: UIColor?
+  public var shadowRadius: CGFloat = .nan
+  public var shadowOpacity: CGFloat = 0.7
+  public var shadowOffset: CGPoint = .zero
+  public var modalPosition: PresentationModalPosition = .center
+  public var modalSize: (PresentationModalSize, PresentationModalSize) = (.half, .half)
+  public var keyboardTranslation = ModalKeyboardTranslation.none
+
+  public init() {}
```
JakeLin: Do we need that? since it is a `struct`, we may omit it.

JakeLin (approved):
👍  Good one, thanks

tbaranes:
@mikhailmulyar The next release will include that feature, until then, you can target `master` directly :)

mikhailmulyar:
Ok. Thanks!

JakeLin:
Hey, @mikhailmulyar release [3.1.3](https://github.com/IBAnimatable/IBAnimatable/releases/tag/3.1.3) is up, you don't have to use `master` for the Cocoapods now, thanks.

Issue #406, opened 2017-02-26
MaskDesignable removes custom mask by default


DanielAsher:
Hi,
thanks again for a great framework!
I require using a gradient mask on a subclass of AnimatableView. I didn't find a way to do this from the designer - so I created a `layer.mask` in my derived class. However, `MaskDesignable` removes the layer mask by default (i.e. when `.none`) in [this](https://github.com/IBAnimatable/IBAnimatable/blob/master/IBAnimatable/MaskDesignable.swift#L53) line.
I guess I'm still learning the philosophy of this library. I love being able to allow designers access to powerful features of UIKit, but - at first blush - this removes the ability to robustly use programmatic animation APIs.
best of wishes and all power to you!

tbaranes:
Thanks for that report! It's true that the current implementation of `MaskDesignable` doesn't support any custom mask. It's a bit restrictive but we are in a complex context since we have to remove the mask if you choose to remove the mask later.

From what you are saying, I see three solutions:
- Use `GradientDesignable` directly to make your gradient. If it's not working let's debug it to make it compatible with `MaskDesignable` 😬 (Can't test it right now)
- Add a `custom` option to `MaskType` which will prevent removing a mask that is not implemented  by IBAnimatable. That could be a nice feature to have if it's been well thought. 
- Make your own `AnimatableView` to conform only designable protocols that you need.

Don't hesitate to make a PR if you decide to go with option 1 or 2!

About the philosophy ans implementation of the framework: it has been fully built using protocols in order to make the most flexible possible. All the Animatable* classes are just some wrappers of these protocols to make it easier to use. It's true that it has been more thought for the interface builder than programmatically usage but, I think, the usage should be working as expected for both. If that's not case, it's definitely an issue! Don't hesitate to open another issue or PR if you have any suggestions to improve it.

Hope that helps :)

DanielAsher:
@tbaranes thanks for your super-quick responses.
I certainly agree about the complexity of the context! I also had similar ideas to all three of your suggestions. To your points in order:
* This doesn't seem orthogonal as a designer may require a fill gradient as well as a mask gradient
* Perhaps a `gradient` option to `MaskType` would handle this scenario, though full details of the mask gradient could clutter the IB property inspector. A simpler alternative to fix the issue would be a flag that tracks whether `MaskType` has be set to anything other than `.none` before removing the layer mask. 
[...]

PR #407, opened 2017-02-28
Add previousMaskType to MaskDesignable to solve `MaskDesignable remov…

DanielAsher:
…es custom mask by default #406`

### DIFF
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -146,13 +146,10 @@ open class AnimatableStackView: UIStackView, CornerDesignable, FillDesignable, B
   // MARK: - MaskDesignable
   open var maskType: MaskType = .none {
     didSet {
-      previousMaskType = oldValue
-      configureMask()
+      configureMask(previousMaskType: oldValue)
       configureBorder()
     }
   }
-
-  open var previousMaskType: MaskType = .none
 
   /// The mask type used in Interface Builder. **Should not** use this property in code.
   @IBInspectable var _maskType: String? {
@@ -199,7 +196,7 @@ open class AnimatableStackView: UIStackView, CornerDesignable, FillDesignable, B
   }
 
   fileprivate func configureAfterLayoutSubviews() {
-    configureMask()
+    configureMask(previousMaskType: maskType)
     configureCornerRadius()
     configureBorder()
     configureGradient()
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -145,11 +145,17 @@ open class AnimatableStackView: UIStackView, CornerDesignable, FillDesignable, B
 
   // MARK: - MaskDesignable
   open var maskType: MaskType = .none {
+    willSet {
+      previousMaskType = newValue
+    }
+
     didSet {
       configureMask()
       configureBorder()
     }
   }
+
+  open var previousMaskType: MaskType? = nil
 
   /// The mask type used in Interface Builder. **Should not** use this property in code.
   @IBInspectable var _maskType: String? {
--- a/IBAnimatable/AnimatableStackView.swift
+++ b/IBAnimatable/AnimatableStackView.swift
@@ -145,17 +145,14 @@ open class AnimatableStackView: UIStackView, CornerDesignable, FillDesignable, B
 
   // MARK: - MaskDesignable
   open var maskType: MaskType = .none {
-    willSet {
-      previousMaskType = newValue
-    }
-
     didSet {
+      previousMaskType = oldValue
       configureMask()
       configureBorder()
     }
   }
 
-  open var previousMaskType: MaskType? = nil
+  open var previousMaskType: MaskType = .none
 
   /// The mask type used in Interface Builder. **Should not** use this property in code.
   @IBInspectable var _maskType: String? {
--- a/IBAnimatable/MaskDesignable.swift
+++ b/IBAnimatable/MaskDesignable.swift
@@ -31,6 +31,7 @@ public protocol MaskDesignable {
    ```
    */
   var maskType: MaskType { get set }
+  var previousMaskType: MaskType? { get set }
 }
 
 public extension MaskDesignable where Self: UIView {
@@ -50,7 +51,9 @@ public extension MaskDesignable where Self: UIView {
     case .triangle:
       maskTriangle()
     case .none:
-      layer.mask?.removeFromSuperlayer()
+      if previousMaskType != nil {
+        layer.mask?.removeFromSuperlayer()
+      }
     }
   }
 }
--- a/IBAnimatable/MaskDesignable.swift
+++ b/IBAnimatable/MaskDesignable.swift
@@ -31,7 +31,7 @@ public protocol MaskDesignable {
    ```
    */
   var maskType: MaskType { get set }
-  var previousMaskType: MaskType? { get set }
+  var previousMaskType: MaskType { get set }
 }
 
 public extension MaskDesignable where Self: UIView {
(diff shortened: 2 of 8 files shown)


IBAnimatable/MaskDesignable.swift
```diff
@@ -50,7 +51,9 @@ public extension MaskDesignable where Self: UIView {
     case .triangle:
       maskTriangle()
     case .none:
-      layer.mask?.removeFromSuperlayer()
+      if previousMaskType != nil {
+        layer.mask?.removeFromSuperlayer()
```
tbaranes: Just wondering, don't you need to set `previousMaskType` to nil once the layer is removed?

tbaranes (changes requested):
Just a small question, otherwise it looks good 👍 
Also, can you add an entry to the CHANGELOG?

IBAnimatable/MaskDesignable.swift
DanielAsher: I wanted to keep the logic as simple (and understandable) as possible - so any use of `MaskType` (after initialization to `.none`) would force a `layer.mask?.removeFromSuperlayer()`. However, my intuition suggests you are right, and resetting to `.none` could reenable custom gradients.
I'll go ahead and reset previousMaskType.

DanielAsher: However... this would require either:
1) a class type protocol `public protocol MaskDesignable : class { ... }`
or
2) a mutating function `public extension MaskDesignable where Self: UIView {
  public func mutating configureMask() { ... }`

DanielAsher: Let me know if either of these options are acceptable.

DanielAsher: Another alternative (I believe) would be to have `previousMaskType : MaskType` initialized to `.none` and:
```
      if case .none = previousMaskType {
        return
      } else {
        layer.mask?.removeFromSuperlayer()
      }
```

DanielAsher:
I've removed the optional from `previousMaskType`, initialized it to `.none` and:
```
    didSet {
      previousMaskType = oldValue
...
```
in client classes.

DanielAsher:
Added Change log.

IBAnimatable/MaskDesignable.swift
```diff
@@ -31,6 +31,7 @@ public protocol MaskDesignable {
    ```
    */
   var maskType: MaskType { get set }
+  var previousMaskType: MaskType { get set }
```
JakeLin: If we still need it. It will be better if we have some documentation comments here to explain what is the property for. e.g. 
```
/// `previousMaskType` is a property to cache the previous mask type. If it is not `.none`, then we will remove `layer.mask` before re-adding it.
```
Feel free to write a better comment for that 😉

IBAnimatable/AnimatableStackView.swift
```diff
@@ -146,11 +146,14 @@ open class AnimatableStackView: UIStackView, CornerDesignable, FillDesignable, B
   // MARK: - MaskDesignable
   open var maskType: MaskType = .none {
     didSet {
+      previousMaskType = oldValue
```
JakeLin: Shall we put this line in `configureMask(previousMaskType: oldValue)`. Then we don't copy & paste it in all `Animatable*` classes. But it will have a problem of introducing breaking API for `configureMask()`. One workaround I can think of is to use optional and default value (`nil`) like

```
public func configureMask(previousMaskType: MaskType? = nil)
```

Then we don't even need `var previousMaskType: MaskType { get set }` any more.

JakeLin (approved):
I am happy with the current implementation. Also please have a look at the comments and see which solution is better from your point of view.

JakeLin:
@DanielAsher thank you very much for bringing this use case and putting up as a PR. I have a few of comments and please have a look. I am happy to discuss it if you have any questions.

@tbaranes thanks for guiding the conversion, I am sorry I didn't involve into #406 . But I think you guys did a great job without me 😅. Can you also have a look at my comments?

IBAnimatable/AnimatableStackView.swift
DanielAsher: this is a good idea. I'll try to get to it later today. Thanks.

tbaranes: That's smart 👍

DanielAsher:
added comment to `previousMaskType`

JakeLin:
@DanielAsher have you got a chance to try out `public func configureMask(previousMaskType: MaskType? = nil)` ? How did it look like? If you haven't, no worries, take your time.

DanielAsher:
@JakeLin I did by starting a review (though maybe I don't understand the Github review mechanism well enough.)

I'll copy my comments here:

I started to implement this, but found `configureMask` being called in `configureAfterLayoutSubviews` - so it looks like an ivar is necessary. I also started to feel that passing a nil as default could be unexpected on the client-side. I'm sure this could be achieved to reduce the replication, however, I'm not confident that I understand the ramifications. Although more verbose, the current solution is explicit and robust.

tbaranes:
When `configureAfterLayoutSubviews` is called, the `previousType` will always be equal to the current `maskType`. The previous differ only when we set a new value to `maskType`.

There's another advantage at this solution: it doesn't introduce any breaking changes. Adding a new ivar to the protocol will break any classes conforming to `MaskDesignable`. However, I'm joining you about passing nil, that's hardly understable for the client side, so... both have pro and cons for me 😬

JakeLin:
@DanielAsher thanks for investigating this issue. You are right, I didn't consider calling `configureMask` in `configureAfterLayoutSubviews`. As @tbaranes mentioned above, it may be fixed by passing current maskType like `configureMask(previousMaskType: maskType)`.

Actually, I don't mind either solution. But @tbaranes raised a good point 
> Adding a new ivar to the protocol will break any classes conforming to MaskDesignable

I think we try to not introduce breaking changes during minor version update.

For the default value `nil` for `public func configureMask(previousMaskType: MaskType? = nil)`. I don't like it either. It was a workaround I thought out to avoid breaking changes. I prefer to not have the default value and force the client code to specify `previousMaskType` every time. If this solution works, I am going to add an github issue to remove the default value for this method for version / release 4. If this solution doesn't work, let's merge the `var previousMaskType: MaskType { get set }` solution.

How do you guys think?

tbaranes:
Sounds good to me. Let's not introduce breaking change for this minor version, but make that breaking change for the next major version.

tbaranes:
@DanielAsher In the end, we are gonna to release a 4.0.0 once all the current PRs are merged, so we can go forward and make a breaking change by breaking the current classes conforming to `MaskDesignable`!

DanielAsher:
I must be blessed with good timing! Thanks :)

JakeLin:
@DanielAsher As discussed before,  I would like to try out the solution `public func configureMask(previousMaskType: MaskType)`. Please see commit e89992d. Now we introduce a breaking change to `func configureMask()` since there is no default value for `previousMaskType`. I think we all agree that we should not have a default value to `nil`. But as @tbaranes, we are going to release version 4 soon. Which will include this PR. We can introduce breaking changes 😂 

The reason I like `public func configureMask(previousMaskType: MaskType)` more is the conforming classes don't need to implement `var previousMaskType: MaskType` property. And this property is used for cache only. 

But I do concern whether this solution fits your requirement or not. If you have time, can you try it out with your custom mask? Please let me know it is working or not, thanks.

tbaranes (approved):

JakeLin:
Hi @tbaranes if @DanielAsher says yes, I would like to merge this PR. It is completed I think.

DanielAsher:
@JakeLin @tbaranes. I've released a TestFlight version including this modification with no UAT problems and believe we should merge the PR. Thanks for your contributions to this great process.

tbaranes:
Thanks for your contribution @DanielAsher !

JakeLin:
Thanks @DanielAsher Go test✈️ 🎉

JakeLin:
@DanielAsher you may be interested in @lastMove 's nice feature #465, please have a look.

Issue #394, opened 2017-01-20
CornerSides is case sensitive


tbaranes:
For example:`TopLeft` in interface builder won't work.

tbaranes:
That's also the case for `BorderSide`.

JakeLin:
@tbaranes interesting, we do call `lowercased()` like `let sideElements = rawValue.lowercased().characters.split(separator: ",")` in `BorderSides`

mmadjer:
@tbaranes, I double checked the border sides and seems to be fine. Any special case you checked?

tbaranes:
Maybe I've been faked by Xcode update. Let's consider it as a bad report, I will open a new issue if I reproduce it, sorry! 🤕

mmadjer:
no worries @tbaranes. That's how XCode functions..or doesn't function. ;o)

PR #408, opened 2017-02-28
Corner sides case sensitive fix

mmadjer:
Make corner sides case insensitive.
See issue #394.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ None
 [#368](https://github.com/IBAnimatable/IBAnimatable/pull/403)
 
 #### Bugfixes
-None
+- Make corner sides case insensitive.
+[#394](https://github.com/IBAnimatable/IBAnimatable/issues/394)
 
 ### [3.1.3](https://github.com/IBAnimatable/IBAnimatable/releases/tag/3.1.3)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ None
 [#368](https://github.com/IBAnimatable/IBAnimatable/pull/403)
 
 #### Bugfixes
-None
+- Make corner sides case insensitive.
+[#394](https://github.com/IBAnimatable/IBAnimatable/issues/394)
```
JakeLin: Could you please add your name here 😊

IBAnimatable/CornerSide.swift
```diff
@@ -57,10 +57,10 @@ public struct CornerSides: OptionSet {
     }
 
     switch side {
-    case .topLeft: self = .topLeft
-    case .topRight: self = .topRight
-    case .bottomLeft: self = .bottomLeft
-    case .bottomRight: self = .bottomRight
+    case .topleft: self = .topleft
```
JakeLin, line 60: can we do `self = side` here without the `switch`?

JakeLin (approved):
Thanks, it is ready to go 👍

IBAnimatable/CornerSide.swift
mmadjer: good catch. will change it.

mmadjer: no we can't, as it maps to the CornerSides option types.

tbaranes (approved):

IBAnimatable/CornerSide.swift
JakeLin: @mmadjer Good found, I got confused 😕

JakeLin:
@mmadjer thanks 👍👍👍

tbaranes:
Second thought: are we introducing a breaking changes by changing the `CornerSide` rawValue ?

JakeLin:
@tbaranes I think we did  😢，good pick-up. It will be no problem when we use it in IB. But need to change when we use it in code. Since we have merged it already, I am going to update the CHANGELOG.md now.

Update: commit 2f05c225345f4452f3c3245e32c374ac238cdcb4 pushed to describe the breaking change.

tbaranes:
Can't we just fix it by giving a value to the enum?

```
case topLeft: "topleft"
...
```

mmadjer:
Ah sh**, good thinking @tbaranes.

JakeLin:
@tbaranes well, such a good idea. 

@mmadjer Could you please give it a go if you have time?

PR #409, opened 2017-03-03
Fix corner breaking change

tbaranes:
Follow up of #408 

Also introducing a new breaking change, a leftover from the swift3 migration: `.AllSides` -> `.allSides`.

I think that one is ok since it's already the default value, users can just remove the set for most of the cases by setting a simple `cornerRadius`.

Does that seem ok to you? Should I revert that commit?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Next
 
 #### API breaking changes
-- Change the rawValue for `CornerSide` enum. Since we fix an issue to make corner sides case insensitive [#394]. If we use `CornerSide` in code, we need to change `.topLeft` to `.topleft`, `.topRight` to `.topright`, `.bottomLeft` to `.bottomleft` and `.bottomRight` to `.bottomright`. If we use `CornerSide` in Interface Builder, there is not breaking change for it.
+- `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programatically a cornerSide to your view, you will just have to lowercase the A.
 
 #### Enhancements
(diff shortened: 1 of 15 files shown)


mmadjer (approved):
awesome @tbaranes. Still a breaking change but much less impact as the pervious approach.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Next
 
 #### API breaking changes
-- Change the rawValue for `CornerSide` enum. Since we fix an issue to make corner sides case insensitive [#394]. If we use `CornerSide` in code, we need to change `.topLeft` to `.topleft`, `.topRight` to `.topright`, `.bottomLeft` to `.bottomleft` and `.bottomRight` to `.bottomright`. If we use `CornerSide` in Interface Builder, there is not breaking change for it.
+- `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programatically a cornerSide to your view, you will just have to lowercase the A.
```
JakeLin, line 9: Right, we missed that 😢. I think most of the client code would not use `.allSides` since it is the default. But still need to change for all classes conforming to `CornerDesignable`.

IBAnimatable/CornerSide.swift
```diff
@@ -9,31 +9,31 @@
 import Foundation
 
 public enum CornerSide: String {
-  case topleft
-  case topright
-  case bottomleft
-  case bottomright
+  case topLeft = "topleft"
```
JakeLin, line 12: Good idea 👍

JakeLin (approved):
Thanks for such a quick fix 👍👍👍

Issue #238, opened 2016-07-07
IBAnimatable App 


lastMove:
**Demo for how to use in Interface Builder**
- [x] All supported Animations with default parameters
- [x] BlurEffectStyle
- [x] BorderSide
- [x] ColorType
- [x] GradientStartPoint
- [x] GradientType
- [x] Refresh Control (`RefreshTableViewController` already ready to be used, just need to add an entry in the storyboard)
- [ ] BackgroundImageDesignable (#416)
- [ ] TextField: placeholders, images...

**Demo for how to use in code**
- [ ] Configurable Animation parameters. e.g. changing type, duration, damping 
- [ ] Configurable UI element. e.g. changing cornerRadius

------- Original post -----
Companion App 

We could maybe add a “companion” App like [Spring](https://github.com/MengTo/Spring) ’s one. 
It allows to show all Masks, Animations, Transitions, Gradients etc doable with IBAnimatable. 
Moreover it’s also a Support for testing those animations / transitions etc directly on-device.
Users could see the code that generate it.

Advantages : 
Demo the animations, gradients, mask, transitions, 
Help to show in context
It could help to choose an animation while editing
It could simplify regression detection since we can use Xcode UITests
We could even put it on the app store ;) 

Disadvantages or alternative : 
One more thing to keep updated
It maybe, at least partially, serves the same purpose than the Playground support

How do you think about it ?

tbaranes:
👍 Love this idea! It's joining what @JakeLin starts  with the transition demo (hidden in "forgot password" ). Moreover, a technical demonstration will be a nice way to avoid some issues about how to use our `Animatable` components. Just few questions:
- Should we add this a another project / demo of IBAnimatable, or put another custom access in the current demonstration? I prefer the first option since it will be more accessible.
- IBAnimatable has a lot of features, how can we group them all while staying readable and easy to understand? `Spring` demo is really nice but focus only on animations. Just a list, sublists... ?
- Is that new demo will be focusing on how to use `IBAnimatable` with interface builder, programatically, or both? The vision of `IBAnimatable` is more focused on interface builder, but that can be a nice way to open have some quick examples of how to use it programatically (by taking  "real code" `Spring` panel)

> It maybe, at least partially, serves the same purpose than the Playground support

That's a good point, but there's one big difference between both: playground can be used to play quickly with the library while a sample like `Spring` brings code ready to used and concrete example. Moreover, we definitely can't use playground to cover all the features... :(

[...]

Issue #53, opened 2016-01-26
AnimatableCollectionView and AnimatableCollectionViewCell


gkye:
An AnimatableCollectionView and/or an AnimatableCollectionViewCell would be nice :+1:

JakeLin:
For basic features, it is straight forward like `AnimatableTableView` and `AnimatableTableViewCell`, For more complicated features like slide in all cells one by one, it may require to do something with `UICollectionViewDelegateFlowLayout` to customise the animation. Any ideas?

tbaranes:
A few interesting UICollectionView layout animations: [AnimatedCollectionViewLayout](https://github.com/KelvinJin/AnimatedCollectionViewLayout?utm_campaign=This%2BWeek%2Bin%2BSwift&utm_medium=web&utm_source=This_Week_in_Swift_121)

JakeLin:
@tbaranes those are like our transitions.

tbaranes:
Yeah, but they are collectionView transitions 😆
I don't know well the API, but we may be able to reuse the viewControllers animations for some cases.

JakeLin:
In our case, we conform to `UIViewControllerAnimatedTransitioning` and implement ` func animateTransition(using transitionContext: UIViewControllerContextTransitioning)` . In collection trainsitons, we need to subclass `UICollectionViewFlowLayout` and override `layoutAttributesForElements`. All animations are based on `attributes`. We need to abstract how collection view attributes to transition `fromVC`, `toVC`. `formView` and `toView`. I don't think that's easy to do  😣 The design of UIKit doesn't consider to share animations to different scenarios. Even for transition and presentation animations, they are very similar but hard to share the code for similar effect 😥

PR #416, opened 2017-03-25
Allow to set a background image to table view and cells views

phimage:
i.e. view with `backgroundView` attribute
Add also `BlurDesignable` to table view to blur the `backgroundView`

- A lot of time I need to put a background image on this ui components without adding a specific `UIImageView`. With the same principle of `BlurDesignable` a view is added. But this time not in subviews hierarchy, with crapping things like scroll view, etc... but the special attribute  `backgroundView`

- to see background of a table, each cell view must be configured with clear color (or some transparency )

- `UICollectionView` could also conform to `BackgroundImageDesignable` but there is no `AnimatableCollectionView` yet, any reason? 

- I must edit the change log on my own? could be weird with conflicts when there is multiple PR

### DIFF
--- a/IBAnimatable/BackgroundImageDesignable.swift
+++ b/IBAnimatable/BackgroundImageDesignable.swift
@@ -1,0 +1,61 @@
+//
+//  Created by phimage on 25/03/2017.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import Foundation
+
+import UIKit
+
+/// Protocol for designing background image
+public protocol BackgroundImageDesignable {
+
+  /**
+   * The background image
+   */
+  var backgroundImage: UIImage? { get set }
+
+}
+
+/// Protocol for designing background
+public protocol BackgroundDesignable: class {
+
+  /**
+   * The background view
+   */
+  var backgroundView: UIView? { get set }
+
+}
+
+extension UITableViewCell: BackgroundDesignable {}
+extension UITableViewHeaderFooterView: BackgroundDesignable {}
+extension UITableView: BackgroundDesignable {}
+extension UICollectionViewCell: BackgroundDesignable {}
+extension UICollectionView: BackgroundDesignable {}
+
+fileprivate typealias BackgroundImageView = AnimatableImageView
+
+public extension BackgroundImageDesignable where Self: BackgroundDesignable {
+  public func configureBackgroundImage() {
+    if let image = backgroundImage {
+      if let imageView = self.backgroundView as? AnimatableImageView {
+        imageView.image = image
+      } else {
+        backgroundView = BackgroundImageView(image: image)
+      }
+    } else {
+      if self.backgroundView is BackgroundImageView {
+        backgroundView = nil
+      }
+    }
+  }
+
+  public var backgroundImageView: UIImageView? {
+    get {
+      return self.backgroundView as? UIImageView
+    }
+    set {
+      self.backgroundView = newValue
+    }
+  }
+}
--- a/IBAnimatable/BlurDesignable.swift
+++ b/IBAnimatable/BlurDesignable.swift
@@ -28,11 +28,14 @@ public extension BlurDesignable where Self: UIView {
    configureBlurEffectStyle method, should be called in layoutSubviews() method
    */
   public func configureBlurEffectStyle() {
+    configureBlurEffectStyle(self)
+  }
+  public func configureBlurEffectStyle(_ blurableView: UIView) {
     // Used for caching the previous visual effect view
     var privateVisualEffectView: PrivateVisualEffectView?
 
     // Remove the existing visual effect view
-    subviews.flatMap { $0 as? PrivateVisualEffectView }
+    blurableView.subviews.flatMap { $0 as? PrivateVisualEffectView }
       .forEach {
         privateVisualEffectView = $0 // Cache it for the subviews
         $0.removeFromSuperview()
@@ -48,7 +51,7 @@ public extension BlurDesignable where Self: UIView {
     if let vibrancyEffectStyle = vibrancyEffectStyle {
       let blurEffectStyleForVibrancy = UIBlurEffect(style: vibrancyEffectStyle)
       let vibrancyEffectView = makeVisualEffectView(effect: UIVibrancyEffect(blurEffect: blurEffectStyleForVibrancy))
-      subviews.forEach {
+      blurableView.subviews.forEach {
         vibrancyEffectView.contentView.addSubview($0)
       }
 
@@ -60,11 +63,11 @@ public extension BlurDesignable where Self: UIView {
     } else {
       // If privateVisualEffectView is not `nil`, re-add them back to the subviews of original UI element.
       privateVisualEffectView?.contentView.subviews.forEach {
-        addSubview($0)
+        blurableView.addSubview($0)
       }
     }
 
-    insertSubview(blurEffectView, at: 0)
+    blurableView.insertSubview(blurEffectView, at: 0)
   }
 }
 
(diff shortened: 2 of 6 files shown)


IBAnimatable/BackgroundImageDesignable.swift
```diff
@@ -0,0 +1,61 @@
...
+extension UITableView: BackgroundDesignable {}
+extension UICollectionViewCell: BackgroundDesignable {}
+extension UICollectionView: BackgroundDesignable {}
+
+fileprivate typealias BackgroundImageView = AnimatableImageView
+
+public extension BackgroundImageDesignable where Self: BackgroundDesignable {
+  public func configureBackgroundImage() {
+    if let image = backgroundImage {
+      if let imageView = self.backgroundView as? AnimatableImageView {
```
tbaranes: That `self.` isn't mandatory, is it? As the others below

IBAnimatable/BackgroundImageDesignable.swift
```diff
@@ -0,0 +1,61 @@
...
+
+}
+
+extension UITableViewCell: BackgroundDesignable {}
+extension UITableViewHeaderFooterView: BackgroundDesignable {}
+extension UITableView: BackgroundDesignable {}
+extension UICollectionViewCell: BackgroundDesignable {}
+extension UICollectionView: BackgroundDesignable {}
+
+fileprivate typealias BackgroundImageView = AnimatableImageView
```
tbaranes: Not sure about that typealias, I feel that bring more confusion than using directly `AnimatableImageView`.

IBAnimatable/BlurDesignable.swift
```diff
@@ -28,11 +28,14 @@ public extension BlurDesignable where Self: UIView {
    configureBlurEffectStyle method, should be called in layoutSubviews() method
    */
   public func configureBlurEffectStyle() {
+    configureBlurEffectStyle(self)
+  }
+  public func configureBlurEffectStyle(_ blurableView: UIView) {
```
tbaranes: newLine missing

tbaranes (changes requested):
I made a first quick review of this, just a few small comments, if you can take a look, I will make a second review after that.

There's also the CHANGELOG and documentation entry missing.

tbaranes:
@phimage Thanks for doing this! It's true that a common behavior in iOS development.

> UICollectionView could also conform to BackgroundImageDesignable but there is no AnimatableCollectionView yet, any reason?

Just needed someone to ask for it 😬 
There's no reason to not have it, just no one take the time to do it or need it until now.

> I must edit the change log on my own? could be weird with conflicts when there is multiple PR>

Yeah, that's annoying part with the CHANGELOGs

IBAnimatable/BackgroundImageDesignable.swift
phimage: I am thinking about doing something like 
```swift
class PrivateImageView: AnimatableImageView {}
```
so making typealias allow to change quickly

but I can remote it

tbaranes: I don't see advantages to not use AnimatableImageView directly here, is there a reason?

IBAnimatable/BackgroundImageDesignable.swift
phimage: self is not mandatory I think (there is no closure)
I put `self` when possible to avoid some issue when renaming stuff and local variable become class variable etc...  
But there is poor or con, and swift lint do not have this rule yet
https://github.com/realm/SwiftLint/issues/321

Like I do in previous PR I will remove it

IBAnimatable/BackgroundImageDesignable.swift
JakeLin: agree with @tbaranes for using `AnimatableImageView` instead of `typealias BackgroundImageView`. 

On the other hand for the private subclass. I think creating a private subclass is good for some case I can think of. For example, we can check like `backgroundView as? PrivateImageView` then we know `backgroundView` must be set within `extension BackgroundImageDesignable`. Another someone may use `AnimatableImageView` for `backgroundView` outside of `IBAnimatable` framework. And we should not  reset `backgroundView = nil` if `backgroundView` is not original set by `extension BackgroundImageDesignable`.

Is it the case you are thinking of, @phimage ? I leave it to you to make a decision to use `private class PrivateAnimatableImageView: AnimatableImageView` or not.

IBAnimatable/BackgroundImageDesignable.swift
JakeLin: @phimage It is a controversial topic for keeping `self` or not. We try to omit `self` if possible for our repo, can you also follow that?

IBAnimatable/BlurDesignable.swift
JakeLin: shall we use `public func configureBlurEffectStyle(for blurableView: UIView)`? using `for` instead of `_` there.

IBAnimatable/BackgroundImageDesignable.swift
```diff
@@ -0,0 +1,61 @@
...
+   */
+  var backgroundView: UIView? { get set }
+
+}
+
+extension UITableViewCell: BackgroundDesignable {}
+extension UITableViewHeaderFooterView: BackgroundDesignable {}
+extension UITableView: BackgroundDesignable {}
+extension UICollectionViewCell: BackgroundDesignable {}
+extension UICollectionView: BackgroundDesignable {}
```
JakeLin, line 32: It took me a while to understand this, I found out is a great idea for having a `protocol BackgroundDesignable` Because we can't use protocol extension for conforming different classes like `public extension BackgroundImageDesignable where Self: UITableViewCell or UITableViewHeaderFooterView or UITableView or UICollectionViewCell or UICollectionView`. To have `BackgroundDesignable` and make the code looks better. Well done 👍

JakeLin (commented):
@phimage well done on this PR, can you please address the comment from @tbaranes , and we should be ready to go after that.

JakeLin:
@phimage thank you very much for a great PR. @tbaranes has answered your questions. I have 2 cents for them.

We do need `AnimatableCollectionView` and `AnimatableCollectionViewCell`, we can see there is a card/issue for that https://github.com/IBAnimatable/IBAnimatable/issues/53. But as @tbaranes said, we didn't do it because no one asked for. It will be great if you can offer some hand on it. `BackgroundImageDesignable` is great for `AnimatableCollectionView` and `AnimatableCollectionViewCell` as well.

For the `CHANGELOG.md`, it is annoying, we have to `rebase` every time before we merge into master. But good for the user to check all changes. Please modify it when you submit a PR. 

@phimage please let me know if you would like to offer some small helps on this project and join IBAnimatable GitHub organisation as an internal contributor. We have a Slack channel for all internal contributors as well.  Thanks again for all you contributions, you make `IBAnimatable` better 💪

tbaranes:
Thanks for the update! It's perfect 💯

phimage:
I will make the change after finishing with #418

@JakeLin ok for organisation and Slack
I have some other topics that I want to add to IBAnimatable, and some like buttons touch animation must be discussed

JakeLin:
@phimage welcome onboard. I have sent you an invite to be an internal contributor of the IBAnimatable organisation. Our rule is simple: There is no pressure to work on the project. Everyone has the same access to the repo. We use feature branch like `feature/animatable-collection-view` or `bugfix/fix-corner-radius` and PR to master. We are open to any opinions. Once we have one or two 👍 , we can merge the PRs. Please shout out if you have any question.

Can you also please send me your Slack account (email)? You can find my email on my GitHub profile page. Thanks

JakeLin (approved):
👍 well done.

JakeLin:
@phimage I merge it now, Travis is failing because of `bundle exec danger`. But I can see the app has been built successfully. Thanks for your PR, well done 👍

tbaranes:
There's a documentation and example app update missing, should we add an issue? Pretty sure we will forget about this otherwise 😬 

I added an entry in #238 for the example app

PR #417, opened 2017-03-25
Make AnimatableSlider inherit from UISlider and add designable images

phimage:
`AnimatableSlider` do not extends `UISlider`, only `UIView`

thumb, minimum and maximum track images added has designable objects in storyboard

PS: 
I try some experiment on slider size by overriding `trackRect()`

```
 @IBInspectable var trackHeight: CGFloat = CGFloat.nan
    
    override func trackRect(forBounds bounds: CGRect) -> CGRect {
       // add here if not trackHeight.isNan
        return CGRect(origin: bounds.origin, size: CGSize(width:bounds.width, height:trackHeight))
    }
 // or
    override func trackRect(forBounds bounds: CGRect) -> CGRect {
       let rect = super.trackRect(forBounds: bounds)
       return ...
    }
```
There is some bug when using maximum and minimum images
to fix some other bugs in storyboard  I use in `trackRect()`
```

#if TARGET_INTERFACE_BUILDER
 // do some stuff to adjust storyboard
#endif
```

### DIFF
--- a/IBAnimatable/SliderImagesDesignable.swift
+++ b/IBAnimatable/SliderImagesDesignable.swift
@@ -42,32 +42,32 @@ public protocol SliderImagesDesignable {
 
 }
 
-extension SliderImagesDesignable where Self: UISlider {
+public extension SliderImagesDesignable where Self: UISlider {
 
   func configureThumbImage() {
-    self.setThumbImage(thumbImage, for: .normal)
+    setThumbImage(thumbImage, for: .normal)
     if let highlighted = thumbHighlightedImage {
-      self.setThumbImage(highlighted, for: .highlighted)
+      setThumbImage(highlighted, for: .highlighted)
     } else {
-      self.setThumbImage(thumbImage, for: .highlighted)
+      setThumbImage(thumbImage, for: .highlighted)
     }
   }
 
   func configureMinimumTrackImage() {
-    self.setMinimumTrackImage(minimumTrackImage, for: .normal)
+    setMinimumTrackImage(minimumTrackImage, for: .normal)
     if let highlighted = minimumTrackHighlightedImage {
-      self.setMinimumTrackImage(highlighted, for: .highlighted)
+      setMinimumTrackImage(highlighted, for: .highlighted)
     } else {
-      self.setMinimumTrackImage(minimumTrackImage, for: .highlighted)
+      setMinimumTrackImage(minimumTrackImage, for: .highlighted)
     }
   }
 
   func configureMaximumTrackImage() {
-    self.setMaximumTrackImage(maximumTrackImage, for: .normal)
+    setMaximumTrackImage(maximumTrackImage, for: .normal)
     if let highlighted = maximumTrackHighlightedImage {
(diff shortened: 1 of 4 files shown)


IBAnimatable/SliderImagesDesignable.swift
```diff
@@ -0,0 +1,74 @@
...
+  var maximumTrackImage: UIImage? { get set }
+
+  /**
+   * The maximum track image when slider state is `highlighted`
+   */
+  var maximumTrackHighlightedImage: UIImage? { get set }
+
+}
+
+extension SliderImagesDesignable where Self: UISlider {
```
JakeLin: I think we need to make it `public` for the others can use those extension methods outside of `IBAnimatable` framework.

IBAnimatable/SliderImagesDesignable.swift
```diff
@@ -0,0 +1,74 @@
...
+
+  func configureThumbImage() {
+    self.setThumbImage(thumbImage, for: .normal)
```
JakeLin: Shall we remove all `self`?

JakeLin (approved):
I like you PR very much with a lot of documentation comments 👍👍👍. I think this PR is ready to go. It will be greate if you can modify the CHANGELOG.md file. It is very simple like https://github.com/IBAnimatable/IBAnimatable/blob/master/CHANGELOG.md#enhancements

One thing I need to confirm with @tbaranes on why did we inherit from `UIView`. If he is OK, we will merge it. Thanks again 💪

JakeLin:
@phimage thank you very much for your PR. Please have a look at my small comments.

@tbaranes In #228, we created `AnimatableSlider`. I know we want to have fully customizable  `AnimatableSlider` and `UISlider` can't give us that 😕. Therefore, we let it inherit from `UIView` instead of `UISlider`. @phimage found out that and made `AnimatableSlider` to inherit from `UISlider`. I think that is better for now since the current `AnimatableSlider` is more like `AnimatableView`. 


If we want to have fully customizable  `AnimatableSlider` in the future, we may create another class like `AnimatableCustomSlider` to inherit from `UIView`. Can I have your opinion on this PR?  thanks.

IBAnimatable/SliderImagesDesignable.swift
tbaranes: +1, also all the others below 😬

tbaranes (approved):

tbaranes:
I agree with you @JakeLin, that PR is a good move since it's making `AnimatableSlider` more _slider_. I don't really remember why we choose `UIView` instead of `UISlider` in #228, but I can't see any reasons. Truth is when we will do the custom sliders, it won't really be related to `AnimatableSlider`, so... who cares?

@phimage Good job on this!

phimage:
Ok for the changes (`public` and `self`)
For changelog I make two lines? one in "api breaking changes" or not?

I will fix also some changelog issue with markdown comma

JakeLin:
@phimage well done. It is in 🎉, thanks a lot. It will be in version 4. @tbaranes thanks for your feedback.

PR #421, opened 2017-03-28
Add `overCurrentContext` to `PresentationDesignable`

### LINKED ISSUE #198: AnimatablePresentationController

tbaranes:
This PR introduces a new way to make custom presentation but also finally close #198 !

What's in?
- A new property in `PresentationDesignable`: `overCurrentContext`.
   - If true, the dimmingView will have the same frame as the controller who started the presentation
   - If false, the dimmingView will take the fullscreen as a `default` presentation

The implementation is a bit tricksy but (almost) working. The internal presentation is still in fullscreen, the main difference is that if we set `overCurrentContext` to true, we will forward all _tap_ which make us able to tap outside.

I have still one issue left that I don't how to fix: in order to do this, I need to know which controllers is our _context_ in order to set its frame to the `dimmingView`. Since we are already in the modal environment, we don't know which controller starts the presentation. There an easy solution when doing presentation programatically, but I have no idea how to get that context while supporting the interface builder. Any suggestions would help!

For example, here the result with a hard coded frame:

simulator screen shot 28 mar 2017 15 49 10

The first part is the one presenting the controller, but while being presented, the second stay clickable.

**Other things missing:**
- [x] Demo app
- [x] Changelog
- [x] Docs
- [ ] Documentation gif

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 #### API breaking changes
 - `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programatically a cornerSide to your view, you will just have to lowercase the A. [#409](https://github.com/IBAnimatable/IBAnimatable/pull/409) by [@tbaranes](https://github.com/tbaranes)
 - `AnimatableSlider` inherit from `UISlider`
-- `PresentationDesignable` now supports `overCurrentScreen` which allow you to present a controller with a custom configuration over another instead of being in fullscreen.
+- `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
 
 #### Enhancements
 - Conserve custom layer mask when using `Animatable*` instead of removing them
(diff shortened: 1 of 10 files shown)


tbaranes:
One solution I think of is to add a custom segue for this particular case. When using programatically, you will have to set programatically the frame, but when using a segue, we can access to the `source`.

tbaranes:
I got a workaround for that issue:

 - If programatically -> set the frame manually
 - If interface builder -> using a custom segue

There's still a few bug for the modal position (related to the nav bar), but the result is closer than before, and easier to use from interface builder.

tbaranes:
Closing in favour of #424

PR #422, opened 2017-03-30
Enable SwiftLint line length #trivial

### LINKED ISSUE #371: Enable SwiftLint line_length
Can be a tricksy task, I don't think all the warnings will be easy to fix but it will definitely improve the code. #369

tbaranes:
Close #371


JakeLin (approved):
👍 all good, that is non-trivial though 😁

PR #423, opened 2017-03-30
Update storyboard to Xcode 8.3 #trivial

phimage:
Storyboards updated using Xcode 8.3

The aim is to not commit all this changes when doing other modifications into storyboards


tbaranes (approved):

tbaranes:
Travis fails because Danger reports that there's no CHANGELOG entry, in this case, since we don't want to add an entry, you can just add `#trivial` to your PR title in order to make Danger ignore that condition.

phimage:
ok thanks for the advice

JakeLin (approved):
👍

JakeLin:
@phimage Can you please merge it when you think it is ready?

PR #424, opened 2017-03-30
Presentation over current context

### LINKED ISSUE #198: AnimatablePresentationController

tbaranes:
Opening that PR over #421 

This PR add a new feature to present a viewController: presenting over current context, in other words, presenting with a custom frame. It will also close #198!

What's in:

- A new property for `PresentationDesignable `: `contextFrameForPresentation: CGRect?`: 
   - If nil: nothing happened, the presentation will be made in fullscreen
   - If not nil: the dimmingView frame will be equal to that one and the modal origin / size will be calculated in function of that new frame. That allow the user to present a viewController in a specific context: for example, if you have a split view controller, you may want to present a controller over the first one but let the second _without_ presentedVC.
- A new segue `PresentOverCurrentContextSegue`: using that segue allow the user to present a controller over the current context directly in storyboards. Before presenting the `destination`, the segue will simply set the `source.view.frame` as `contextFrameForPresentation`. If presenting programatically, the user can directly set `contextFrameForPresentation` to the value he wants.
- Introducing `PresentationBackgroundView`: in reality, the presentation is still in fullscreen. We are just hacking around the frames to give the impression that isn't in fullscreen. That new clear background allow us to forward the `hit` to the others subviews, which make the interface _not presented_ clickable.
- A few code improvements in the presentation classes
- Docs update: CHANGELOG and Presentations.md
- Demo update

This PR is ready for review, and ready to go!

*Note*: In the example app, the back button won't catch the events while presenting a viewController. After a deeper search, I concluded that's due to the navigation: we are trying to `pop` while `presenting`, so the pop is just ignored. The back button catch the event as expected, but doesn't treat it. We can probably make a workaround for this by handling this state programatically, but since I didn't want to introduce code for that feature, I let it _buggy_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 - `AnimatableSlider` inherit from `UISlider`
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
-- `PresentationDesignable` now supports `overCurrentScreen` which allow you to present a controller with a custom configuration over another instead of being in fullscreen.
+- `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
 
 #### Enhancements
 - Conserve custom layer mask when using `Animatable*` instead of removing them
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - `AnimatableSlider` inherit from `UISlider`
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
+- `PresentationDesignable` now supports `overCurrentScreen` which allow you to present a controller with a custom configuration over another instead of being in fullscreen.
 
 #### Enhancements
 - Conserve custom layer mask when using `Animatable*` instead of removing them
--- a/Documentation/Presentations.md
+++ b/Documentation/Presentations.md
@@ -15,7 +15,6 @@ Then we can configure the **Presentation Animations**, **Transition Duration** a
 
 | Property | Description |
 | ------------- | ------------- |
-| Over Current Context | If true, the presentation will have the same size as the controller doing the presentation. If false, the presentation will be in fullscreen. In both cases, the others parameters will be applied. |
 | Presentation Animation | The animation effect when *Present* a `ViewController`, you can find all supported transition animations in [Presentation Animators](#presentation-animators) section. |
 | Dismissal Animation | The animation effect when *Dismiss* a `ViewController`, you can find all supported transition animations in [Presentation Animators](#presentation-animators) section. By default, it will use the Presentation animation value. For example, if the Presentation animation is cover down from the top, then the default Dismissal Animation is cover up to the top. |
 | Transition Duration | The duration of the transition animation in seconds. The default value is `0.4` |
@@ -211,6 +210,40 @@ If your modal contains a `UITextField` or `UITextView`, you can adjust its posit
 ![Transition - Fade Transition](https://raw.githubusercontent.com/IBAnimatable/IBAnimatable-Misc/master/IBAnimatable/KeyboardTranslationAboveKeyboard.gif)
 
 You can see all supported Keyboard Translation in the demo App, open the App and tap on "Playground" button, then tap on "presentations", then play with the different options.
+
+## Presenting over current context
+
+Using the above configuration, you can't have the same result as using `UIModalPresentationStyle.currentContext`, but `IBAnimatable` have a workaround to support it while supporting *all the above customisation*.
+
+That can be useful in a few cases, for example: having a split view controller, but you may want to present a controller over the first one, and let the second one clickable. You can see an example in the demo app: Playground -> Presentation -> Over context.
+
+![Presentation - Over Current Context](https://raw.githubusercontent.com/IBAnimatable/IBAnimatable-Misc/master/IBAnimatable/PresentationOverCurrentContext.gif)
+
+### Using storyboards and segues
+
+In order to use this feature in storyboards, you have to use a custom segue. To use custom Segue, we can **control drag** from one `ViewController` to another `ViewController`, then select a custom Segue and set its class to `PresentOverCurrentContextSegue`.
+
+Using that usage allow you to use all the features of a custom presentation while writing 0 lines of code.
+
+You can see an example in "Presentation.storyboard".
+
+### Programatically
+
+For this special case, `AnimatableModalViewController` has another property that's not designable:
+
+| Property | Description |
+| ------------- | ------------- |
+| context frame for presentation | If not nil, the presented view controller will have use this frame, imitates `UIModalPresentationStyle.currentContext`  If nil, the presented view controller will be in fullscreen. *Note:* The modal position / size will be calculated based on this if not nil. |
+
+Anywhere before `presenting` your viewController, just set the `contextFrameForPresentation` that it should have.
+For example, if you want your presented view controller to have the same frame as the `viewController` presenting it, you will just have to do:
+
+```
+modalViewController.contextFrameForPresentation = presentingVC.view.frame
+modalViewController.present(presentingVC, animated: true)
+```
+
+That's all. You have presented a controller over the current context with a custom configuration!
 
 ## Contribution
 
--- a/Documentation/Presentations.md
+++ b/Documentation/Presentations.md
@@ -15,6 +15,7 @@ Then we can configure the **Presentation Animations**, **Transition Duration** a
 
 | Property | Description |
 | ------------- | ------------- |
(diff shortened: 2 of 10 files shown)


JakeLin:
@tbaranes please allow me to spend some time on this PR. As Danger said it is a "Big PR" 😉

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 #### API breaking changes
 - `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programmatically a cornerSide to your view, you will just have to lowercase the A. [#409](https://github.com/IBAnimatable/IBAnimatable/pull/409) by [@tbaranes](https://github.com/tbaranes)
 - `AnimatableSlider` inherit from `UISlider`
+- `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
```
JakeLin: Are we imitating `overCurrentContext`?

Documentation/Presentations.md
```diff
@@ -211,6 +211,40 @@ If your modal contains a `UITextField` or `UITextView`, you can adjust its posit
 
 You can see all supported Keyboard Translation in the demo App, open the App and tap on "Playground" button, then tap on "presentations", then play with the different options.
 
+## Presenting over current context
+
+Using the above configuration, you can't have the same result as using `UIModalPresentationStyle.currentContext`, but `IBAnimatable` have a workaround to support it while supporting *all the above customisation*.
```
JakeLin: Should be `overCurrentContext` ?

Documentation/Presentations.md
```diff
@@ -211,6 +211,40 @@ If your modal contains a `UITextField` or `UITextView`, you can adjust its posit
...
+
+That can be useful in a few cases, for example: having a split view controller, but you may want to present a controller over the first one, and let the second one clickable. You can see an example in the demo app: Playground -> Presentation -> Over context.
+
+![Presentation - Over Current Context](https://raw.githubusercontent.com/IBAnimatable/IBAnimatable-Misc/master/IBAnimatable/PresentationOverCurrentContext.gif)
+
+### Using storyboards and segues
+
+In order to use this feature in storyboards, you have to use a custom segue. To use custom Segue, we can **control drag** from one `ViewController` to another `ViewController`, then select a custom Segue and set its class to `PresentOverCurrentContextSegue`.
+
+Using that usage allow you to use all the features of a custom presentation while writing 0 lines of code.
```
JakeLin, line 226: Sounds like magic ⚡️

Documentation/Presentations.md
```diff
@@ -211,6 +211,40 @@ If your modal contains a `UITextField` or `UITextView`, you can adjust its posit
...
+
+You can see an example in "Presentation.storyboard".
+
+### Programatically
+
+For this special case, `AnimatableModalViewController` has another property that's not designable:
+
+| Property | Description |
+| ------------- | ------------- |
+| context frame for presentation | If not nil, the presented view controller will have use this frame, imitates `UIModalPresentationStyle.currentContext`  If nil, the presented view controller will be in fullscreen. *Note:* The modal position / size will be calculated based on this if not nil. |
```
JakeLin: `overCurrentContext` as well?

IBAnimatable/AnimatablePresentationController.swift
```diff
@@ -45,9 +50,12 @@ public class AnimatablePresentationController: UIPresentationController {
 
 private extension AnimatablePresentationController {
 
-  func setupDimmingView() {
-    let tap = UITapGestureRecognizer(target: self, action: #selector(chromeViewTapped))
+  func configureDimmingView() {
+    var tap = UITapGestureRecognizer(target: self, action: #selector(dimmingViewTapped))
     dimmingView.addGestureRecognizer(tap)
+    tap = UITapGestureRecognizer(target: self, action: #selector(dimmingViewTapped))
```
JakeLin, line 56: Can we use the single `tap` variable or shall we have separate instances of `UITapGestureRecognizer` here?

IBAnimatable/AnimatablePresentationController.swift
```diff
@@ -79,7 +87,7 @@ private extension AnimatablePresentationController {
 
 extension AnimatablePresentationController {
 
-  fileprivate func setupObservers() {
+  fileprivate func makeObservers() {
```
JakeLin: should `make` return an object of objects?

IBAnimatable/AnimatablePresentationController.swift
```diff
@@ -116,25 +124,18 @@ extension AnimatablePresentationController {
...
-      return nil
-    }
-
-    return presentationConfiguration.modalPosition.calculateCenter(containerBounds: containerBounds, modalSize: modalSize())
+  var modalCenter: CGPoint? {
+    return presentationConfiguration.modalPosition.modalCenter(in: containerFrame, modalSize: modalSize)
   }
 
-  func modalOrigin() -> CGPoint? {
+  var modalOrigin: CGPoint? {
```
JakeLin, line 138: They are all better 👍

IBAnimatable/PresentationBackgroundView.swift
```diff
@@ -0,0 +1,30 @@
...
+//  PassthroughBackgroundView.swift
+//  IBAnimatable
+//
+//  Created by Tom Baranes on 28/03/2017.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import UIKit
+
+final class PresentationBackgroundView: UIView {
```
JakeLin, line 11: Interesting idea to have a passthrough view. I don't know how system's `.overCurrentContext` achieve this goal. Do they do the same?

JakeLin (approved):
:+1 Well done @tbaranes. Do we need to check `defines​Presentation​Context` before we

JakeLin:
@tbaranes For some network issue, I can't change the text in the review. I was thinking do we need to check `defines​Presentation​Context` before we present the `overCurrentContext`? I am not good at this API.  Can you make a decision which one is better?

CHANGELOG.md
tbaranes: Yeah, the implementation is different I suppose, but the result is expected the same, if it's not, it's a bug 😬

Documentation/Presentations.md
tbaranes: 😨 Wups.

IBAnimatable/AnimatablePresentationController.swift
tbaranes: Should I rename it `configureObservers` ?

IBAnimatable/PresentationBackgroundView.swift
tbaranes: I have no idea... either they do that, otherwise they must have access to a few private API.

IBAnimatable/AnimatablePresentationController.swift
tbaranes: I think `UITapGestureRecognizer` can be bind to only one view at once which means if I add it to a second view, the first one will lost its gesture.
I will do a test to confirm.

tbaranes: I confirm this behaviour.

tbaranes:
@JakeLin About your last point, if we want to use custom presentaiton, the `modalPresentationStyle` must be set to `custom`, otherwise it won't work. Since `definesPresentationContext` works only with `currentContext` and `overCurrentContext`, whatever its value is, it won't change the behaviour of the custom presentation. So checking its value in our context is useless.

IBAnimatable/AnimatablePresentationController.swift
JakeLin: In that case, shall we have two local variables for each tap gesture?

IBAnimatable/PresentationBackgroundView.swift
JakeLin: @tbaranes you're right.

JakeLin:
@tbaranes got it. Because we need to use .custom . Please merge it when you think its ready👍👍👍

IBAnimatable/AnimatablePresentationController.swift
tbaranes: Not sure, since they are exactly the same and we don't stored them as a property. I can change it if you think that's better

JakeLin: I thought about again. In the memory model, using one var or two let are same. Because it will be two instances and retained by two different views. I am OK with that now.

PR #418, opened 2017-03-26
Add `AnimatableTableViewController` with designable `UIRefreshControl`

phimage:
and other features same as  `AnimatableViewController`

`RefreshControlDesignable` allow to custom the internal `UIRefreshControl`

`UIRefreshControl` must not be set in subview but using `refreshControl` attribute of `UITableViewController`
I think apple must do some job to make it accessible through Interface Builder, but until that this feature allow to custom the basic attribute of `UIRefreshControl`

---

I can go further to allow text edition
by adding something like that in protocol `RefreshControlDesignable`
```swift
var refreshControlTitle: String? { get set }
var refreshControlTitleFont: UIFont? { get set }
var refreshControlTitleColor: UIFont? { get set }
```
and there in configure function do the job to create an attributed string 
```swift

if let title = refreshControlTitle {
  var attributes: [String : Any] = [:]
  if let font = refreshControlTitleFont {
    attributes[NSFontAttributeName] = font
  }
  if let color = refreshControlTitleColor {
    attributes[NSForegroundColorAttributeName] = color
  }
  refreshControl?.attributedTitle = NSAttributedString(string: title, attributes: attributes)
} else {
  refreshControl?.attributedTitle = nil
} 
```

PR #425, opened 2017-03-31
Add `AnimatableTableViewController` with designable `UIRefreshControl`

phimage:
and other features same as `AnimatableViewController`

`RefreshControlDesignable` allow to custom the internal `UIRefreshControl`

`UIRefreshControl` must not be set in subview but using refreshControl attribute of `UITableViewController`

ReStart of PR #418 due to many conflict
Modification since review by @tbaranes and @JakeLin 
- remove some space and add missing open on IBoutlet variable copied from AnimatableViewController
- add `TimeLineTableViewController` in demo app , to show how refresh control could be customised (simulate the refresh using `DispatchQueue`)
- fix a bug : if view not already loaded. If view not loaded, refreshControl could not be set

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Next
 
 #### API breaking changes
-- `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programatically a cornerSide to your view, you will just have to lowercase the A. [#409](https://github.com/IBAnimatable/IBAnimatable/pull/409) by [@tbaranes](https://github.com/tbaranes)
+- `CornerSide`'s swift3 migration leftovers: renaming `.AllSides` to `.allSides`. If you were setting programmatically a cornerSide to your view, you will just have to lowercase the A. [#409](https://github.com/IBAnimatable/IBAnimatable/pull/409) by [@tbaranes](https://github.com/tbaranes)
 - `AnimatableSlider` inherit from `UISlider`
 
 #### Enhancements
@@ -16,13 +16,15 @@ All notable changes to this project will be documented in this file.
 [#403](https://github.com/IBAnimatable/IBAnimatable/pull/403) by [@tbaranes](https://github.com/tbaranes)
 - Make images of `AnimatableSlider` designable.
 [#417](https://github.com/IBAnimatable/IBAnimatable/pull/417) by [@phimage](https://github.com/phimage)
-
+- Add `AnimatableTableViewController` to support `UIRefreshControl` customization.
+[#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage)
+CHANGELOG
 
 #### Bugfixes
 - Make corner sides case insensitive.
 [#394](https://github.com/IBAnimatable/IBAnimatable/issues/394) by [@mmadjer](https://github.com/mmadjer)
 - Frame is converted to window coordinate space to fix miscalculations in computed values (used with `slideOut`, ...) [#412](https://github.com/IBAnimatable/IBAnimatable/issues/412) by [@redent](https://github.com/redent)
- 
+
 ---
 ### [3.1.3](https://github.com/IBAnimatable/IBAnimatable/releases/tag/3.1.3)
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -16,13 +16,15 @@ All notable changes to this project will be documented in this file.
 [#403](https://github.com/IBAnimatable/IBAnimatable/pull/403) by [@tbaranes](https://github.com/tbaranes)
 - Make images of `AnimatableSlider` designable.
 [#417](https://github.com/IBAnimatable/IBAnimatable/pull/417) by [@phimage](https://github.com/phimage)
-
+- Add `AnimatableTableViewController` to support `UIRefreshControl` customization.
+[#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage)
+CHANGELOG
```
tbaranes: There's a `CHANGELOG` to remove here!

Documentation/APIs.md
```diff
@@ -170,6 +171,26 @@ Easily add color layer on top of the UI element especially `AnimatableImageView`
 | ------------- |:-------------:| ----- |
 | hideNavigationBar | Bool | whether to hide navigation bar. The default value is `false`. |
 
+#### `RefreshControlDesignable`
+Display a `UIRefreshControl` in `AnimatableTableViewController`
```
tbaranes: I wouldn't add this line, I think the protocol name is clear enough (moreover, we may conform to this protocol in other classes).

IBAnimatable/AnimatableTableViewController.swift
```diff
@@ -0,0 +1,111 @@
+//
+//  AnimatableTableViewController.swift
```
tbaranes: You can remove these two lines. Take a look at the other headers of the project, we are just keeping the author lines.

IBAnimatable/RefreshControlerDesignable.swift
```diff
@@ -0,0 +1,41 @@
...
+   The background color of the refresh control
+   */
+  var refreshControlBackgroundColor: UIColor? { get set }
+
+}
+
+public extension RefreshControlDesignable where Self: UITableViewController {
+
+  func configureRefreshController() {
+    if isViewLoaded {
```
tbaranes: Guard?

IBAnimatableApp/TimeLineTableViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+    self.refreshControl?.addTarget(self, action: #selector(TimeLineTableViewController.refresh(_:)), for: .valueChanged)
+  }
+
+  public func refresh(_ refreshControl: UIRefreshControl) {
+    // could update attributedTitle of refreshControl here
+
+    // Simulate an asynchrone refresh, could be a network request...
+    let time: TimeInterval = 5
+    self.updateMessage(refreshControl: refreshControl, time: time)
+    DispatchQueue.background.after(time) {
```
tbaranes: Why not directly running this on `main`?

IBAnimatableApp/TimeLineTableViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+      // end refreshing, maybe reload table data if you do not implement table delegate to update each insert, update and delete events
+      DispatchQueue.main.async {
+        refreshControl.endRefreshing()
+      }
+    }
+  }
+
+  // Function to update message recursively
+  func updateMessage(refreshControl: UIRefreshControl, time: TimeInterval) {
+    if time < 0 {
```
tbaranes: guard?

tbaranes (changes requested):
I let you a few more minor comments, if you can take a look at them... :)

Also, a new swiftlint warning appeared since we enabled `line_length`, can you take care of it? If you want to see the warnings directly in Xcode, you just have to install [SwiftLint](https://github.com/realm/SwiftLint)

tbaranes:
One more thing that we should have precise it earlier, my apologies. We don't add features to the `Demo.storyboard`, but directly in the `Playground` part which is a showoff of all the features available in `IBAnimatable`.

For the refresh control, you could add an entry either in:

- `Playground` -> `User Interface` -> `Refresh Control`
- `Playground` -> `RefreshControl`  (new) -> `UITableView` 
- Third place?

Both cases are ok, it's up to you, maybe depending if we want to add an example for the `collectionView` (when we will support it)

Sorry for not letting you know about this before 😟

IBAnimatableApp/TimeLineTableViewController.swift
phimage: To simulate a good practice, doing a  job in background like get data from network

phimage:
I use swiftlint and update it
I have missing one

IBAnimatable/RefreshControlerDesignable.swift
phimage: agree

IBAnimatableApp/TimeLineTableViewController.swift
phimage: ummm here not agree `if` is shorter and more understandable  but I will do it

phimage:
I have no time to make new storyboards in playground 
And I have no internet at home currently, could merge without the demo?

tbaranes (approved):
Let's wait for @JakeLin review, but  I'm ok to merge it without the demo. We can add it later! Thanks for taking care of the comments.

IBAnimatable/AnimatableTableViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+  @IBInspectable open var hideNavigationBar: Bool = false
+
+  // MARK: - StatusBarDesignable
+  @IBInspectable open var lightStatusBar: Bool = false
+
+  // MARK: - RootWindowDesignable
+  @IBInspectable open var rootWindowBackgroundColor: UIColor?
+
+  // MARK: - TransitionAnimatable
+  @IBInspectable open var _transitionAnimationType: String? {
```
JakeLin: I think we don't want that to be `open` and let it only visible in Interface Builder. We can remove `open` and let it as  the default `internal`.

IBAnimatable/AnimatableTableViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+        transitionAnimationType = TransitionAnimationType(string: _transitionAnimationType)
+      }
+    }
+  }
+  open var transitionAnimationType: TransitionAnimationType = .none
+
+  @IBInspectable open var transitionDuration: Double = .nan
+
+  open var interactiveGestureType: InteractiveGestureType = .none
+  @IBInspectable open var _interactiveGestureType: String? {
```
JakeLin: Same comment as above.

IBAnimatableApp/RefreshTableViewController.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  Created by Eric Marchand on 31/03/2017.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import UIKit
+import IBAnimatable
+
+public class RefreshTableViewController: AnimatableTableViewController {
```
JakeLin, line 9: Nice demo app 👍

JakeLin (approved):
👍 Well done @phimage

JakeLin:
@phimage I have fixed my comment and merged the PR now 🎉🎉🎉. Thank you very much for creating such a great feature for IBAnimatable. It makes our life easier to configure the refersh control in Interface Builder.

For `@IBInspectable var _transitionAnimationType: String?`, because Interface Builder can't support enum. And we want to have nice programmatical API with strongly type like `open var transitionAnimationType: TransitionAnimationType = .none`. To do that, we found out a workaround, we use `String` as the type to make the inspectable property can be configured in Interface Builder. But don't want the user/client code to see `_transitionAnimationType`, then we make them as `internal`. In that case, we can configure `transitionAnimationType` in Interface Builder and use it in code as an enum. More info can be found in #218 @lastMove and @tbaranes did a great job on that.

tbaranes:
For your information, I added an entry in #238 in order to not forget that we have to update the playground storyboard in order to have a demo for refresh control.

PR #429, opened 2017-04-02
Add RefreshControlDesignable example

tbaranes:
Related to #238 and #425 

A few things in this PR:

- Make `AnimatableTableView` conforms to `RefreshControlDesignable`<s>, iOS10 only, I think we can add a support for lower version, by adding manually the refreshControl. Is it worth it?</s>
- Add `RefreshControlDesignable` examples in playground:
   - For `UITableViewController`.
  - For `UITableView`

Seems big, but its mainly storyboards!

<s>Just one thing: I'm getting these warnings in the console for the `UITableViewController` example, but don't know why 🤔 

```
Failed to set (hasRefreshControl) user defined inspected property on (IBAnimatableApp.RefreshControlTableViewController): Could not load NIB in bundle: 'NSBundle <../IBAnimatableApp.app> (loaded)' with name '' 
Failed to set (refreshControlTintColor) user defined inspected property on (IBAnimatableApp.RefreshControlTableViewController): Could not load NIB in bundle: 'NSBundle <../IBAnimatableApp.app> (loaded)' with name '' 
Failed to set (refreshControlBackgroundColor) user defined inspected property on (IBAnimatableApp.RefreshControlTableViewController): Could not load NIB in bundle: 'NSBundle <../IBAnimatableApp.app> (loaded)' with name '' 
```
</s>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ All notable changes to this project will be documented in this file.
 [#403](https://github.com/IBAnimatable/IBAnimatable/pull/403) by [@tbaranes](https://github.com/tbaranes)
 - Make images of `AnimatableSlider` designable.
 [#417](https://github.com/IBAnimatable/IBAnimatable/pull/417) by [@phimage](https://github.com/phimage)
-- Add `AnimatableTableViewController` to support `UIRefreshControl` customization.
-[#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage)
+- Add `RefreshControlDesignable` to make `UIRefreshControl` customization available in interface builder. Currently supported by `UITableViewController` and `UITableView`
+[#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage) and [#429](https://github.com/IBAnimatable/IBAnimatable/pull/429) by [@tbaranes](https://github.com/tbaranes)
 
 #### Bugfixes
 - Make corner sides case insensitive.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All notable changes to this project will be documented in this file.
 [#403](https://github.com/IBAnimatable/IBAnimatable/pull/403) by [@tbaranes](https://github.com/tbaranes)
 - Make images of `AnimatableSlider` designable.
 [#417](https://github.com/IBAnimatable/IBAnimatable/pull/417) by [@phimage](https://github.com/phimage)
-- Add `RefreshControlDesignable` to make `UIRefreshControl` customization available in interface builder. Currently supported by `UITableViewController` and `UITableView`
+- Add `RefreshControlDesignable` to make `UIRefreshControl` customization available in Interface Builder. Currently supported by `UITableViewController` and `UITableView`
 [#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage) and [#429](https://github.com/IBAnimatable/IBAnimatable/pull/429) by [@tbaranes](https://github.com/tbaranes)
 
 #### Bugfixes
--- a/IBAnimatable/RefreshControlerDesignable.swift
+++ b/IBAnimatable/RefreshControlerDesignable.swift
@@ -25,6 +25,10 @@ public protocol RefreshControlDesignable {
 public extension RefreshControlDesignable where Self: UITableViewController {
 
   func configureRefreshController() {
+    guard isViewLoaded else {
+      return
+    }
+
     configureRefreshController(hasRefreshControl: hasRefreshControl, refreshControl: &refreshControl)
   }
 
--- a/IBAnimatable/RefreshControlerDesignable.swift
+++ b/IBAnimatable/RefreshControlerDesignable.swift
@@ -25,18 +25,32 @@ public protocol RefreshControlDesignable {
 public extension RefreshControlDesignable where Self: UITableViewController {
 
   func configureRefreshController() {
-    guard !isViewLoaded else {
-      return
-    }
-    if hasRefreshControl {
-      if refreshControl == nil {
-        refreshControl = UIRefreshControl()
-      }
-      refreshControl?.tintColor = refreshControlTintColor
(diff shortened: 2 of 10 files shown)


phimage:
For the warning I think it's because we play with the `refreshControl` before view loading
That's why I add in `configureRefreshController` 
```swift
 guard !isViewLoaded else {
      return
    }
```
https://github.com/IBAnimatable/IBAnimatable/blob/master/IBAnimatable/RefreshControlerDesignable.swift

Adding for the previous ios version is welcome :+1:
I want to work on that later, because I want to add support to collection too.

tbaranes:
Wups, I miss this. You are right the warning has been fixed by putting back the guard, good catch! Thanks

tbaranes:
Added support of `AnimatableTableView` `RefreshControlDesignable`  for iOS version inferior to 10.

There's still a restriction: the user has to add himself the refresh control to its tableView, then call the `configureRefreshControls` (such as in the example). That make the usage harder, and ask for more code than the iOS10 version. Still, I choose this because I think it would have been even more complicated to ask the user to _search_ its `refreshControl` in the `tableView`'s subviews in order to add a target.

What do you prefer? I'm ok to change, and I will update the docs following the final decision

CHANGELOG.md
```diff
@@ -16,8 +16,8 @@ All notable changes to this project will be documented in this file.
 [#403](https://github.com/IBAnimatable/IBAnimatable/pull/403) by [@tbaranes](https://github.com/tbaranes)
 - Make images of `AnimatableSlider` designable.
 [#417](https://github.com/IBAnimatable/IBAnimatable/pull/417) by [@phimage](https://github.com/phimage)
-- Add `AnimatableTableViewController` to support `UIRefreshControl` customization.
-[#418](https://github.com/IBAnimatable/IBAnimatable/pull/418) by [@phimage](https://github.com/phimage)
+- Add `RefreshControlDesignable` to make `UIRefreshControl` customization available in interface builder. Currently supported by `UITableViewController` and `UITableView`
```
JakeLin: **Interface Builder** 😅

IBAnimatableApp/RefreshControlTableViewController.swift
```diff
@@ -6,24 +6,23 @@
...
 
-public class RefreshTableViewController: AnimatableTableViewController {
+public class RefreshControlTableViewController: AnimatableTableViewController {
 
   override public func viewDidLoad() {
     super.viewDidLoad()
 
     // Install action on refresh
-    self.refreshControl?.addTarget(self, action: #selector(RefreshTableViewController.refresh(_:)), for: .valueChanged)
+    self.refreshControl?.addTarget(self, action: #selector(refresh(_:)), for: .valueChanged)
```
JakeLin: Remove `self` here?

IBAnimatableApp/RefreshControlTableViewController.swift
```diff
@@ -36,21 +35,24 @@ public class RefreshTableViewController: AnimatableTableViewController {
...
+
+    var attributes = [String: Any]()
+    if let color = refreshControlTintColor {
       attributes[NSForegroundColorAttributeName] = color
     }
     refreshControl.attributedTitle = NSAttributedString(string: "\(Int(time))", attributes: attributes )
 
-    DispatchQueue.main.after(1) { [unowned self] in
-      self.updateMessage(refreshControl: refreshControl, time: time - 1)
+    DispatchQueue.main.after(1) { [weak self] in
```
JakeLin: I don't think we need to use `[weak self]` or `[unowned self]` for `DispatchQueue.main`. They are static methods.

IBAnimatable/RefreshControlerDesignable.swift
```diff
@@ -25,18 +25,40 @@ public protocol RefreshControlDesignable {
...
+}
+
+public extension RefreshControlDesignable where Self: UITableView {
+
+  func configureRefreshController() {
+    if #available(iOSApplicationExtension 10.0, *) {
+      configureRefreshController(hasRefreshControl: hasRefreshControl, refreshControl: &refreshControl)
     } else {
-      refreshControl = nil
+      var refreshControl = self.subviews.first { $0 is UIRefreshControl } as? UIRefreshControl
```
JakeLin, line 43: Can we do this automatically? For example, we check there is a `UIRefreshControl` or not. Add one if not.

IBAnimatableApp/RefreshControlTableViewViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+    super.viewDidLoad()
+
+    // Install action on refresh
+    let refreshControl: UIRefreshControl?
+    if #available(iOS 10.0, *) {
+      refreshControl = tableView.refreshControl
+    } else {
+      refreshControl = UIRefreshControl()
+      tableView.addSubview(refreshControl!)
+      tableView.configureRefreshController()
```
JakeLin, line 27: I made a comment on https://github.com/IBAnimatable/IBAnimatable/pull/429/files#diff-f56d938957f5c1840f24beee6993086fR43

If we can't do it automatically, we may need to put this code to `AnimatableTableView` as a comment for pre-iOS 10.

JakeLin (approved):
@tbaranes it is very good to support refresh control for pre-iOS 10. I have an idea: Can you add it automatically for pre-iOS 10? If we can't, I think we need to add the example code to `AnimatableTableView` as a comment. 

Apart from that, all good 👍, well done 💪

tbaranes:
@JakeLin We can add it automatically, but that means when the user will add a target, he will have to look into IBAnimatable to see how to get it then bind it. The example `RefreshControlTableViewViewController` (yeah, that's a big name 😃) is working for both case, iOS10 and pre-10. Just need to update the documentation following our final decision.

What do you prefer? Should I still add it automatically?

IBAnimatable/RefreshControlerDesignable.swift
tbaranes: We can,  see my answer below!

CHANGELOG.md
tbaranes: Fixed

JakeLin:
@tbaranes I think what you are doing is good. Please merge it when you think it is ready.

PR #430, opened 2017-04-02
Update Dangerfile #no-public-changes

tbaranes:
As discussed on slack:

- Replace `#trivial` by `#no-public-changes`
- Add a warning when the demo app has not been updated (just in case)


JakeLin (approved):
👍 you are so quick ⚡️

PR #431, opened 2017-04-03
Fix animations app on pre-ios10 #no-public-changes

tbaranes:
On pre-iOS10, the animation picker in playground won't be enabled after playing one animation


JakeLin (approved):
👍

PR #435, opened 2017-04-07
[WIP]: Refactor ActivityIndicators

SD10:
## Summary of Pull Request

1. Condensed `ActivityIndicatorTriangleSkewSpin` and `ActivityIndicatorSquareSpin` 
into a new class called `ActivityIndicatorSkewSpin`

2. I removed `ActivityIndicatorBallZigZagDeflect`

### Reasoning
- `ActivityIndicatorTriangleSkewSpin` and `ActivityIndicatorSquareSpin`

These two classes shared the same animation and only differed in their shape.
I provided an initializer that takes the shape as an argument.

- `ActivityIndicatorBallZigZag` and `ActivityIndicatorBallZigZagDeflect`

These two classes were also nearly identical.
The only difference is `ActivityIndicatorBallZigZagDeflect` auto-reverses
I provided an initializer that takes a flag `deflects` as an argument

### Extra

I haven't added a CHANGELOG entry yet. This is just record keeping.


tbaranes (approved):
Nice refactoring 👍 
I don't think that needs a changelog entry since it's only internal refactatoring, you can add #no-public-changes to your PR title

JakeLin (commented):
I think there are breaking changes for this PR. There's no problem with breaking changes, we just need to update the change log to describe that. I think we also need to update the demo app to test this change.

tbaranes:
@SD10 Can you just update the CHANGELOG? Otherwise, i think that ready to go 👍

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file.
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
 - `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
+- `ActivityIndicatorTriangleskewSpin/ActivityIndicatorSquareSpin` replaced by `ActivityIndicatorSkewSpin`
```
JakeLin, line 14: Can you also add `by [@SD10](https://github.com/sd10)`?

PR #437, opened 2017-04-07
Add Gradient to NavigationBar

phimage:
Instead of adding a layer, I set the background image for UINavigationBar

- for the frame height it's not very clean. I set 64. Don't know if there is a constant somewhere
If you put only the bound height you do not draw gradient to the top of the screen

Tested on playground

### DIFF
--- a/IBAnimatable/GradientDesignable.swift
+++ b/IBAnimatable/GradientDesignable.swift
@@ -25,42 +25,10 @@ public extension GradientDesignable where Self: UIView {
 
 fileprivate extension GradientDesignable where Self: UIView {
   func configureGradient(startColor: UIColor, endColor: UIColor) {
-    // Default value is `.Top`
 
-    let gradientStartPoint = startPoint
     let gradientLayer = makeGradientLayer()
     gradientLayer.colors = [startColor.cgColor, endColor.cgColor]
-    switch gradientStartPoint {
-    case .top:
-      gradientLayer.startPoint = CGPoint(x: 0.5, y: 0)
-      gradientLayer.endPoint = CGPoint(x: 0.5, y: 1)
-    case .topRight:
-      gradientLayer.startPoint = CGPoint(x: 1, y: 0)
-      gradientLayer.endPoint = CGPoint(x: 0, y: 1)
-    case .right:
-      gradientLayer.startPoint = CGPoint(x: 1, y: 0.5)
-      gradientLayer.endPoint = CGPoint(x: 0, y: 0.5)
-    case .bottomRight:
-      gradientLayer.startPoint = CGPoint(x: 1, y: 1)
-      gradientLayer.endPoint = CGPoint(x: 0, y: 0)
-    case .bottom:
-      gradientLayer.startPoint = CGPoint(x: 0.5, y: 1)
-      gradientLayer.endPoint = CGPoint(x: 0.5, y: 0)
-    case .bottomLeft:
-      gradientLayer.startPoint = CGPoint(x: 0, y: 1)
-      gradientLayer.endPoint = CGPoint(x: 1, y: 0)
-    case .left:
-      gradientLayer.startPoint = CGPoint(x: 0, y: 0.5)
-      gradientLayer.endPoint = CGPoint(x: 1, y: 0.5)
-    case .topLeft:
-      gradientLayer.startPoint = CGPoint(x: 0, y: 0)
-      gradientLayer.endPoint = CGPoint(x: 1, y: 1)
-    case let .custom(start, end):
-      gradientLayer.startPoint = CGPoint(x: start.x, y: start.y)
-      gradientLayer.endPoint = CGPoint(x: end.x, y: end.y)
-    case .none:
-      break
-    }
+    gradientLayer.configurePoints(with: startPoint)
 
     subviews.filter { $0 is PrivateGradientView }.forEach {
       $0.removeFromSuperview()
@@ -78,12 +46,46 @@ fileprivate extension GradientDesignable where Self: UIView {
   }
 
   func configurePredefinedGradient() -> GradientColor? {
-
     guard let gradientType = predefinedGradient else {
       return nil
     }
     return gradientType.colors
   }
+}
+
+public extension GradientDesignable where Self: UINavigationBar {
+  public func configureGradient() {
+    let predefinedGradient = configurePredefinedGradient()
+    if let startColor = startColor, let endColor = endColor {
+      configureGradient(startColor: startColor, endColor: endColor)
+    } else if let startColor = predefinedGradient?.start, let endColor = predefinedGradient?.end {
+      configureGradient(startColor: startColor, endColor: endColor)
+    }
+  }
+}
+
+fileprivate extension GradientDesignable where Self: UINavigationBar {
+
+  func configureGradient(startColor: UIColor, endColor: UIColor) {
+    let gradientLayer = makeGradientLayer()
+    gradientLayer.colors = [startColor.cgColor, endColor.cgColor]
+    gradientLayer.configurePoints(with: startPoint)
+    let image = gradientLayer.image()
+
+    self.setBackgroundImage(image, for: .default)
+  }
+
+  func makeGradientLayer() -> CAGradientLayer {
+    let gradientLayer: CAGradientLayer = CAGradientLayer()
+
+    let statusBarHeight: CGFloat = 64 // UIApplication.shared.statusBarFrame.height
+    let navBarFrame = CGRect(x: 0, y: 0, width: bounds.width, height: statusBarHeight)
+    gradientLayer.frame = navBarFrame
+
+    gradientLayer.cornerRadius = layer.cornerRadius
+    return gradientLayer
+  }
+
 }
 
 private class PrivateGradientView: UIView {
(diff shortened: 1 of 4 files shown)


IBAnimatable/GradientDesignable.swift
```diff
@@ -78,14 +46,48 @@ fileprivate extension GradientDesignable where Self: UIView {
...
+    let predefinedGradient = configurePredefinedGradient()
+    if let startColor = startColor, let endColor = endColor {
+      configureGradient(startColor: startColor, endColor: endColor)
+    } else if let startColor = predefinedGradient?.start, let endColor = predefinedGradient?.end {
+      configureGradient(startColor: startColor, endColor: endColor)
+    }
+  }
+}
+
+fileprivate extension GradientDesignable where Self: UINavigationBar {
```
tbaranes: Why not keeping only one extension for `UINavigationBar` and making the methods `private`? I think it would be more readable to group all the related methods by UIKit components

IBAnimatable/GradientDesignable.swift
```diff
@@ -78,14 +46,48 @@ fileprivate extension GradientDesignable where Self: UIView {
...
+
+  func configureGradient(startColor: UIColor, endColor: UIColor) {
+    let gradientLayer = makeGradientLayer()
+    gradientLayer.colors = [startColor.cgColor, endColor.cgColor]
+    gradientLayer.configurePoints(with: startPoint)
+    let image = gradientLayer.image()
+
+    self.setBackgroundImage(image, for: .default)
```
tbaranes: Is that a mandatory `self`?

IBAnimatable/GradientDesignable.swift
```diff
@@ -78,14 +46,48 @@ fileprivate extension GradientDesignable where Self: UIView {
...
+}
+
+fileprivate extension GradientDesignable where Self: UINavigationBar {
+
+  func configureGradient(startColor: UIColor, endColor: UIColor) {
+    let gradientLayer = makeGradientLayer()
+    gradientLayer.colors = [startColor.cgColor, endColor.cgColor]
+    gradientLayer.configurePoints(with: startPoint)
+    let image = gradientLayer.image()
+
```
tbaranes: Would remove that newline, but add one before the `let image` if you wanted to separate them (up to you)

tbaranes (approved):
Beside the hard-coded constant (which I have no idea how to improve right now), and a few coding style comments, it looks good 👍 

Can you also update the example app? I think you can directly set a gradient to the navigation bar of the gradient example in playground 😬

IBAnimatable/GradientDesignable.swift
```diff
@@ -104,3 +106,48 @@ private class PrivateGradientView: UIView {
   }
 
 }
+
+extension CAGradientLayer {
```
JakeLin: should it be a `fileprivate`, will those two methods be used in this file only?

IBAnimatable/GradientDesignable.swift
```diff
@@ -104,3 +106,48 @@ private class PrivateGradientView: UIView {
...
+      self.endPoint = CGPoint(x: 1, y: 1)
+    case let .custom(start, end):
+      self.startPoint = CGPoint(x: start.x, y: start.y)
+      self.endPoint = CGPoint(x: end.x, y: end.y)
+    case .none:
+      break
+    }
+  }
+
+  func image() -> UIImage? {
```
JakeLin: should call `makeImage()` or use `var image: UIImage?` . `makeImage()` means we create a new UIImage instance. And `var image: UIImage?` means `image` is a property of `CAGradientLayer`. I more prefer `makeImage` since this image is a copy of the `CAGradientLayer` not a property of it. How do you think?

IBAnimatable/GradientDesignable.swift
```diff
@@ -104,3 +106,48 @@ private class PrivateGradientView: UIView {
...
+    UIGraphicsBeginImageContext(self.frame.size)
```
JakeLin: Can we remove all non-necessary `self`? Sorry, I am very picky on `self` stuff to try to keep the consistency of the code style.

IBAnimatable/GradientDesignable.swift
```diff
@@ -78,14 +46,48 @@ fileprivate extension GradientDesignable where Self: UIView {
...
+    gradientLayer.configurePoints(with: startPoint)
+    let image = gradientLayer.image()
+
+    self.setBackgroundImage(image, for: .default)
+  }
+
+  func makeGradientLayer() -> CAGradientLayer {
+    let gradientLayer: CAGradientLayer = CAGradientLayer()
+
+    let statusBarHeight: CGFloat = 64 // UIApplication.shared.statusBarFrame.height
```
JakeLin: Shall we use `UIApplication.shared.statusBarFrame.height` instead of `64`? is it possible to change?

IBAnimatable/GradientDesignable.swift
JakeLin: @tbaranes I think started to use the private extension to group private methods like `fileprivate extension GradientDesignable where Self: UIView`. The original idea may be from https://github.com/raywenderlich/swift-style-guide#code-organization for using the `extension`s to organise the code.

JakeLin (changes requested):
@phimage this is a great step to use the designable protocols in non `UIView` components. I didn't know how to do that before, well done 👍 👍 👍. Can you also have a look at some small comments for the code style?

IBAnimatable/GradientDesignable.swift
phimage: I think you cannot access `UIApplication.shared` here

IBAnimatable/GradientDesignable.swift
tbaranes: I fully understand the idea of grouping in extension to organise the code, and really like that practice, but in this case, we are dividing a code that will be used only from the first extension, that's why I'm wondering if we shouldn't keep together as private.

IBAnimatable/GradientDesignable.swift
phimage: yes I can
All my code have `self` when I can 😄  so when I copy it to animatable I forgot to remove it

IBAnimatable/GradientDesignable.swift
phimage: I just copy the code above for UIView, so as you want, make a choice, I have no preference

In my code I put some // MARK:   to keep it clear
here MARK: UIView . then MARK UINavigationBar

JakeLin: I got it, we try to group extension for `where Self: UIView` or `where Self: UINavigationBar`. I agree with @tbaranes now. That can save to `private extension`s. I hope Swift can support something like 
```
extension GradientDesignable where Self: UIView {
  func foo() {
  }

  private extension {
    func privateFoo() {
    }

  }
}
```

But they don't, we have to use 

```
extension GradientDesignable where Self: UIView {
  func foo() {
  }

  private func privateFoo() {
  }
}
```

@phimage How do you think we use @tbaranes 's practice to group the extensions?

IBAnimatable/GradientDesignable.swift
JakeLin: I have a second thought, do they support orientation changed? If you yes, it is perfect 💪

phimage: @JakeLin unfortunately not
But I will fix it right now by doing gradient in `layoutSubviews`

IBAnimatable/GradientDesignable.swift
phimage: I will group in one extension

JakeLin (approved):
Sorry for the late reply, all good, well done 👍👍👍

JakeLin:
@phimage can you please merge it when you think it is ready? 🚢🇮🇹 Well done. A great step to extend the protocols to non-UIView components. 🎉🎉🎉 I learned how to do it as well 👍

IBAnimatable/GradientDesignable.swift
```diff
@@ -78,16 +46,46 @@ fileprivate extension GradientDesignable where Self: UIView {
...
+  public func configureGradient() {
+    let predefinedGradient = configurePredefinedGradient()
+    if let startColor = startColor, let endColor = endColor {
+      configureGradient(startColor: startColor, endColor: endColor)
+    } else if let startColor = predefinedGradient?.start, let endColor = predefinedGradient?.end {
+      configureGradient(startColor: startColor, endColor: endColor)
+    }
+  }
+
+  fileprivate  func configureGradient(startColor: UIColor, endColor: UIColor) {
```
tbaranes, line 66: You can set them as private now that they are grouped in one extension 🙃

tbaranes (approved):

IBAnimatable/GradientDesignable.swift
phimage: configueGradient must be public. Another XxxNavigationBar could implement it

tbaranes: I agree for the `configureGradient`, I was talking about `configureGradient(startColor: UIColor, endColor: UIColor)` and `func makeGradientLayer() -> CAGradientLayer` (which are currently file private)

phimage:
Thanks for the review

PR #441, opened 2017-04-11
Fix M_PI_2 migration which was breaking transitions

### LINKED ISSUE #305: Most of transitions do not respect non translucent navigation bar leading to glitches and wrong frames.
When using some transition animations with `push`, if the Navigation bar set to `translucent=false`. The following transitions are not working properly:
- [ ] Flip
- [x] Explode
- [x] Card
- [x] Turn
- [ ] Fold

There are no problems for `present` transition or if Navigation bar set to `translucent=true`.

----- Original message ------
Hi. When navigation controller uses opaque navigation bar, translucent=false most of the transitions end up with the wrong frame for destination view, not respecting navigation bar height and UIViewControllerContextTransitioning finalFrameForViewController method.
In most cases like in Fade animation, final frame for view just wrong.
In several cases like in Flip animation it leads to UI jumps and glitches.
In rare cases like Turn left and right animation(not top and bottom, those also have glitches) it works as expected.

### LINKED ISSUE #440: Turn transition animation doesn't work. π stuff
we have in `TurnAnimator`
```swift
rotate(angle: factor * -.pi * 2)
rotate(angle: factor * .pi * 2)
```
instead of
```swift
rotate(angle: factor * -.pi / 2)
rotate(angle: factor * .pi / 2)
```

Maybe when converting the deprecated `M_PI_2` some mistake happen
I think other `.pi` must be checked

tbaranes:
Close #440 
Fix a few animators, introduced in https://github.com/IBAnimatable/IBAnimatable/pull/237/commits/10104cd0a15cfc29d0c38112d5b4b6ad04aa0a3c:

- Flip
- Turn
- Fold

Thanks to @phimage for the fix! 
@SD10 This fix the broken transition you reported a few days ago 

May also close #305 ? Need more testing to confirm

### DIFF
--- a/IBAnimatable/FlipAnimator.swift
+++ b/IBAnimatable/FlipAnimator.swift
@@ -91,7 +91,7 @@ private extension FlipAnimator {
     axesValues = valuesForAxe(initialValue: reverse ? 1.0 : 0.0, reverseValue: 0.5)
     updateAnchorPointAndOffset(anchorPoint: CGPoint(x: axesValues.0, y: axesValues.1), view: flippedSectionOfToView)
 
-    flippedSectionOfToView.layer.transform = rotate(angle: reverse ? .pi * 2 : -.pi * 2)
+    flippedSectionOfToView.layer.transform = rotate(angle: (reverse ? .pi : -.pi) / 2)
     return ((flippedSectionOfFromView, flippedSectionOfFromViewShadow), (flippedSectionOfToView, flippedSectionOfToViewShadow))
   }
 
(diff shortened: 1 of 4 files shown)


tbaranes:
After a few testing for #305, it seems to fix `Turn`, but also  `Explode` (so not related to this PR).
The `Flip` and `Fold` have still a frame issues when the navigation bar is opaque (`maxY` or `height` is wrong), but the animation is much better than before since it's working 😬

IBAnimatable/FlipAnimator.swift
```diff
@@ -173,7 +173,7 @@ private extension FlipAnimator {
                              completion: @escaping AnimatableCompletion) {
     UIView.animateKeyframes(withDuration: transitionDuration, delay: 0, options: .layoutSubviews, animations: {
       UIView.addKeyframe(withRelativeStartTime: 0.0, relativeDuration: 0.5, animations: {
-        flippedSectionOfFromView.0.layer.transform = self.rotate(angle: self.reverse ? -.pi * 2 : .pi * 2)
+        flippedSectionOfFromView.0.layer.transform = self.rotate(angle: (self.reverse ? -.pi : .pi) / 2)
```
JakeLin, line 176: Do we need the `self` here?

JakeLin (approved):
Sorry, must be me introduced those bugs. All good, thanks for fixing them 👍👍👍

IBAnimatable/FlipAnimator.swift
SD10: Yeah we do but #446 removes all of them where not necessary. I could've missed some but I think I got them all 😛

JakeLin: @SD10 thanks for taking care of `self`.

JakeLin:
> The Flip and Fold have still a frame issues when the navigation bar is opaque (maxY or height is wrong)

@tbaranes I merged the PR now. I think we saw this issue before when we have a navigation bar, I looked at it before and didn't find a solution for fixing it.

PR #442, opened 2017-04-11
Fix pre-iOS10 compilation: replace iOSApplicationExtension by iOS #no-public-changes

tbaranes:


tbaranes:
I'm going ahead and merging since it should be ok for everyone, hope you don't mind. 
I need it because Travis won't passed anymore in any projects using IBAnimatable 😨

PR #443, opened 2017-04-11
Add AnimatableTabBarController to support custom transition animation

phimage:
Same as `AnimatableNavigationController` with reverse animation when we can using child controller index

Added to the demo too, near the other tab bar controller where the animation is coded, not configured using storyboard

### DIFF
--- a/IBAnimatable/Navigator.swift
+++ b/IBAnimatable/Navigator.swift
@@ -6,7 +6,7 @@
 import UIKit
 
 /**
- Navigator for `UINavigationController` to support custom transition animation for Push and Pop
+ Navigator for `UINavigationController` or `UITabBarController` to support custom transition animation
  */
 public class Navigator: NSObject {
   var transitionAnimationType: TransitionAnimationType
@@ -43,6 +43,7 @@ public class Navigator: NSObject {
   }
 }
 
+// MARK: - navigation controller delegate
 extension Navigator: UINavigationControllerDelegate {
   // MARK: - animation controller
   public func navigationController(_ navigationController: UINavigationController,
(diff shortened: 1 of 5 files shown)


IBAnimatable/Navigator.swift
```diff
@@ -74,3 +75,43 @@ extension Navigator: UINavigationControllerDelegate {
...
+
+    guard let viewControllers = tabBarController.viewControllers,
+      let fromVCIndex = viewControllers.index(of: fromVC),
+      let toVCIndex = viewControllers.index(of: toVC) else {
+        return nil
+    }
+    if toVCIndex < fromVCIndex {
+      animator?.transitionDuration = transitionDuration
+      return animator
+    } else {
```
tbaranes: The else is not mandatory since the if returns

IBAnimatable/Navigator.swift
```diff
@@ -74,3 +75,43 @@ extension Navigator: UINavigationControllerDelegate {
...
+    }
+  }
+
+  // MARK: - interaction controller
+  public func tabBarController(_ tabBarController: UITabBarController,
+                               interactionControllerFor animationController: UIViewControllerAnimatedTransitioning)
+    -> UIViewControllerInteractiveTransitioning? {
+      if let interactiveAnimator = interactiveAnimator, interactiveAnimator.interacting {
+        return interactiveAnimator
+      } else {
```
tbaranes: Same here, you can drop the else (if that's a copy from elsewhere in the code, you can remove it there as well)

tbaranes (changes requested):
Looks good, I just let you a small a few code style comments 👍 

Can you also update the CHANGELOG?

tbaranes:
_The following comments is not directly related to this PR_

I'm wondering, should we move the both tabBar examples in the `Transition` part (by diving in 3 subnavigations) directly?

IBAnimatable/Navigator.swift
phimage: Yes it's a copy paste like always, I am a copy past coder 😉 

here I think I could do
```swift
interactiveAnimator?.interacting
```
edit/ my mistake I don`t watch carefully the code. I will just remove the else

IBAnimatable/Navigator.swift
phimage: I can simplify the code here, and I will remove unnecessary else if there is some return

IBAnimatable/Navigator.swift
tbaranes: Even better! Best way to conform to designable protocols: copy 😬

tbaranes (approved):

JakeLin (approved):
Great PR. Animatable everything 💪

JakeLin:
@phimage Great PR 👍👍👍, can you merge it when you feel it is ready?

Issue #445, opened 2017-04-11
Remove unnecessary `self.` if possible


JakeLin:
It is a controversial topic to use `self.` or not within the class, struct or extension. But in our project, we try to not use unnecessary `self.` if possible.

## What is unnecessary `self.`?
If we remove the `self.` , the code still can compile correctly. We call them unnecessary `self.`. For example within a subclass of `UIView`. `self.bounds =.zero` is same as `bounds = .zero`. `self.addSubview(aView)` is same as `addSubview(aView)`.

## Which `.self` can be removed?
If we use the `.self` within a closure, then we can't remove it.

## How to do it?
Search `self.` and see whether we can remove them.

SD10:
I would love to do this it has been silently KILLING me 😃

SD10:
Resolved in #446

PR #446, opened 2017-04-12
Remove self where not necessary

SD10:
### Summary of Pull Request:
Remove use of `self` where not necessary, #445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
 - `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
+- Transition Animator classes initializer parameter changed from transitionDuration -> duration
+[#446](https://github.com/IBAnimatable/IBAnimatable/pull/446) by [@SD10](https://github.com/sd10)
 
 #### Enhancements
 - Conserve custom layer mask when using `Animatable*` instead of removing them
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -133,7 +133,7 @@ public extension Animatable where Self: UIView {
   func autoRunAnimation() {
     if autoRun {
       autoRun = false
-      animate(self.animationType)
+      animate(animationType)
     }
   }
 }
@@ -453,7 +453,7 @@ fileprivate extension Animatable where Self: UIView {
     var scaleY: CGFloat = 1
 
     var frame: CGRect
-    if let window = self.window {
+    if let window = window {
       frame = window.convert(self.frame, to: window)
     } else {
       frame = self.frame
@@ -579,7 +579,7 @@ fileprivate extension Animatable where Self: UIView {
   }
 
   var screenSize: CGSize {
-    return self.window?.screen.bounds.size ?? CGSize.zero
+    return window?.screen.bounds.size ?? CGSize.zero
   }
 }
 // swiftlint:enable variable_name_min_length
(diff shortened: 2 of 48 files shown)


IBAnimatable/Animatable.swift
```diff
@@ -579,7 +579,7 @@ fileprivate extension Animatable where Self: UIView {
   }
 
   var screenSize: CGSize {
-    return self.window?.screen.bounds.size ?? CGSize.zero
+    return window?.screen.bounds.size ?? CGSize.zero
```
JakeLin: I think we can remove `CGSize` and just use `.zero`.

IBAnimatable/FlipAnimator.swift
```diff
@@ -21,14 +21,13 @@ public class FlipAnimator: NSObject, AnimatedTransitioning {
   fileprivate var horizontal: Bool = false
 
   // MARK: - Life cycle
-  public init(from direction: TransitionAnimationType.Direction, transitionDuration: Duration) {
+  public init(from direction: TransitionAnimationType.Direction, duration: Duration) {
```
JakeLin, line 24: That will introduce breaking changes. Please update the CHANGELOG.md

JakeLin (approved):
Well done @SD10

JakeLin:
@SD10 thanks for fixing that in no time ⚡️⚡️⚡️, you rock 🚀. And please have a look at the comment on CHANGELOG.md.

IBAnimatable/Animatable.swift
SD10: Done 👍

IBAnimatable/FlipAnimator.swift
SD10: Do I just need to reference this Pull Request or should I actually state the parameter name change? I updated the CHANGELOG but let me know if I should change it. I'm learning what an API breaking change is now 😃

tbaranes (approved):

IBAnimatable/FlipAnimator.swift
tbaranes: I don't think you should add an entry for the change of self. since it's only internal changes, but you should add an entry in the API breaking change to explain the prototype change transitionDuration -> duration because that may break for the users.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
 - `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
+- Remove the use of `self` where not necessary
```
JakeLin: @SD10 sorry, I don't think we need to put the title as "Remove the use of `self` where not necessary" for breaking changes, because it doesn't make sense for the user. I meant we should specify the API changes. something like Change the initializer for `PortalAnimator`, `NatGeoAnimator`, `TurnAnimator`, `CardsAnimator`, `FlipAnimator`, `SlideAnimator`, `ZoomAnimator` and `DropDownAnimator`. Rename the parameter from `transitionDuration` to `duration`. 

And please add `by [@SD10](https://github.com/sd10)` as well.

IBAnimatable/FlipAnimator.swift
SD10: Ok I did this plus changed the remaining 3 that were not changed as to keep the API consistent. Thank you for clarifying @tbaranes

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 - Replace all `SystemAnimator` classes with `SystemTransitionAnimator`
 [#427](https://github.com/IBAnimatable/IBAnimatable/pull/427) by [@SD10](https://github.com/sd10)
 - `PresentationDesignable` now supports `contextFrameForPresentation` which allow you to present a controller with a custom configuration over another instead of being in fullscreen. Imitates `UIModalPresentationStyle.currentContext`
+- transition Animator classes initializer parameter changed from transitionDuration -> duration
```
tbaranes: One last point: can you uppercase the first character and credit yourself?

Issue #434, opened 2017-04-06
Add unit tests


tbaranes:
As suggested by @SD10 on slack, adding tests can be a great move. Creating this issue to keep a track on this!

@SD10 I'm assigning this to you since you wanted to give it a first try (to avoid that someone else starts something), but don't hesitate to unassigned yourself if you are not interesting in anymore, not enough time... or for whatever reason is it

SD10:
I think it's highly unlikely that anyone wants to take a stab at this issue, but if you do please get in touch with me first as I've already begun this process.

SD10:
## Current Coverage: - 6.64%
### Protocols:
- [X] CornerDesignable
- [X] FillDesignable
- [X] SliderImagesDesignable
- [X] PaddingDesignable

PR #447, opened 2017-04-12
Refactoring travis #no-public-changes

tbaranes:
I wanted to add iOS9 compilation to travis to avoid merge of PR that doesn't compile on that version (cf #442), but in the current state of travis, I think it was too complicated to add a simple platform. So I went ahead, and refactor it. Now travis may looks more complicated, but in fact it is much simpler!

- Grouped the `matrix` which means every updates of the build platform. That also works when we want to add a new platform
- Avoid duplicating schemes. Our schemes are the same whatever the build is 
- Simplify the process to add a new build in the `script` phase. Starting now, variables are available which means, adding a new build phase will simply ask an update for the xcodebuild's properties, but the `IBAnimatible` related part won't change (thinking of #434 )
- Build the framework and the app separately for each platforms
- ...

Anyway, I think that's a good move, let's see what travis says!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,31 @@
-osx_image: xcode8
 language: objective-c
-before_install: 
+osx_image: xcode8.3
+env:
+  global:
+    - LC_CTYPE=en_US.UTF-8
+    - LANG=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


JakeLin (approved):
Cool, thanks for optimising the CI 👍

PR #449, opened 2017-04-12
Put back a useful self. in GradientDesignable #no-public-changes

tbaranes:
or crash violently 😡

May be worth to take a second look at #446 just in case there was others like this one?

ÉDIT: Just did a second review of it, the others seem ok to me

### DIFF
--- a/IBAnimatable/GradientDesignable.swift
+++ b/IBAnimatable/GradientDesignable.swift
@@ -34,7 +34,7 @@ fileprivate extension GradientDesignable where Self: UIView {
       $0.removeFromSuperview()
     }
 
-    let gradientView = PrivateGradientView(frame: bounds, layer: gradientLayer)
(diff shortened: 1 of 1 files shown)


IBAnimatable/GradientDesignable.swift
```diff
@@ -89,7 +89,7 @@ private class PrivateGradientView: UIView {
 
   init(frame: CGRect, layer: CAGradientLayer) {
     super.init(frame: frame)
-    layer.insertSublayer(layer, at: 0)
+    self.layer.insertSublayer(layer, at: 0)
```
JakeLin: next time, we may use `gradientLayer` for the parameter instead of the `layer` which is same as the property.

JakeLin (approved):
Sorry, I didn't look at this one carefully.

tbaranes:
We all miss this one.
I renamed the parameter, that's a good move

Issue #436, opened 2017-04-07
FatalError: "No opposing HollowState for HollowState.none" in playground


phimage:
When trying to see all transition animation available I try systemCameraIris 

maybe related to #427 because we get in `SystemTransitionAnimator`

```swift
self.reverseAnimationType = transitionAnimationType.reversed
```

Then in playground the definition is
```swift
    transitionAnimations.append(["SystemCameraIris", "SystemCameraIris(HollowOpen)", "SystemCameraIris(HollowClose)"])
```
instead of 
```swift
    transitionAnimations.append(["SystemCameraIris", "SystemCameraIris(Open)", "SystemCameraIris(Close)"])
```
so we have always `none` type and the `fatalError`

tbaranes:
Oh you are right, I think we should drop the hard code string from the example app in order to avoid this kind of issues. If I remember well the only reason that we had strings instead of the enum type was to set the title in the navigation bar

JakeLin:
I think we used the string values to demonstrate how can we use those values in Interface Builder. If you have a toString method for the enum. We should always use enum in code. Since @SD10 is working on refactoring. The compiler won't give any error on those changes.

SD10:
Ok I'm closing this because it has been taken care of by #438

JakeLin:
@SD10 thanks for taking care of that.

PR #450, opened 2017-04-14
Fix activity indicator broken animation #no-public-changes

tbaranes:
Instead of reintroducing the `self.`, I renamed the variable to be sure that it won't compile.
I didn't test them all, just the broken one in my app. I think we should fully retest the playground app to be sure there's no regression.

### DIFF
--- a/IBAnimatable/ActivityIndicatorAnimationBallGridBeat.swift
+++ b/IBAnimatable/ActivityIndicatorAnimationBallGridBeat.swift
@@ -21,6 +21,7 @@ public class ActivityIndicatorAnimationBallGridBeat: ActivityIndicatorAnimating 
     let durations = [0.96, 0.93, 1.19, 1.13, 1.34, 0.94, 1.2, 0.82, 1.19]
     let beginTime = CACurrentMediaTime()
     let beginTimes = [0.36, 0.4, 0.68, 0.41, 0.71, -0.15, -0.12, 0.01, 0.32]
+    let animation = defaultAnimation
 
     // Draw circles
     for i in 0 ..< 3 {
@@ -45,7 +46,7 @@ public class ActivityIndicatorAnimationBallGridBeat: ActivityIndicatorAnimating 
 
 private extension ActivityIndicatorAnimationBallGridBeat {
 
-  var animation: CAKeyframeAnimation {
+  var defaultAnimation: CAKeyframeAnimation {
     let animation = CAKeyframeAnimation(keyPath: "opacity")
     animation.keyTimes = [0, 0.5, 1]
     animation.timingFunctions = [timingFunction, timingFunction]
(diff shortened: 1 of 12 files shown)


JakeLin (approved):
Thanks for fix that.

JakeLin:
@tbaranes thanks for fixing that.

@SD10 since we have a comprehensive demo app (IBAnimatable Playground). Can you please use it for checking when you do the refactor? I suggest we create a small UI Automation target to run this demo. We don't need to cover everything at the first time. For example, we add one for ActivityIndicators. Then can help us check the before and after when we do refactor.

SD10:
@tbaranes @JakeLin Sorry for the trouble! 😱

To be honest I shouldn't have made those changes because they were unrelated to removing `self`. I just removed all these while going through the files because I thought they were unnecessary assignments. I thought all animation code was encapsulated in the `animation` property. Renaming the property to `defaultAnimation` is even more clear.

JakeLin:
@SD10 As a human being, we all make mistakes. Good software practice can prevent us from making mistakes. For example, using the enum instead of string values. Having unit tests for refactoring. As we discussed that before, I don't know the best practice for unit testing UI. We developed a demo app (IBAnimatable Playground) to test our features. We can use it for testing new features, fixing bugs and doing refactor. I know it is not straight forward and we have to check them manually. The first step to automate it is to create some [UI Testing in Xcode](https://developer.apple.com/videos/play/wwdc2015/406/). And we can add it to CI to check whether we break something. Although we still can't check whether we break the presentation of the UI. But we can pick up some issue like #436 

@SD10 if you are still doing the factor for Activity Indicators. Can you have a look at  [UI Testing in Xcode](https://developer.apple.com/videos/play/wwdc2015/406/), and see wheter we can add a test for activity indicators. There is VC with a picker. We can select the picker automatically to check all activity indicators.

tbaranes:
UI tests will be useful for a few cases like the gradient crash, but won't help at all here. We can't check if an animation is running as expected, but it will fails if the app crash.

JakeLin:
Yes, if we integrate it with CI, it can only find crashes. I don't know the best practice for automatically testing UI. I always do it manually because it is hard to say it right or wrong for animations unless we see it. But it may help a bit besides crashes, we can side back and see the test runs automatically without tapping on it.

I put up a card #453, happy to chat whether we should do it or not.

SD10:
@jakelin Yeah I've already been looking at UI Testing. I'm thinking we can have a test that runs each section of the demo app (activity indicators, transition animations, etc.) and then one main test that runs all the subtests. Then whenever we want to test something we could just watch the test to make sure everything is performing properly. 

The problem is the only thing this saves us is some clicking. I definitely think it's worth implementing to increase productivity but UI testing wouldn't have caught this issue. The main problem is I didn't do the testing to begin with because I thought I was making the most trivial of changes.

PR #456, opened 2017-04-21
Support timing function for animation. Fix #12

### LINKED ISSUE #12: Support animations with timing functions
Currently, all animations are Spring animations, can support Easing animations as well.

phimage:
- Timing function enum added
  - Add it to `Animatable` objects and `AnimationConfiguration`
  - Use this enum when possible instead of `CAMediaTimingFunction`

TODO
- [x] add more type in enum (see `extension CAMediaTimingFunction`)
- [x] add more type in enum  see `CAMediaTimingFunction(` contructors in code
- [x] add custom type with control points in enum ?
- [x] manage spring animation or not, animation default value
- [x] Add an example
- [x] Update CHANGELOG
- [x] Documentation

### DIFF
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -44,9 +44,14 @@ public protocol Animatable: class {
    */
   var force: CGFloat { get set }
 
+  /**
+   Animation function as a timing curve. (default value should be none)
+   */
+  var timingFunction: TimingFunctionType { get set }
+
 }
 
-public extension Animatable where Self: UIView {
+public extension Animatable {
   public func configureAnimatableProperties() {
     // Apply default values
     if duration.isNaN {
@@ -65,7 +70,9 @@ public extension Animatable where Self: UIView {
       force = 1
     }
   }
+}
 
+public extension Animatable where Self: UIView {
   @discardableResult
   public func animate(_ animation: AnimationType,
                       duration: TimeInterval? = nil,
@@ -299,7 +306,7 @@ fileprivate extension Animatable where Self: UIView {
       let animation = CAKeyframeAnimation(keyPath: "position.x")
       animation.values = [0, 30 * configuration.force, -30 * configuration.force, 30 * configuration.force, 0]
       animation.keyTimes = [0, 0.2, 0.4, 0.6, 0.8, 1]
-      animation.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+      animation.timingFunctionType = configuration.timingFunction ?? .easeInOut
       animation.duration = configuration.duration
       animation.isAdditive = true
       animation.repeatCount = Float(repeatCount)
@@ -313,7 +320,7 @@ fileprivate extension Animatable where Self: UIView {
       let animation = CAKeyframeAnimation(keyPath: "transform.scale")
       animation.values = [0, 0.2 * configuration.force, -0.2 * configuration.force, 0.2 * configuration.force, 0]
       animation.keyTimes = [0, 0.2, 0.4, 0.6, 0.8, 1]
-      animation.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+      animation.timingFunctionType = configuration.timingFunction ?? .easeInOut
       animation.duration = configuration.duration
       animation.isAdditive = true
       animation.repeatCount = Float(repeatCount)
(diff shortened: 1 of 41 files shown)


IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,210 @@
...
+      self = .easeIn
+    case "easeout":
+      self = .easeOut
+    case "easeinout":
+      self = .easeInOut
+    case "default":
+      self = .default
+
+    // customs
+    case "easeInSine":
```
lastMove: It's a little bit confusing but `extractNameAndParams` returns only lowercased string (I will add it to the function documentation) so the `case` have to be lowercase too. (It's this way so case is ignored via InterfaceBuilder

phimage:
added more types using sourcery to generate switch case from static variables in `CAMediaTimingFunction`

 stencil template 
```
{% for type in types.all %}

{% for variable in type.variables %}
case {{ variable.name }}
{% endfor %}

{% for variable in type.variables %}
case .{{ variable.name }}:
return .{{ variable.name }}
{% endfor %}

{% for variable in type.variables %}
case "{{ variable.name|lowercase }}":
self = .{{ variable.name }}
{% endfor %}

{% endfor %}
```

IBAnimatable/Animatable.swift
```diff
@@ -44,9 +44,14 @@ public protocol Animatable: class {
...
 
+  /**
+   Animation function as a timing curve. (default value should be none)
+   */
+  var timingFunction: TimingFunctionType { get set }
+
 }
 
-public extension Animatable where Self: UIView {
+public extension Animatable {
```
lastMove, line 54: Nice one

IBAnimatable/Animatable.swift
```diff
@@ -327,12 +334,12 @@ fileprivate extension Animatable where Self: UIView {
...
       squashX.values = [1, 1.5 * configuration.force, 0.5, 1.5 * configuration.force, 1]
       squashX.keyTimes = [0, 0.2, 0.4, 0.6, 0.8, 1]
-      squashX.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+      squashX.timingFunctionType = .easeInOut
 
       let squashY = CAKeyframeAnimation(keyPath: "transform.scale.y")
       squashY.values = [1, 0.5, 1, 0.5, 1]
       squashY.keyTimes = [0, 0.2, 0.4, 0.6, 0.8, 1]
-      squashY.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+      squashY.timingFunctionType = .easeInOut
```
lastMove: Should allow to customize the timingFunction for squash animation

IBAnimatable/Animatable.swift
```diff
@@ -327,12 +334,12 @@ fileprivate extension Animatable where Self: UIView {
       let squashX = CAKeyframeAnimation(keyPath: "transform.scale.x")
       squashX.values = [1, 1.5 * configuration.force, 0.5, 1.5 * configuration.force, 1]
       squashX.keyTimes = [0, 0.2, 0.4, 0.6, 0.8, 1]
-      squashX.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+      squashX.timingFunctionType = .easeInOut
```
lastMove: Should allow to customize the timingFunction for squash animation

lastMove (dismissed):
Great work ! 
It's almost OK for me.
It could be useful to add the possibility to choose the timingFunction in the "Animations" Section of  the IBAnimatable App.

IBAnimatable/Animatable.swift
phimage: Forgot to remove this change for this PR
(Trying to make `UIBarItem` also `Animatable`, I see that I can remove that)

phimage:
default value managed using operator `??` to replace `none`with a default one at the right of the operator

I think I made a fix in `ActivityIndicatorAnimationOrbit` ring2OpacityAnimation
same could be done in ring1OpacityAnimation


For the demo
maybe as @lastMove say in the "Animations" Section 
or a new one with
- two translation animations
- two or one timing function type menu selectors
this allow to compare two timing functions (could compare always with linear if only one menu selector)

lastMove:
> a new one with
two translation animations
two or one timing function type menu selectors
this allow to compare two timing functions (could compare always with linear if only one menu selector)

 Nice Idea !
It's a great first step.
The inconvenient of this solution is that we can see the effect of the timing function only on Translations.

phimage:
Maybe I will do the two case
1 - in Animations add a button to display a new view controller to select the timing function
screen shot 2017-05-07 at 11 54 28
I am working on it today

2 - add a new entry with two elements, one animation (maybe later an animation type selector in a new view controller)
a picker for timing function

phimage:
mediatimingfunction
added in "Animations" a view to select the Timing Function as illustrated in the previous message

EDIT: added two falling stars to see effect of Timing Function
Also moveBy and moveTo could now use cubic animation with timing function instead of spring animation if timing function defined

Need some help for the Documentations, where to document it?
- APIs.md there is the new Animatable protocol attribute

tbaranes:
@phimage Sorry for the delay, do you still need help for the documentation?

Issue #451, opened 2017-04-15
TextField Placeholder / Image Programatically 


DonBaronFactory:
Hi, I'm facing a weird problem. When I create an AnimatableTextField via IB then all works as expected. In contrast when I create the same Textfield programatically and apply a left image and placeholder, then the latter don't get displayed... 

```
var emailField: AnimatableTextField = {
        var t = AnimatableTextField()
        t.translatesAutoresizingMaskIntoConstraints = false
        t.borderWidth = 1
        t.borderColor = .white
        t.borderSides = .bottom
        t.leftImage = #imageLiteral(resourceName: "email_text")
        t.placeholderColor = .white
        t.placeholderText = "eMail Address"
        t.tintColor = .white
        t.keyboardType = .emailAddress
        return t
    }()
```
Any Idea what I'm doing wrong?

PS: the Borders get displayed properly

phimage:
you do it when? as class attribute? in viewDidLoad of your controller?

sometime user interface code don`t support to be at class initialization
you can make your variable lazy but one time the bug could come back

DonBaronFactory:
I do create these fields as class attributes in a custom UIView... but from what I understand this may be a bug right?

phimage:
I think you could change your code to fix it for the moment

And for IBAnimatable, maybe `configureImages` must be called in `configureAfterLayoutSubviews` and not `configureInspectableProperties`

---

how/when did you add the emailField to the CustomView?

to add an image to the textfield I think component bounds or frame is needed
and if your component is not in view hierarchy or If you don't set a frame size as init then there is some issue

try to set a frame in init of your TextField
or best setup your component after adding it to the view hierarchy

DonBaronFactory:
will be trying out your suggestions... I basically add the textfield to my view (addSubView) during the init of my custom class. The frame as such might be zero, nonetheless I do constrain it in my code using AutoLayout and anchors

DonBaronFactory:
I did manage to make it work by calling `configureImages` after setting the left image of the textfield and its padding... @phimage you mentioned "for the moment", does that mean you'll be fixing it in a future release? if yes, we should mark this issue as bug / enhancement

phimage:
I have no time to reproduce it at the moment so I am not sure about if it is a real bug or wrong use of IBAnimatable
I ask question just to help and to allow someone to reproduce ;-)

tbaranes:
[...]

PR #457, opened 2017-04-21
Fix UITextField borderStyle when no borders set

tbaranes:
That's a bit tricky: in the current configuration, if using `AnimatableTextField` without setting a custom border, the `borderStyle` will simply set to `.none` even if we want to keep it. I didn't find a way to fix it using `borderSides`, so I used `borderWidth` instead. Due to that situation, placeholders won't be displayed anymore too.

<strike>I need more testing to confirm but may be related to #451, the issue there is different since borders are used</strike>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ N/A
 
 #### Bugfixes
 
-N/A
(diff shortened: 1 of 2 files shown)


tbaranes:
Travis failing due to iOS 9.3, device not found. Not sure why it's suddenly breaking

tbaranes:
@JakeLin Can you take a look at this? 🙃

JakeLin (approved):
You are right, borderStyle should not be set unconditionally. Relaying on borderWidth should be right because we can't apply custom border sides if border width is zero.

PR #462, opened 2017-04-26
[WIP] Fix FillDesignable bugs

### LINKED ISSUE #461: backgroundColor for AnimatableStackView not set by FillDesignable
### Issue
backgroundColor for `AnimatableStackView` not set by `FillDesignable` protocol and remains `nil`

Apparently `UIStackView` is a non-drawing view so `drawRect()` is never called and the `backgroundColor` is not applied.

EDIT: This obviously affects setting the `opacity` as well

### Possible Solutions
We can possibly solve this by:
1) Adding a backing `CAShapeLayer` with the background color in `layoutSubviews()`
2) Adding a backing `UIView` with the background color in `layoutSubviews()`

This is not my domain of expertise so I thought I would bounce some ideas off the rest of you.

SD10:
### Current Changes (Work In Progress):

1. Fixes `isOpaque` flag that is never getting set because it defaults to `false`

2. Sets `backgroundColor` of `contentView` for `UICollectionViewCell`
We have this behavior for `UITableViewCell` so I assume we want it on `UICollectionViewCell`

I would also like to fix #461 and complete testing before merging this in.

I figure this would be the best place to post and discuss issues rather than spam the chat over every little minor detail.


tbaranes:
I'm surprised this is compiling, `UICollectionViewCell` doesn't have a `contentView` as `UITableViewCell`. How is it working? 🤔

SD10:
I see a `contentView` property for `UICollectionViewCell` in the header file

JakeLin:
@SD10 yes, it is
```
open class UICollectionViewCell : UICollectionReusableView {

    
    open var contentView: UIView { get } // add custom subviews to the cell's contentView

```

I assume what's on top of @tbaranes is why do we use `contentView.layer` instead of `layer` for `AnimatableCollectionViewCell`. Because we can treat `AnimatableTableViewCell` same as the other `UIView` subclasses.

JakeLin:
@mmadjer Do you remember why did we implement `configureCornerRadius` differently for `AnimatableCollectionViewCell`?

IBAnimatable/FillDesignable.swift
```diff
@@ -44,6 +42,18 @@ public extension FillDesignable where Self: UITableViewCell {
   }
 }
 
+public extension FillDesignable where Self: UICollectionViewCell {
+  public func configureFillColor() {
+    if let fillColor = fillColor {
+      backgroundColor = fillColor
+      contentView.backgroundColor = fillColor
```
JakeLin, line 49: Look like this is a case we keep `fillColor` 😁

JakeLin (approved):
👍

mmadjer:
@JakeLin, because for `UIColletionViewCell` you have to use the layer of the `contentView`. Is this what you're after?

JakeLin:
@mmadjer you are right, we should use `contentView` for both `UICollectionViewCell` and `UITableViewCell`.

JakeLin:
@SD10 I tried something to eliminate the code

```
public protocol CellFillDesignable: class {
  var backgroundColor: UIColor? { get set }
  var contentView: UIView { get }
}

extension UITableViewCell: CellFillDesignable {}
extension UICollectionViewCell: CellFillDesignable {}

public extension FillDesignable where Self: CellFillDesignable {
  public func configureFillColor() {
    if let fillColor = fillColor {
      backgroundColor = fillColor
      contentView.backgroundColor = fillColor
    } else if let predefinedColor = predefinedColor?.color {
      backgroundColor = predefinedColor
      contentView.backgroundColor = predefinedColor
    }
  }
}
```

But I ended up the error like `/Users/jake.lin/dev/Jake/IBAnimatable/IBAnimatable/AnimatableTableViewCell.swift:35:7: Ambiguous use of 'configureFillColor()'` for `AnimatableTableViewCell` and `AnimatableCollectionViewCell` because they don't know which to use, there are two `configureFillColor` methods for them.

SD10:
@JakeLin yeah I thought about this but it's tricky. I think most solutions I came up with did more harm than good...

e.g) polluting UIKit API, a new class, changing access control. 

The object graph for this project is huge... I'm trying to limit introducing new types of any kind :(

Right now my personal policy is not to introduce a new type unless it replaces two existing types. With the exception of new features of course.

JakeLin:
@SD10 I think your solution is acceptable. let's do it like this now.

JakeLin:
@SD10 I think we should merge this PR first, then rebase the `testing/unit-testing` branch from `master`. Then we can review the changes trunk by trunk.

PR #463, opened 2017-04-27
Add class only requirement to Designable protocols

SD10:
### Summary of Pull Requests:
Changes the `Designable` protocols to be `class` only protocols

How should we handle the CHANGELOG entry? 

Technically this is an API Breaking change if someone was using these protocols on a `struct` for some reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 
 #### API breaking changes
 
-N/A
+- Designable protocols have been changed to `class` only protocols. [#463](https://github.com/IBAnimatable/IBAnimatable/pull/463) by [@SD10](https://github.com/SD10)
 
 #### Enhancements
 
--- a/IBAnimatable/ActivityIndicatorAnimating.swift
+++ b/IBAnimatable/ActivityIndicatorAnimating.swift
@@ -6,7 +6,7 @@
 import UIKit
 
 /// Protocol for all activity indicator classes.
-public protocol ActivityIndicatorAnimating {
+public protocol ActivityIndicatorAnimating: class {
   /**
    Define the animation for the activity indicator.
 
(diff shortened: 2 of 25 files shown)


JakeLin (approved):
👍

JakeLin:
I don't think we can use those protocol in `struct`, it may have some case like we have to use `var view = view`. But that's not a breaking change. It is an improvement, right? 😂

SD10:
I meant if a user was using our library like this:
```struct MyCustomAnimatableView: FillDesignable {}```
The protocols are public so people can access them directly if they want.
Idk why someone would use a `struct` for a UI element... but it IS possible 😅

JakeLin:
I don't think someone would do that. If you think adding a breaking change will be safer, please go ahead, but only affect someone using it in a non-common way.

tbaranes:
What's the advantage to do this?
(I'd like to approve it but can't do it on the application 😢)

SD10:
@tbaranes There's no real advantage but generally UI elements should be classes.
When I'm writing Unit Tests I'm treating the element I'm testing as the protocol type.

However, the compiler can't ensure that the element is a reference type.
Therefore, I need to create a local copy of the element inside the scope of the test to modify it.

The problem is the compiler then gives me the warning to change `var` to `let` since it knows
it's a reference type and it was never modified.

If I change it to `let` the code won't compile. It says "cannot modify a let constant".

So I'm left with either using `let` and the code won't compile. 
Or using `var` and the code being littered with yellow warnings.

If you take a look at some of the old Travis build logs you will see these warnings.

EDIT:

This is currently failing because it broke something in the Demo App. I'll update it soon.

SD10:
@JakeLin this is ready to go if it's all good with you.

JakeLin (approved):
👍

JakeLin:
@SD10 yes, merged.

PR #467, opened 2017-04-30
Unit Testing for CornerDesignable & Fill Designable

SD10:
### Testing:
- Adds unit tests for `CornerDesignable` protocol
- Adds unit tests for `FillDesignable` protocol
- Updates TravisCI to run the added tests
- Updates Travis for CodeCov

~~### Changes:~~
~~- Fixes isOpaque flag that is never getting set because it defaults to false~~
~~- Sets backgroundColor of contentView for UICollectionViewCell~~
(Already merged)

After this initial PR, I'll start creating a new PR for each class or protocol that is fully tested.
I'm closing #462 in favor of this, as all the discussed changes are included here.

Should I add testing under the CHANGELOG? Possibly under enhancements?
I'll add an entry for the change of `backgroundColor` under bug fixes.


.travis.yml
```diff
@@ -20,7 +20,7 @@ script:
   - xcodebuild -showsdks
 
   # Build Framework in Release
-  - xcodebuild clean build -project "$XCODEPROJ" -scheme "$SCHEME" -destination "$DESTINATION" -configuration Release | xcpretty;
+  - xcodebuild clean build -project "$XCODEPROJ" -scheme "$SCHEME" -destination "$DESTINATION" -configuration Release ONLY_ACTIVE_ARCH=NO ENABLE_TESTABILITY=YES test | xcpretty;
```
tbaranes: You should duplicate this line, it's always good to do both building and the testing

IBAnimatableTests/ColorTypeTests.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  ColorTypeTests.swift
+//  IBAnimatable
+//
+//  Created by Steven on 4/26/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import XCTest
+@testable import IBAnimatable
```
tbaranes, line 10: Can we remove the @testable? It would help to ensure that we don't forget a public

IBAnimatableTests/CornerDesignableTests.swift
```diff
@@ -0,0 +1,122 @@
+//
+//  CornerDesignableTests.swift
+//  IBAnimatable
+//
+//  Created by Steven on 4/24/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import XCTest
+@testable import IBAnimatable
```
tbaranes, line 10: Same here

IBAnimatableTests/CornerDesignableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+    helper_testCornerRadius_UIView(element: AnimatableImageView())
+    helper_testCornerRadius_UIView(element: AnimatableLabel())
+    helper_testCornerRadius_UIView(element: AnimatableStackView())
+    helper_testCornerRadius_UIView(element: AnimatableTableViewCell())
+    helper_testCornerRadius_UIView(element: AnimatableStackView())
+    helper_testCornerRadius_UIView(element: AnimatableView())
+    helper_testCornerRadius_UICollectionViewCell(element: AnimatableCollectionViewCell())
+  }
+
+  private func helper_testCornerRadius_UIView(element: MockUIViewCornerDesignable) {
```
tbaranes, line 28: Can we divide this in two tests method? One for the radius and the other for the corner sides? Don't know what you think of this, but I generally to separate each feature tested

tbaranes, line 28: Follow up of the previous comment. We can also have the commun testing (layer, ...) here then call a `testCornerRadius` and a `testCornerSides`

IBAnimatableTests/FillDesignableTests.swift
```diff
@@ -0,0 +1,190 @@
+//
+//  FillDesignableTests.swift
+//  IBAnimatable
+//
+//  Created by Steven on 4/25/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import XCTest
+@testable import IBAnimatable
```
tbaranes, line 10: Same here

IBAnimatableTests/FillDesignableTests.swift
```diff
@@ -0,0 +1,190 @@
...
+
+class FillDesignableTests: XCTestCase {
```
tbaranes: Final? Should make the compilation faster!

IBAnimatableTests/FillDesignableTests.swift
```diff
@@ -0,0 +1,190 @@
...
+
+  // MARK: - testOpacity
+
+  func testOpacity() {
+    helper_testOpacity(element: AnimatableView())
+    helper_testOpacity(element: AnimatableScrollView())
+    helper_testOpacity(element: AnimatableButton())
+    helper_testOpacity(element: AnimatableCheckBox())
+    helper_testOpacity(element: AnimatableImageView())
+    //helper_testOpacity(element: AnimatableStackView()) // broken see #461
```
tbaranes, line 57: Should be ok, isn't it? (Can harldly check on phone, sorry if not the case)

SD10: #461 is regarding `AnimatableStackView` being a non-drawing view
So changing things like the `backgroundColor`, `opacity`, won't take effect.
My goal is to re-evaluate this class when I have higher test coverage for it.
This way we can see what else is not taking effect on it.

IBAnimatableTests/FillDesignableTests.swift
SD10: Duh! 😃

IBAnimatableTests/ColorTypeTests.swift
SD10: I'm not really sure of the difference. 
I thought that using **@testable** allows me to test internal methods of IBAnimatable.

IBAnimatableTests/CornerDesignableTests.swift
SD10: Yeah, I completely agree. I don't know why I have these two things squashed together.

IBAnimatableTests/ColorTypeTests.swift
tbaranes: That's right, if we have internal things to test, you need `@testable`, but I was thinking that we don't have any, do we?

SD10: @tbaranes I tried changing this to a normal import but it clashes with the mock protocols that I only want to give a `fileprivate` scope.

.travis.yml
SD10: Sorry @tbaranes does this line not do both the build and the tests?

IBAnimatableTests/CornerDesignableTests.swift
SD10: After going back to refactor this now I remember why it is the way that it is.

Basically, there's nothing to test for `cornerSides` that isn't already tested under `cornerRadius` because setting `cornerSides` just calls `configureCornerRadius()` behind the scenes.

SD10:
@tbaranes basically the only change I made was marking the classes as final.
If you and @JakeLin are fine with this I will merge it!

IBAnimatableTests/FillDesignableTests.swift
```diff
@@ -0,0 +1,190 @@
...
+
+  // MARK: - testFillColor
+
+  func testFillColor() {
+    helper_testFillColor_UIView(element: AnimatableView())
+    helper_testFillColor_UIView(element: AnimatableScrollView())
+    helper_testFillColor_UIView(element: AnimatableButton())
+    helper_testFillColor_UIView(element: AnimatableCheckBox())
+    helper_testFillColor_UIView(element: AnimatableImageView())
+    //helper_testFillColor_UIView(element: AnimatableStackView()) // broken see #461
```
JakeLin, line 30: May make some comment like To behavior as same as UIKit, `UIStackView` doesn't respect to `fillColor`.

JakeLin (approved):
Good to go. Well done 👍

IBAnimatableTests/FillDesignableTests.swift
SD10: That issue #461 that I referenced has all those details. 
If you want I'll add more comments but this is temporary until it is resolved.

.travis.yml
tbaranes: It does, but it's always good to do both separately to make sure everything is ok in their right configuration

tbaranes (approved):

tbaranes:
Note for future: we could add an entry to Danger in order to generate a warning when framework code has changed, but not the unit tests.

SD10:
I think that's a great idea @tbaranes. I was also thinking we could have a `#testing` header for PR's since we won't be updating the CHANGELOG but rather linking to an issue that states the current test coverage. 

The issue regarding coverage is here:
https://github.com/IBAnimatable/IBAnimatable/issues/434

If you have a better idea for documenting testing, let me know. My concern was polluting the CHANGELOG with a massive number of PR's/test updates.

tbaranes:
I'm joining you to say that we shouldn't pollute the CHANGELOG. I think it's ok since we have entry saying that we started to do unit tests. Otherwise, just a test coverage badge in the README would be sufficient, isn't it? Plus the issue of course

SD10:
Yeah, my idea was to add a badge when test coverage is higher. 
I think publicizing low test coverage would just be offputting at the moment.

PR #470, opened 2017-05-07
Replace CGFloat.pi with CGFloat(Double.pi) for ActivityIndicatorShape #no-public-changes

broadwaylamb:
I have experienced an issue that had been described here ninjaprox/NVActivityIndicatorView#138.

Since activity indicator animations in IBAnimatable have been ported from the `NVActivityIndicatorView` project, I thought that the solution had to be the same, and it worked.

This patch fixes missing activity indicator animations on iPhone 5.

P. S. Can we make a patch release if this PR is merged?

### DIFF
--- a/IBAnimatable/ActivityIndicatorShape.swift
+++ b/IBAnimatable/ActivityIndicatorShape.swift
@@ -54,7 +54,7 @@ private extension ActivityIndicatorShape {
     path.addArc(withCenter: CGPoint(x: size.width / 2, y: size.height / 2),
                           radius: size.width / 2,
                           startAngle: 0,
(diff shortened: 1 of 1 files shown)


JakeLin:
@broadwaylamb thanks for fixing that issue 👍. I remember I migrated all `.pi` manually since we ported the project in Swift 2. But I don't know `CGFloat.pi` would cause problem.

I am going to make a release like 4.0.1, Can I ask you a favor to add an item to https://github.com/IBAnimatable/IBAnimatable/blob/master/CHANGELOG.md#bugfixes because you do fix a bug. I didn't recongize when I mreged it.

PR #476, opened 2017-05-10
Add tests for RotationDesignable #no-public-changes

SD10:
### Summary of Pull Request:
- Adds tests for `RotationDesignable` conforming classes.

### DIFF
--- a/IBAnimatableTests/RotationDesignableTests.swift
+++ b/IBAnimatableTests/RotationDesignableTests.swift
@@ -1,0 +1,32 @@
+//
+//  RotationDesignableTests.swift
+//  IBAnimatable
+//
+//  Created by Steven on 5/15/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import XCTest
+@testable import IBAnimatable
+
+protocol RotationDesignableTests {
+
+  func testRotate()
+
+}
+
+// MARK: - UIView Tests
+
+extension RotationDesignableTests {
+
+  func _testRotate<E: UIView>(_ element: E) where E: RotationDesignable {
+    element.rotate = -360
+    XCTAssertEqual(element.transform, .identity)
+    element.rotate = 360
+    XCTAssertEqual(element.transform, .identity)
+    element.rotate = 90
+    let mockTransform = CGAffineTransform(rotationAngle: .pi * 90 / 180)
+    XCTAssertEqual(element.transform, mockTransform)
+  }
+
+}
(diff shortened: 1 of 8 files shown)


tbaranes (approved):

IBAnimatableTests/RotationDesignableTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    XCTAssertEqual(element.transform, .identity)
+    element.rotate = 90
+    let mockTransform = CGAffineTransform(rotationAngle: .pi * 90 / 180)
+    XCTAssertEqual(element.transform, mockTransform)
+  }
+}
+
+// MARK: - MockRotationDesignable
+
+protocol MockRotationDesignable: class, RotationDesignable {
```
tbaranes: Won't we need these mocks in others tests? For example, to test mixing designable properties. Won't be better to create a file by mock? (just wondering, up to you)

SD10: I'm not sure what you mean by creating a file by mock. Right now I'm testing each protocol and creating a mock for that protocol. One protocol per file.

As for testing mixing designable properties... basically, you're talking about one designable protocol affecting the performance of another designable protocol? That's an interesting point you bring up. I'm just testing the individual components right now... mixing them does add another degree of complexity.

tbaranes: You got my point. I think the main point to do mixing tests is your mock, it shouldn't be too complicated with your architecture (besides the complexity of the test itself). Saying that I'm wondering if we should put out each mock protocol (in their own files) to make easier to find them when we will start to work on mixing tests.

Since they are internal, it will work as expected, but it may be harder to find them in the project when we will need them in other context.

tbaranes: You got my point. I think the main point to do mixing tests is your mock, it shouldn't be too complicated with your architecture (besides the complexity of the test itself). Saying that I'm wondering if we should put out each mock protocol (in their own files) to make easier to find them when we will start to work on mixing tests.

Since they are internal, it will work as expected, but it may be harder to find them in the project when we will need them in other context.

tbaranes: You got my point. I think the main point to do mixing tests is your mock, it shouldn't be too complicated with your architecture (besides the complexity of the test itself). Saying that I'm wondering if we should put out each mock protocol (in their own files) to make easier to find them when we will start to work on mixing tests.

Since they are internal, it will work as expected, but it may be harder to find them in the project when we will need them in other context.

SD10: I see what you're saying. 
I know each file we add will add like 3-4 seconds on to the build time though!
I'm open to doing this but I think mixing tests is a little way down the road.
-- I started to write a long response regarding mixing tests... a lot going through my mind now. You just had to bring this up!

tbaranes: We can continue this discussion when we will start mixing tests, for now, it's not blocking this PR!

SD10:
#434 - marking this for record keeping

Issue #13, opened 2015-12-28
Linear animations


JakeLin:
Except for Spring animations, sometimes, we need linear animations.

phimage:
close by #456

PR #478, opened 2017-05-11
Support timing function for animation. Fix #12 #13

### LINKED ISSUE #12: Support animations with timing functions

phimage:
Re pull #456
Timing function enum added `TiminFunctionType`
  - Add it to `Animatable` objects and `AnimationConfiguration`
  - Use this enum when possible instead of `CAMediaTimingFunction`

TODO
- [x] add more type in enum (see `extension CAMediaTimingFunction`)
- [x] add more type in enum  see `CAMediaTimingFunction(` contructors in code
- [x] add custom type with control points in enum ?
- [x] manage spring animation or not, animation default value
- [x] Add an example
- [x] Update CHANGELOG
- [x] Documentation

### DIFF
--- a/IBAnimatable/ActivityIndicatorAnimationOrbit.swift
+++ b/IBAnimatable/ActivityIndicatorAnimationOrbit.swift
@@ -76,12 +76,12 @@ private extension ActivityIndicatorAnimationOrbit {
   }
 
   var coreScaleAnimation: CAKeyframeAnimation {
-    let inTimingFunction = CAMediaTimingFunction(controlPoints: 0.7, 0, 1, 0.5)
-    let outTimingFunction = CAMediaTimingFunction(controlPoints: 0, 0.7, 0.5, 1)
-    let standByTimingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear)
+    let inTimingFunction: TimingFunctionType = .custom((0.7, 0), (1, 0.5))
+    let outTimingFunction: TimingFunctionType = .custom(( 0, 0.7), (0.5, 1))
+    let standByTimingFunction: TimingFunctionType = .linear
     let scaleAnimation = CAKeyframeAnimation(keyPath: "transform.scale")
     scaleAnimation.keyTimes = [0, 0.45, 0.55, 1]
-    scaleAnimation.timingFunctions = [inTimingFunction, standByTimingFunction, outTimingFunction]
+    scaleAnimation.timingFunctionsType = [inTimingFunction, standByTimingFunction, outTimingFunction]
     scaleAnimation.values = [1, 1.3, 1.3, 1]
     scaleAnimation.duration = duration
     scaleAnimation.repeatCount = .infinity
@@ -118,7 +118,7 @@ private extension ActivityIndicatorAnimationOrbit {
   var ring1ScaleAnimation: CAKeyframeAnimation {
     let scaleAnimation = CAKeyframeAnimation(keyPath: "transform.scale")
     scaleAnimation.keyTimes = [0, 0.45, 0.45, 1]
-    scaleAnimation.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear)
+    scaleAnimation.timingFunctionType = .linear
     scaleAnimation.values = [0, 0, 1.3, 2]
     scaleAnimation.duration = duration
     return scaleAnimation
@@ -126,9 +126,8 @@ private extension ActivityIndicatorAnimationOrbit {
 
   var ring1OpacityAnimation: CAKeyframeAnimation {
     let opacityAnimation = CAKeyframeAnimation(keyPath: "opacity")
-    let timingFunction = CAMediaTimingFunction(controlPoints: 0.19, 1, 0.22, 1)
     opacityAnimation.keyTimes = [0, 0.45, 1]
-    ring1ScaleAnimation.timingFunctions = [CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear), timingFunction]
+    opacityAnimation.timingFunctionsType = [.linear, .easeOutExpo]
     opacityAnimation.values = [0.8, 0.8, 0]
     opacityAnimation.duration = duration
     return opacityAnimation
@@ -164,7 +163,7 @@ private extension ActivityIndicatorAnimationOrbit {
   var ring2ScaleAnimation: CAKeyframeAnimation {
     let scaleAnimation = CAKeyframeAnimation(keyPath: "transform.scale")
     scaleAnimation.keyTimes = [0, 0.55, 0.55, 1]
-    scaleAnimation.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear)
+    scaleAnimation.timingFunctionType = .linear
     scaleAnimation.values = [0, 0, 1.3, 2.1]
     scaleAnimation.duration = duration
     return scaleAnimation
@@ -172,9 +171,8 @@ private extension ActivityIndicatorAnimationOrbit {
 
   var ring2OpacityAnimation: CAKeyframeAnimation {
     let opacityAnimation = CAKeyframeAnimation(keyPath: "opacity")
-    let timingFunction = CAMediaTimingFunction(controlPoints: 0.19, 1, 0.22, 1)
     opacityAnimation.keyTimes = [0, 0.55, 0.65, 1]
-    ring2ScaleAnimation.timingFunctions = [CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear), timingFunction]
+    opacityAnimation.timingFunctionsType = [.linear, .easeOutExpo]
     opacityAnimation.values = [0.7, 0.7, 0, 0]
     opacityAnimation.duration = duration
     return opacityAnimation
--- a/IBAnimatable/Animatable.swift
+++ b/IBAnimatable/Animatable.swift
@@ -44,9 +44,14 @@ public protocol Animatable: class {
    */
   var force: CGFloat { get set }
 
+  /**
+   Animation function as a timing curve. (default value should be none)
+   */
+  var timingFunction: TimingFunctionType { get set }
+
 }
 
-public extension Animatable where Self: UIView {
+public extension Animatable {
   public func configureAnimatableProperties() {
     // Apply default values
     if duration.isNaN {
@@ -65,7 +70,9 @@ public extension Animatable where Self: UIView {
       force = 1
     }
   }
+}
 
+public extension Animatable where Self: UIView {
   @discardableResult
   public func animate(_ animation: AnimationType,
                       duration: TimeInterval? = nil,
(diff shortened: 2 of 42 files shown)


tbaranes:
@phimage Sorry it takes me time to review it, it's quite big and a lot to digest! 
I have already a first comment: can you update the docs? I think you will the one to update is `Documentation/APIs.md`

phimage:
No problem!

API doc has not been reverted https://github.com/IBAnimatable/IBAnimatable/blob/master/Documentation/APIs.md
I add `timingFunction`  in "Animatable protocol" section.
If you think I can add something elsewhere, I will do it :)

I see x, y, repeat.. attributes in the same section, maybe old parameters, now directly  in animationType

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+
+    // customs
+    case "spring" where params.count == 1:
+      self = .spring(damping: params[0].toFloat() ?? 0)
+    case "custom" where params.count == 4:
+      let c1 = (params[0].toFloat() ?? 0, params[1].toFloat() ?? 0)
+      let c2 = (params[2].toFloat() ?? 0, params[3].toFloat() ?? 0)
+      self = .custom(c1, c2)
+
+    case "easeinsine":
```
tbaranes: Can you add a comment as the ones above to keep the same structure as the enum?

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+  // from http://easings.net/
+  case easeInSine, easeOutSine, easeInOutSine
+  case easeInCubic, easeOutCubic, easeInOutCubic
+  case easeInQuad, easeOutQuad, easeInOutQuad
+  case easeInQuart, easeOutQuart, easeInOutQuart
+  case easeInQuint, easeOutQuint, easeInOutQuint
+  case easeInExpo, easeOutExpo, easeInOutExpo
+  case easeInCirc, easeOutCirc, easeInOutCirc
+  case easeInBack, easeOutBack, easeInOutBack
+
```
tbaranes: Typo: there's one newline to remove

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+    case "easeinoutcirc":
+      self = .easeInOutCirc
+    case "easeinback":
+      self = .easeInBack
+    case "easeoutback":
+      self = .easeOutBack
+    case "easeinoutback":
+      self = .easeInOutBack
+
+    default:
```
tbaranes: Same here, I know it's already explicit, but I think it's better to keep the same structure as the enums. It helps to follow

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+      return left
+    }
+  }
+
+}
+extension TimingFunctionType {
+
+  public var caType: CAMediaTimingFunction {
+    switch self {
```
tbaranes, line 158: Same as above, if you can add the comments to separate the main groups

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+    case .easeInOut:
+      return .curveEaseInOut
+    default:
+      return nil
+    }
+  }
+
+}
+
+@available(iOSApplicationExtension 10.0, *)
```
tbaranes: `iOSApplicationExtension`, did you mean `iOS`?

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+    getControlPoint(at: 0, values: &cps[0])
+    getControlPoint(at: 1, values: &cps[1])
+    getControlPoint(at: 2, values: &cps[2])
+    getControlPoint(at: 3, values: &cps[3])
+    return ((cps[0], cps[1]), (cps[2], cps[3]))
+  }
+
+}
+
+internal extension CAAnimation {
```
tbaranes, line 333: Is there a reason to explicitly make it `internal`?

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,355 @@
...
+
+  // conveniant setter, getter not implemented
```
tbaranes: Typo:  `conveniant` -> `convenient`

IBAnimatableApp/AnimationsTimingFunctionViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+fileprivate let c2x = ParamType.number(min: 0, max: 1, interval: 0.1, ascending: true, unit: "")
+fileprivate let c2y = ParamType.number(min: 0, max: 1, interval: 0.1, ascending: true, unit: "")
+
+fileprivate let damping = ParamType.number(min: 0, max: 1, interval: 0.1, ascending: true, unit: "")
+
+protocol TimingFunctionPickDelegate: NSObjectProtocol {
+  func timingFunctionSelected(_ timingFunction: TimingFunctionType)
+}
+
+class AnimationsTimingFunctionViewController: UIViewController {
```
tbaranes: final

IBAnimatableApp/AnimationsViewController.swift
```diff
@@ -12,13 +12,14 @@ private let fadeWayParams = ParamType(fromEnum: AnimationType.FadeWay.self)
 private let axisParams = ParamType(fromEnum: AnimationType.Axis.self)
 private let rotationDirectionParams = ParamType(fromEnum: AnimationType.RotationDirection.self)
 private let positiveNumberParam = ParamType.number(min: 0, max: 50, interval: 2, ascending: true, unit:"")
-private let numberParam = ParamType.number(min: -50, max: 50, interval: 5, ascending: true, unit: "")
+private let numberParam = ParamType.number(min: -50, max: 200, interval: 10, ascending: true, unit: "")
 private let repeatCountParam = ParamType.number(min: 1, max: 10, interval: 1, ascending: true, unit:"")
 
 class AnimationsViewController: UIViewController {
```
tbaranes: final

tbaranes (changes requested):
That's a HUGE work, thanks for doing this, it's an awesome feature.
I let a few comments, mainly typos / minor stuff, it looks pretty good to me 👍 

Can you add the docs update once more? I think that's enough.

IBAnimatable/TimingFunctionType.swift
phimage: It's a way for me to mark dangerous extension
I want to use it internally but never let it public, or someone else does it.

With `internal` explicitly marked, people ask themselves why and edit carefully

tbaranes:
Just tested the example, great one. A few feedbacks:

- At first, I though that the timing function will update the animation. Since it doesn't, you should either do it, or refactoring makes a subsections to "Animations" to have two possibilities: "View Animations", and "Timing functions"
- The image in the navigation bar isn't explicit enough (I didn't get it at first), if you choose option 1 from the previous point, we should replace the image by a text, or move the button in a place even more visible / explicit
- Since some `timingFunctions` are quite long, they are truncated in the picker. Can we add the selected name somewhere? Below the graph for example. I think we have this issue in others picker as well, time to take care of this by starting with this one? 😬 
- I know there's a few incoherence about this in the playground: should we have a "back" text and no title, or no "back" text and a title? By using "TimingFunctions", I would go for option 2 to make it more explicit, but it's up to you. Some day it would be great to uniformise that behaviour

IBAnimatableApp/AnimationsViewController.swift
phimage: I think do a PR to set all `final` it's a better choice
Because here it's not a new code here. But I copy this code to make my new controller :-)

tbaranes: Yeah, I know that we didn't add the final at the beginning of this project, we should have this in our roadmap: add all the missing finals 💥

phimage:
The animation is updated in the previous view, `TimingFunctionPickDelegate` do it

But not all animation are compatible ie. Spring animation are not compatible
In the same way velocity, dumping are only for spring animation and update it will not affect a lot of animations
Maybe a documentation file with a list of animation with two columns: Spring animation and timing function animation(or cubic animation) will be necessary, to check option compatibility

I edit moveTo and moveBy to be compliant with the two modes.

---
Yes, the image is not explicit
I take an available image, a timer. I think about putting a timing function graph but failed to create it using my TimingFunctionView in navigation bar items.

phimage:
So new change added

- back button and navigation title for the timing function picker
- icon changed to a Gear, and the button is enabled only if the selected animation is configurable using the timing function

Added on AnimationType two compute boolean variables to know is the animation use velocity and damping or use timing function
This method could help for future animation documentations

SD10:
@phimage this is great work! I may be able to leverage this approach to do more cleanup or adopt a similar strategy. love it!

IBAnimatable/TimingFunctionType.swift
tbaranes: Github didn't see as fixed, but it's fixed!

tbaranes (changes requested):
@phimage Awesome! I agree with you that a docs about timing function compatibility would be great, but we can merged that one first. So, just the missing entry in the docs and that's ready to go 👍

phimage:
For the doc, as I say, I do already something in main branch, see Animatable protocol
https://github.com/IBAnimatable/IBAnimatable/blob/master/Documentation/APIs.md

tbaranes (approved):
My bad, I misunderstood 😶

IBAnimatable/ActivityIndicatorAnimationOrbit.swift
```diff
@@ -118,17 +118,16 @@ private extension ActivityIndicatorAnimationOrbit {
...
     scaleAnimation.duration = duration
     return scaleAnimation
   }
 
   var ring1OpacityAnimation: CAKeyframeAnimation {
     let opacityAnimation = CAKeyframeAnimation(keyPath: "opacity")
-    let timingFunction = CAMediaTimingFunction(controlPoints: 0.19, 1, 0.22, 1)
     opacityAnimation.keyTimes = [0, 0.45, 1]
-    ring1ScaleAnimation.timingFunctions = [CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear), timingFunction]
+    opacityAnimation.timingFunctionsType = [.linear, .easeOutExpo]
```
JakeLin, line 130: good to encapsulate to an extension for timing function.

IBAnimatable/Animatable.swift
```diff
@@ -44,9 +44,14 @@ public protocol Animatable: class {
...
 
+  /**
+   Animation function as a timing curve. (default value should be none)
+   */
+  var timingFunction: TimingFunctionType { get set }
+
 }
 
-public extension Animatable where Self: UIView {
+public extension Animatable {
```
JakeLin, line 54: I hope Swift can support the syntax like `public extension Animatable where Self: UIView or UIBarItem` later.

IBAnimatable/Animatable.swift
```diff
@@ -186,22 +193,45 @@ fileprivate extension Animatable where Self: UIView {
...
+        yToMove = position.y
+      } else {
+        yToMove = CGFloat(y) + frame.height / 2
+      }
+
+      let path = UIBezierPath()
+      path.move(to: position)
+      path.addLine(to: CGPoint(x: xToMove, y: yToMove))
+
+      animatePosition(path: path, configuration: configuration, completion: completion)
```
JakeLin, line 233: Cool, before iOS 10, only Core Animation can apply custom timing.

IBAnimatable/Animatable.swift
```diff
@@ -589,3 +655,42 @@ public extension Animatable where Self: UIBarItem {
...
+      return false
+    case .fade(way: .inOut), .fade(way: .outIn):
+      return false
+    case .none:
+      return false
+    }
+  }
+
+  /// This animation use timing function parameter.
+  public var isCubic: Bool {
```
JakeLin, line 680: Should it call something like `doesSupportTimingFunction`? I am not good at naming though 😂

IBAnimatable/AnimatableBarButtonItem.swift
```diff
@@ -41,6 +41,12 @@ open class AnimatableBarButtonItem: UIBarButtonItem, BarButtonItemDesignable, An
   @IBInspectable open var damping: CGFloat = CGFloat.nan
   @IBInspectable open var velocity: CGFloat = CGFloat.nan
   @IBInspectable open var force: CGFloat = CGFloat.nan
+  @IBInspectable open var _timingFunction: String = "" {
```
JakeLin: This one should not be `open`. It only is used in Interface Builder. Please have a look at the comment around https://github.com/IBAnimatable/IBAnimatable/issues/218#issuecomment-230324948 . We don't want the `_` properties be used in programmatical APIs.

IBAnimatable/AnimatedTransitioning.swift
```diff
@@ -47,8 +47,7 @@ public extension AnimatedTransitioning {
         transition.subtype = subtype
       }
       transition.duration = self.transitionDuration(using: transitionContext)
-      // Use `EaseOutQubic` for system built-in transition animations. Thanks to @lexrus
-      transition.timingFunction = CAMediaTimingFunction(controlPoints: 0.215, 0.61, 0.355, 1)
+      transition.timingFunctionType = .easeOutCubic
```
JakeLin, line 50: Great 🎉

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,360 @@
...
+  case `default`
+
+  // not defined
+  case none
+
+  // custom one
+  case custom((x: Float, y: Float), (x: Float, y: Float))
+  case spring(damping: Float)
+
+  // from http://easings.net/
```
JakeLin, line 23: 👍

IBAnimatable/TimingFunctionType.swift
```diff
@@ -0,0 +1,360 @@
...
+    default:
+      return nil
+    }
+  }
+
+}
+
+@available(iOS 10.0, *)
+extension TimingFunctionType {
+  var cubicTimingParameters: UICubicTimingParameters {
```
JakeLin, line 256: Supper cool to support `UIPropertyAnimator` API.

JakeLin (approved):
All good, great work 👍👍👍. Look forward to using it.

JakeLin:
@phimage thanks for such a great PR. I think PR is ready to go. Please have a look at this comment https://github.com/IBAnimatable/IBAnimatable/pull/478#discussion_r116907694 for the `_timingFunction`. It should be `internal` like the other `_` properties.

JakeLin:
@phimage Well done 🎉🎉🎉🎉🎉🎉

PR #480, opened 2017-05-13
Refactor Current Unit Tests to Protocol Oriented Approach #no-public-changes

SD10:
### Summary of Pull Request:
- This Pull Request refactors the recently merged CornerDesignable & Fill Designable tests.

### Basis for Refactoring:
- Previously, I was performing tests on each `Designable` protocol and came to the realization that this was limiting in the sense that we could not easily test how one `Designable` protocol affected the performance of another `Designable` protocol.

### Cons:
- There are a bit more repeated code and fewer opportunities to extract helper level test logic to tests.
(I think it's important to note that the repeated code in the short run is considerably less than what it would be in the long run when testing for clashes between protocols.)

### Pros:
- Organizing the tests this way matches IBAnimatable's protocol oriented architecture.
- We can now test for interference between designable protocols and different states of UI elements.
- This makes it easier to test support for other platforms. e.g) tvOS (which I hope we support soon)
- The tests are now decoupled from the protocols if we move forward with modularizing the frameworks based on the different categories of UIElements.
- Removed a more than half of the mock protocols.

### Suggestions:
Please let me know if anyone has any ideas or suggestions. 
It's far easier to change a testing pattern now than it will be later.


IBAnimatable.xcodeproj/project.pbxproj
```diff
@@ -791,14 +819,68 @@
 		B0564B441EB01A8100D1C5AB /* IBAnimatableTests */ = {
 			isa = PBXGroup;
 			children = (
-				B0564B4E1EB01AD000D1C5AB /* CornerDesignableTests.swift */,
-				B0564B501EB046F600D1C5AB /* FillDesignableTests.swift */,
-				B0564B471EB01A8100D1C5AB /* Info.plist */,
-				B0564B521EB06BA500D1C5AB /* ColorTypeTests.swift */,
+				B0FF35461EC6A691005C5F62 /* MockProtocols */,
```
JakeLin, line 822: Nice groups.

IBAnimatableTests/AnimatableButtonTests.swift
```diff
@@ -0,0 +1,61 @@
...
+
+// MARK: - CornerDesignable Tests
+
+extension AnimatableButtonTests: CornerDesignableTests {
+
+  func testCornerRadius() {
+    _testCornerRadius(animatableButton)
+  }
+
+  func test_cornerSides() {
```
JakeLin, line 35: Why do we have different naming conversions like `_` or Camel case? Shall we use only **Camel case**? Something like `testCornerSideString` or `testCornerSideForInterfaceBuilder`?

IBAnimatableTests/StringCornerDesignable.swift
```diff
@@ -0,0 +1,32 @@
...
+extension AnimatableScrollView: StringCornerDesignable {}
+extension AnimatableButton: StringCornerDesignable {}
+extension AnimatableCheckBox: StringCornerDesignable {}
+extension AnimatableImageView: StringCornerDesignable {}
+extension AnimatableLabel: StringCornerDesignable {}
+extension AnimatableStackView: StringCornerDesignable {}
+extension AnimatableTableViewCell: StringCornerDesignable {}
+extension AnimatableTextField: StringCornerDesignable {}
+extension AnimatableTextView: StringCornerDesignable {}
+extension AnimatableCollectionViewCell: StringCornerDesignable {}
```
JakeLin, line 32: Nice one to make all classes testable with generic/common protocols.

JakeLin (approved):
I like this approach to mimic the why we extract the protocol for `IBAnimatable`, and easy to find out which conforming class has problem during the test.

JakeLin:
@SD10 I like your new approach to organizing the tests in protocol oriented way, well done 👍, I think you need to merge your PRs carefully with the correct order.

IBAnimatableTests/AnimatableButtonTests.swift
SD10: In IBAnimatable we have a few internal properties using the _ prefix.
These are mainly the the `String` properties used in Interface Builder.

My current approach is to name tests based on the properties/methods they test:

For example:
- `cornerSides` property -> `testCornerSides()`
- `_cornerSides` property -> `test_cornerSides()`

You will also see the _ prefix is the methods provided in the extensions for the test protocols:

For example:
```swift
protocol CornerDesignableTests {
   func testCornerSides()
} 

extension CornerDesignableTests {
   func _testCornerSides() {}
}

class AnimatableView: XCTestCase, CornerDesignableTests {
   func testCornerSides() {
      _testCornerSides()
   }
}
```

I'm using this prefix here so it's really clear which method should be called inside the stub when conforming to this protocol.

Let me know if you have a better naming convention though! This is all trial and error.

SD10:
Yeah, I will proceed with rewriting all the other tests using this same approach. I didn't want to start doing all that work until I received feedback on this new approach.

tbaranes (approved):
👏

SD10:
#434 - marking this for record keeping

Issue #479, opened 2017-05-11
Project Organization & Submodules


SD10:
## Project Organization & Submodules
Recently, we have been discussing the current organization of the project on Slack.

### Basis for the Discussion:
- IBAnimatable is a large codebase and will continue to grow as we add more features.

- We want to limit the amount of unused imported code and decrease the size of the dependency.
    e.g) Using IBAnimatable solely for `Transitions` yet still having to import all `ActivityIndicators`

### Proposed Solution:
- One possible approach would be to split the project using CocoaPods subspecs.

    For example:

    `IBAnimatable` would import the entire framework.

    `IBAnimatable/Core` would import core protocols/components required by other subspecs.

    `IBAnimatable/Transitions` imports transitions and `IBAnimatable/Core`

    `IBAnimatable/ActivityIndicators` imports activity indicators and `IBAnimatable/Core`

I won't continue listing possible subspecs because they're a rough draft and likely to change. 

- I do think the inheritance hierarchy for `UIKit` could be a good guideline for this:

    e.g) `IBAnimatable/UIControls` or `IBAnimatable/Controllers`

### Pros:

1. Decreases the size of the dependency for IBAnimatable users and reduces unused imported code.

2. A more modular project is easier to debug, test, maintain, and optimize performance.

3. Working with more modular components would let us take each individual component further without bloating the code base and opens up the possibility for IBAnimatable plugins.

4. It could make supporting other platforms such as tvOS even easier.

### Cons:

1. The change in imports could be an API breaking change for many users of IBAnimatable.
- This could be limited by still having IBAnimatable import all subspecs.

2. It would most likely require decoupling some of the framework's components which could also lead to API breaking changes.

tbaranes:
I agree with the needs here, a few questions:
- How will it works for Carthage, Submodules... ? If one subspecs is a dedicated framework, then it will be straightforward
- How will we name them? If you named it `UIControls`, does that mean I will have to do `import UIControls`? It's not quite clear _what_ it represents, but if you named it `AnimatableUIControls`, it's starting to be long to write.
- When using one module, will I have to important `IBAnimatable` and `UIControls`? I think it can quickly become complex to use.
- Should we have one repository by dedicated "subproject" or keep them all in the same? It's quite the same as `IBAnimatableMaterial`

Beside a few unclear points, I'm totally ok for introducing API breaking change since it will really improve the usage. We will have a big work on installation / migration documentation. 

[...]

PR #484, opened 2017-05-18
Embedded xcodeprojs in workspace #no-public-changes

tbaranes:
First step for #479 

- No code updates
- Adding `IBAnimatable.xcworkspace`
- Adding two new xcodeproj:
    - `IBAnimatable.xcodeproj` (framework)
    - `IBAnimatableApp.xcodeproj` (demo / playground). In a long term process, I'm even wondering if we shouldn't make two demos: one for the _full_ demo, and a second one for playgrounds 
- Reorganising sources: map Xcode groups with finder (I may have broken a few groups orders in Xcode, I will reorganise them if the PR is approved)
- Updating paths for: travis, danger, cocoapods
- Move all the framework sources in a directory `Sources`

One question: I know that, from the beginning, we separate storyboards and codes in the demo app. I'm personally a huge fan of this pattern because I find it hard to navigate, and find back the related files. What do you guys prefers about this? If you are ok to group them, I can update the PR.

I'm targeting master because it doesn't introduce any breaking changes, only internal restructuration, but it will be very hard to rebase any PRs due to these changes 😬


SD10:
@tbaranes you mean group each storyboard file with its respective view controller?

I do it both ways depending on the project. I agree that grouping them together would significantly help when working with the demo app. I would vote yes for this change 👍

tbaranes:
That what I mean!

tbaranes:
I just reordered xcodeproj groups (same as master), I think it's ready to go. My question about the storyboards organisation still here, but we can do it in a second a PR. If travis passes, and everything is ok with it, it's ok to go 💥

JakeLin (approved):
👍 I think we need to update the README.md as well. `open IBAnimatable.xcworkspace`

JakeLin:
>One question: I know that, from the beginning, we separate storyboards and codes in the demo app. I'm personally a huge fan of this pattern because I find it hard to navigate, and find back the related files. What do you guys prefers about this? If you are ok to group them, I can update the PR.

I prefer to group them by features. I don't know why I did it that way (separating storyboards and codes)

PR #485, opened 2017-05-18
[WIP] Refactor MaskType Fix #482

### LINKED ISSUE #482: New mask types and change proposal
# Change proposal for faster new mask type implementation

- `MaskType` enum could provide an `UIBezierPath` using a function like this `func bezierPath(for rect: CGRect)`
- Then in `MaskDesignable` we ask the path and call the draw function only one time code

I think also a lot of code to create `UIBezierPath` must done in one `UIBezierPath` `extension` instead of `MaskDesignable` or `MaskType`
For instance
```swift
extension UIBezierPath {
    public convenience init(triangleInRect rect: CGRect) {
        // code from getTriangleBezierPath
    }
}
```
# New mask type

I have already a lot of `UIBezierPath` code to create the following path

- [x]  .heart ❤️ 
- [x]  .drop 💧 
- [x]  .gear(radius, cogs) ⚙️ 
- [x]  .ring(radius) 💍  
- [x]  .superEllipse 💠  https://en.wikipedia.org/wiki/Superellipse
- [x]  .plus ➕ 
- [x]  .crescent 🌙 

You can copy paste the list and checked the one you think that could be useful as mask layer

phimage:
To implement feature request #482

- [x] Move UIBezierPath creation code in one UIBezierPath extension

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ N/A
 #### Enhancements
 
 Add `TimingFunctionType` attribute to `Animatable` to specifies the speed curve of an animation. [#478](https://github.com/IBAnimatable/IBAnimatable/pull/478) by [@phimage](https://github.com/phimage)
+Add new mask type `.ellipse`. [#481](https://github.com/IBAnimatable/IBAnimatable/pull/481) by [@phimage](https://github.com/phimage)
 
 #### Bugfixes
 
--- a/IBAnimatable/MaskDesignable.swift
+++ b/IBAnimatable/MaskDesignable.swift
@@ -39,6 +39,8 @@ public extension MaskDesignable where Self: UIView {
     switch maskType {
     case .circle:
       maskCircle()
+    case .ellipse:
+      maskEllipse()
     case let .parallelogram(angle):
       maskParallelogram(with: angle)
     case let .polygon(sides):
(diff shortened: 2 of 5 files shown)


phimage:
Even if the PR is WIP, you can review it
the change is explained in #482

SD10:
Looks great! Very clean API 👍

IBAnimatable/UIBezierPathExtension.swift
```diff
@@ -0,0 +1,171 @@
...
+extension UIBezierPath {
+
+  // MARK: - Circle
+
+  /**
+   Create a Bezier path for a circle shape.
+   
+   - Parameter bounds: The bounds of shape.
+   */
+  convenience init(circleIn bounds: CGRect) {
```
tbaranes: Should it be public?

tbaranes (commented):
Nice idea! What about merge this first then make a PR per mask?

phimage:
I think I will wait for #484 before doing a new PR with new branch
And I will have my new mask ready
I want to be sure that I can go on

and one PR by mask it's too many constraints, due to CHANGELOG for instance...
I ask in #482 to validate the wanted mask, if some are useless, please let me know

IBAnimatable/UIBezierPathExtension.swift
phimage: I see that the XXXExtension.swift files contain public extension
but here is this no need to share externally this code

tbaranes:
Since #484 is merged, you can rebased (or restart a new PR) and go ahead 👍

JakeLin (approved):
Cool, good start 👍

PR #487, opened 2017-05-20
Rollback xcodeproj conf #no-public-changes

tbaranes:
- Two spaces indentation by default
- SwiftLint scripts: I duplicated the .swiftlint.yml to make sure it takes effects for the frameworks and the demo app, a better solution would be to add only one in the root directory, but I don't know how makes this working right now


tbaranes:
Seems like Danger is broken as well 🤔

tbaranes:
Update: Add `IBAnimatable.framework` as embedded binaries in the example app xcodeproj to ensure it will be compile each time (if necessary)

IBAnimatable/IBAnimatable.xcodeproj/project.pbxproj
```diff
@@ -418,7 +418,10 @@
 				E27FD0041ECD79DD00F74840 /* IBAnimatableTests */,
 				E27FCFF81ECD79DD00F74840 /* Products */,
 			);
+			indentWidth = 2;
```
JakeLin, line 421: Good.

JakeLin (approved):
👍

PR #488, opened 2017-05-20
Move test method constraints to extension level #no-public-changes

SD10:
### Summary of Pull Request:
This Pull Request adds an `associatedtype` to the test protocols so that we can constrain the IBAnimatable element being tested at the extension level as opposed to in the function signature.

~~It's not linted so it's only ready for review not merging.~~ I'll have to rebase after #487 anyways.

### DIFF
--- a/IBAnimatable/IBAnimatableTests/TestProtocols/CornerDesignableTests.swift
+++ b/IBAnimatable/IBAnimatableTests/TestProtocols/CornerDesignableTests.swift
@@ -13,6 +13,10 @@ import XCTest
 
 protocol CornerDesignableTests: class {
 
+  associatedtype Element
+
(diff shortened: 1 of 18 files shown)


tbaranes (approved):

IBAnimatable/IBAnimatableTests/TestProtocols/CornerDesignableTests.swift
```diff
@@ -13,16 +13,20 @@ import XCTest
 
 protocol CornerDesignableTests: class {
 
+  associatedtype Element
```
JakeLin, line 16: great idea 💪

JakeLin (approved):

PR #489, opened 2017-05-22
`stickOrMoveUp` keyboard translation option

tbaranes:
Any idea for a better name? I don't really like this one 😞 
<strike>The gif is missing, will do it later</strike>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ N/A
 
 #### Enhancements
 
-Add `TimingFunctionType` attribute to `Animatable` to specifies the speed curve of an animation. [#478](https://github.com/IBAnimatable/IBAnimatable/pull/478) by [@phimage](https://github.com/phimage)
+- Add `TimingFunctionType` attribute to `Animatable` to specifies the speed curve of an animation. [#478](https://github.com/IBAnimatable/IBAnimatable/pull/478) by [@phimage](https://github.com/phimage)
 Add new mask type `.ellipse`. [#481](https://github.com/IBAnimatable/IBAnimatable/pull/481) by [@phimage](https://github.com/phimage)
+- Add `stickOrMoveUp` keyboard translation option when presenting a viewController. [#489](https://github.com/IBAnimatable/IBAnimatable/pull/489) by [@tbaranes](https://github.com/tbaranes)
 
 #### Bugfixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,8 +11,9 @@ N/A
 
 #### Enhancements
 
-Add `TimingFunctionType` attribute to `Animatable` to specifies the speed curve of an animation. [#478](https://github.com/IBAnimatable/IBAnimatable/pull/478) by [@phimage](https://github.com/phimage)
+- Add `TimingFunctionType` attribute to `Animatable` to specifies the speed curve of an animation. [#478](https://github.com/IBAnimatable/IBAnimatable/pull/478) by [@phimage](https://github.com/phimage)
 Add new mask type `.ellipse`. [#481](https://github.com/IBAnimatable/IBAnimatable/pull/481) by [@phimage](https://github.com/phimage)
+- Add `stickOrMoveUp` keyboard translation option when presenting a viewController
```
JakeLin: Can you add `by @tbaranes`?

JakeLin (approved):
The effect looks very good  👍👍👍. I am not good at naming and can't figure out a better one. May ask the others whether we can find a better one.

IBAnimatableApp/IBAnimatableApp/Playground/Presentations/PresentingViewController.swift
```diff
@@ -45,7 +45,7 @@ final class PresentingViewController: AnimatableViewController, UIPickerViewData
   fileprivate let animations = ["None", "Flip", "CrossDissolve", "Cover(Left)", "Cover(Right)", "Cover(Top)", "Cover(Bottom)", "Zoom", "DropDown"]
   fileprivate let positions = ["Center", "TopCenter", "BottomCenter", "LeftCenter", "RightCenter"]
   fileprivate let sizes = ["Half", "Full", "custom(100)"]
-  fileprivate let keyboardTranslations = ["None", "MoveUp", "AboveKeyboard"]
+  fileprivate let keyboardTranslations = ["None", "MoveUp", "AboveKeyboard", "stickOrMoveUp"]
```
SD10, line 48: Are these meant to have mismatching casing?

tbaranes: We accept both. So I didn't really look at it, it may be better to keep the right syntax in order to avoid bad usage

SD10:
I can't really think of a better name than `stickOrMoveUp` either. 
Everything I try to think of is far worse and lacks clarity.
It almost seems like it should be the default behavior of `.aboveKeyboard`.

IBAnimatableApp/IBAnimatableApp/Playground/Presentations/PresentingViewController.swift
SD10: Yeah I just checked out `IBEnum`. Not a big deal. I see more mismatching casing in the file. If anything this demonstrates that we accept both formats.

tbaranes:
I thought about this too @SD10, but I have the feeling it's not explicit enough as well (and introducing breaking change 😬 ). Well, we can stick with that for now, and update later if someone comes up with a better idea!

PR #492, opened 2017-06-02
Update PaddingDesignable to set rects for text, edit, and placeholder instead of UIView spacer

SD10:
### Summary of Pull Request:

This pull request changes the implementation of `PaddingDesignable` to set the padding by insetting the bounds of the text, editing, and placeholder rects -- as opposed to the previous approach that used a UIView as a spacer.

### To Do:
- [x] CHANGELOG

### DIFF
--- a/IBAnimatable/IBAnimatableTests/TestProtocols/PaddingDesignableTests.swift
+++ b/IBAnimatable/IBAnimatableTests/TestProtocols/PaddingDesignableTests.swift
@@ -15,9 +15,15 @@ protocol PaddingDesignableTests: class {
 
   var element: Element { get set }
 
-  func testPaddingLeft()
-  func testPaddingRight()
-  func testPaddingSide()
+  func testLeftTextPadding()
+  func testRightTextPadding()
+  func testSideTextPadding()
+  func testLeftEditPadding()
+  func testRightEditPadding()
+  func testSideEditPadding()
+  func testLeftPlaceholderPadding()
+  func testRightPlaceholderPadding()
+  func testSidePlaceholderPadding()
 
 }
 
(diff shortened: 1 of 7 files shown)


IBAnimatable/IBAnimatableTests/TestProtocols/PaddingDesignableTests.swift
```diff
@@ -15,37 +15,126 @@ protocol PaddingDesignableTests: class {
...
+    XCTAssertEqual(newRect, mockRect)
+
+  }
+
+}
+
+// MARK: - Helper Enum
+
+fileprivate enum Edge {
+  case left(CGFloat)
```
JakeLin, line 72: I like this idea, can we use it in our real API for the library? But don't know how to support IB yet.

JakeLin (changes requested):
We discussed this before, your approach will have better control. I give 👍 for this PR. There are some breaking changes,can you update the CHANGELOG. let's discuss how can we ship it.

IBAnimatable/IBAnimatableTests/TestProtocols/PaddingDesignableTests.swift
SD10: I have to see if this can be used instead of the helper methods in `PaddingDesignable`.
Less code repetition.

tbaranes:
Looks good beside the missing points your reported :)

JakeLin:
Because there are breaking changes, I came up an idea, Should we keep the old variables, for example, paddingLeft will apply to all leftTextPadding, leftEditPadding and leftPlaceholderPadding . If those will have more priority, if we set leftTextPadding and paddingLeft. We use leftTextPadding over paddingLeft. I know the names are not consistent, it should be leftPadding instead of paddingLeft. But I think we need to fix it for next big release. Can I have you guys opinions on this? @SD10 @tbaranes thanks

SD10:
I don't think we should be too concerned about API breaking changes if they enhance they project. 
I hope I can add a lot more value to IBAnimatable in the long run so breaking changes will be inevitable.

That being said, I do think heavily documenting these changes and providing releases accordingly is extremely important.

One approach I think we could take is to do something like this:
We can use an `@available` attribute to trigger a deprecation warning on the old API
Then call the new implementation from within that API. 

For example (just a rough idea):

```Swift
public protocol PaddingDesignable: class {

  @available(*, deprecated: 4.2.0, message: "Deprecated. Please use leftTextPadding, leftEditPadding, leftPlaceholderPadding.")
  var paddingLeft: CGFloat { get set }
  
  var leftTextPadding: CGFloat { get set }
  var leftEditPadding: CGFloat { get set }
  var leftPlaceholderPadding: CGFloat { get set }
  
}

extension PaddingDesignable {

  var paddingLeft: CGFloat {
    get {
      return max(leftTextPadding, leftEditPadding, leftPlaceholderPadding)
    }
    set {
      leftTextPadding = newValue
      leftEditPadding = newValue
      leftPlaceholderPadding = newValue
    }
  }
}
```
We could even add a warning to the message saying that in `IBAnimatable 5.0` this property will be dropped completely. It's like a way of phasing out the old API.

SD10:
Another concern I have regarding keeping the old API is then we would have a total of 12 `@IBInspectable` properties just for the padding of a `UITextField`.

I already felt unease about having 9 with the new API and considered removing the `side` properties :(

JakeLin:
I have the same concern, in most of cases, we use the same padding for text, edit and placeholder. Why don't we use only three properties instead of nine? But we use your better implementation to configure three internally. If someone has the requirement to have different paddings for text, edit and placeholder. Then we add that back. If we have three different properties for one padding, the user has to put the same number in three places.  How do you think?

Thanks
Jake

SD10:
I agree. I can't think of a use case where I'd want to have different padding for the text rect, edit rect, and placeholder rect. It's usually the same on all 3. 

So just to make sure we're on the same page, let's stick to just 3 properties but we'll configure the bounds internally because it will be more performant than having the overhead of a UIView.

Good thinking to keep it simple.

JakeLin:
Cool, can we go ahead with only three properties first, add more if needed in the future?

IBAnimatable/IBAnimatableTests/TestProtocols/PaddingDesignableTests.swift
SD10: Couldn't manage to reuse this code internally, just helps clean up testing

SD10:
@JakeLin 
alright I kept the same API and updated the implementation, so no changes to Documentation. 
Just need to add CHANGELOG entry... and get a build to pass 😅
 Let me know what you think.

tbaranes (approved):

Sources/Protocols/Designable/PaddingDesignable.swift
```diff
@@ -6,53 +6,49 @@
...
-    rightViewMode = .always
-    rightView = padding
+    return UIEdgeInsetsInsetRect(bounds, paddingInsets)
   }
 
-  public func configurePaddingSide() {
+  private var paddingInsets: UIEdgeInsets {
     if paddingSide.isNaN {
-      return
+      return insets(left: paddingLeft, right: paddingRight)
```
JakeLin, line 38: How about one of this property is nan? We may need to check individually

Sources/Views/AnimatableTextField.swift
```diff
@@ -115,22 +115,23 @@ open class AnimatableTextField: UITextField, CornerDesignable, FillDesignable, B
...
+    return paddedRect(forBounds: bounds)
   }
 
-  @IBInspectable open var paddingSide: CGFloat = CGFloat.nan {
-    didSet {
-      configurePaddingSide()
-    }
+  override open func placeholderRect(forBounds bounds: CGRect) -> CGRect {
+    return paddedRect(forBounds: bounds)
   }
```
JakeLin, line 135: 👍 expect for a comment

JakeLin (approved):
👍 expect for a comment

Sources/Protocols/Designable/PaddingDesignable.swift
SD10: I did the check inside of `insets(left:right:)`. If either `isNaN` it will default to 0. 
I could extract the logic if it's not clear.

Sources/Views/AnimatableTextField.swift
SD10: Except**? Meaning you would like me to add some comments? 😅 Sorry

JakeLin: No, I mean my comment😅 I think you are good to go.

Sources/Protocols/Designable/PaddingDesignable.swift
JakeLin: Good👍

PR #498, opened 2017-06-09
[WIP] Create tvOS target and initial support

SD10:
### Summary of Pull Request:
This is me messing around seeing what supporting tvOS would look like 😅


IBAnimatable/IBAnimatable tvOS/IBAnimatable tvOS.h
```diff
@@ -0,0 +1,19 @@
...
+//  IBAnimatable tvOS
+//
+//  Created by Steven on 6/9/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+//! Project version number for IBAnimatable tvOS.
+FOUNDATION_EXPORT double IBAnimatable_tvOSVersionNumber;
```
JakeLin: Do we need this file?

IBAnimatable/IBAnimatable tvOSTests/IBAnimatable_tvOSTests.swift
```diff
@@ -0,0 +1,36 @@
...
+//  IBAnimatable tvOSTests
+//
+//  Created by Steven on 6/9/17.
+//  Copyright © 2017 IBAnimatable. All rights reserved.
+//
+
+import XCTest
+@testable import IBAnimatable
+
+class IBAnimatable_tvOSTests: XCTestCase {
```
JakeLin: When we test IBAnimatable, do we need to have a separate test for tvOS?

Sources/Views/AnimatableTableViewCell.swift
```diff
@@ -156,7 +159,9 @@ open var startPoint: GradientStartPoint = .top
   fileprivate func configureInspectableProperties() {
     configureAnimatableProperties()
     configureOpacity()
+    #if os(iOS)
```
JakeLin, line 162: Can we make all changes in the protocols, then someone conforms to the protocols and deson't need to apply if os in their subclasses.

JakeLin (commented):
I am excited to see IBAnimatable running on tvOS 👍👍👍 but a little bit upset to see so many if os

IBAnimatable/IBAnimatable tvOS/IBAnimatable tvOS.h
SD10: No, I think I can remove the whole `IBAnimatable tvOS` & `IBAnimatable tvOSTests` folders

IBAnimatable/IBAnimatable tvOSTests/IBAnimatable_tvOSTests.swift
SD10: No, I can do the same thing as the source target and wrap the tests that are only available on iOS with `#if os(iOS)`.

Sources/Views/AnimatableTableViewCell.swift
SD10: The only time a user would need to apply `#if os(iOS)` in a subclass is if they plan on using that subclass for both iOS and tvOS. I realize this is a hack solution to support tvOS -- so I'm open to something more elegant -- but haven't thought of anything yet. The fact that extensions cannot have stored properties is very limiting :(

SD10:
> I am excited to see IBAnimatable running on tvOS 👍👍👍 but a little bit upset to see so many if os

Yeah, this solution is not ideal. I'd only be happy with it if it was the only way. The goal here was to see what's available on tvOS and getting the project to build. Now we have to make tvOS support clean and elegant.

Sources/Animators/Common/Navigator.swift
```diff
@@ -26,6 +26,7 @@ public class Navigator: NSObject {
 
     animator = AnimatorFactory.makeAnimator(transitionAnimationType: transitionAnimationType, transitionDuration: transitionDuration)
 
+    #if os(iOS)
     // If interactiveGestureType has been set
     if let interactiveGestureType = interactiveGestureType {
       // If configured as `.Default` then use the default interactive gesture type from the Animator
```
SD10: This could be extracted into another initializer only for iOS. 
It may duplicate code but I think its better than a nested OS condition.

Sources/Views/AnimatableTableViewCell.swift
JakeLin: If we put the checking `#if os(iOS)` within `configureSeparatorMargins()` doesn't it help to eliminate the checking int he subclass? The subclass still calls it, and on tvOS, it doesn't do anything by default.

SD10: @JakeLin I think this can be done but I have to try it first. But then we would have to expose `TableViewCellDesignable` to the tvOS target.

SD10: I'll see what I can do to hide some things and then we can compare the two approaches.

JakeLin: yes, you are right, we have to expose TableViewCellDesignable to the tvOS target. That's not good either.

SD10:
Going to close this because unfortunately, it's over a year old. It is possible for IBAnimatable to achieve tvOS support, however, it requires a lot of `#if os` statements. There may be a better way to do this in the newer versions of iOS/Swift

JakeLin:
@SD10 sure, thanks for your attempt.

PR #499, opened 2017-06-12
Fix PlaceholderDesignable & update Demo for PaddingDesignable changes

SD10:
###Summary of Pull Request:

One of my commits in #492 cleared the values in IB for `PaddingDesignable` this updates the Demo app to replace them.

This led me to find that `PlaceholderDesignable` for `UITextField` was using the value from `placeholderText` (which defaults to an empty `String`) before checking the actual `placeholder` property. 

This resulted in an empty placeholder always being applied if you specified a `placeholderColor`.


tbaranes (approved):
We should definitely drop this designable property 😑

JakeLin (approved):
👍

PR #506, opened 2017-07-07
Remove BackgroundDesignable corruption of UIKit

### LINKED ISSUE #505: There is an mistake in v4.0, UITableView dose not have an outlet named delegate.
i find that v3.0 does not have the mistake.

there is only one pod in the Podfile, and didn't include any other code in the project.

SD10:
### Summary of Pull Request:
Removes direct extensions to UIKit classes using `BackgroundDesignable`.
Instead, I made `BackgroundImageDesignable` conform to `BackgroundDesignable`.
This is not the usual approach we take for designable protocols, but to remove this protocol I would have to repeat the same code 3+ times.

Should fix #505. It seems like if you extend a UIKit class it can cause Interface Builder to stop recognizing the IBOutlets for that class.


tbaranes (approved):

tbaranes (changes requested):
CHANGELOG entry missing
Thanks for taking care of this weird issue!

tbaranes (approved):

PR #510, opened 2017-09-10
Migration to Swift4 & iOS 11

SD10:
### Summary of Pull Request:
- Updates IBAnimatable to Swift 4.0
- Updates IBAnimatableApp to Swift 4.0
    - Adds @objc attribute to selector methods because this is no longer done by default
    - NSForegroundColorAttributeName -> .foregroundColor
- Updates IBAnimatableApp to iOS 11 target (**should we do this? was iOS 10.3 before**)
- Updates TravisCI to run on Xcode 9 and iOS 11

This could fail the build, need to see how it runs on iOS 9/iOS 10 in Travis. We could need some preprocessor wrapping to keep backward compatibility. 

Ideally, we still need to do a release representing the current state of IBAnimatable `master`.
This is just a look at what needs to be done to prepare for Swift 4. Doesn't look too bad 😃


SD10:
I forgot to fix the SwiftLint errors. About the failed build:
- Make the constant public (not ideal IMO)
- Use a literal for the default value (reptition)
- Move the default value into the initializer body... then we would need a default of `nil`?

I'm also not sure why I get this error with Travis but can't reproduce it in the project. Both app and framework are compiling fine

IBAnimatableApp/IBAnimatableApp.xcodeproj/project.pbxproj
```diff
@@ -476,7 +483,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.3;
+				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
tbaranes, line 486: I think both are ok for this one, but if we merged it in master, we need to keep 10.3 for people who didn't download the beta.

Sources/Enums/AnimationType.swift
```diff
@@ -121,8 +121,12 @@ extension AnimationType: IBEnum {
     case "moveby":
       self = .moveBy(x: params[safe: 0]?.toDouble() ?? 0, y: params[safe: 1]?.toDouble() ?? 0)
     case "scale":
-      self = .scale(fromX: params[safe: 0]?.toDouble() ?? 0, fromY: params[safe: 1]?.toDouble() ?? 0,
-                    toX: params[safe: 2]?.toDouble() ?? 1, toY: params[safe: 3]?.toDouble() ?? 1)
+      // We need to split this up or it results in `expression too complex to compile in a reasonable amount of time`
```
tbaranes: I think you can drop the comment here

tbaranes (approved):
That's cool!

tbaranes:
Not sure why you can't reproduce your issue, but I had it too in Xcode beta 6, could it be that Travis hasn't an up to date beta?
Anyway, if you don't reproduce it in local, I suggest that we don't fix it for now.

Sources/Enums/AnimationType.swift
SD10: Yeah, I just included it for the review 👍

SD10:
@tbaranes If we're going to ship something that fails travis it would be nice if another IBAnimatable contributor could check that they cannot reproduce this locally as well.

I found this related issue:
https://bugs.swift.org/browse/SR-5111

Although in this scenario we're dealing with a `public init` and `internal` constant as a default argument.

JakeLin:
@tbaranes I can reproduce the issue (internal constant) locally with Xcode 9 beta 6, we can fix it by changing it to be `public` .

JakeLin:
I have fixed all compilation warnings for Xcode 9 beta 6. Ready to review again. I think we need to merge it after Xcode GM releases.

tbaranes (approved):

SD10:
Yes @JakeLin I didn't have Xcode beta 9.6 which is why I couldn't reproduce it locally. Thank you for taking care of it for me 😃 It was nice to wake up and see it already done!

We definitely should wait to merge it after Xcode GM. I just didn't want any surprises having to convert the week of the release

tbaranes:
Isn't it time to merge this? 🙃

JakeLin:
@tbaranes I think it is ready. @SD10 do you think anything needs to do before merging it?

PR #516, opened 2017-10-13
When device orientation change reset the modal dialog context frame Fix #515

### LINKED ISSUE #515: Issue when changing device orientation with modal view presented over context
When changing orientation the `dimmingView` of `AnimatablePresentationController` keep the same frame

I have implemented a solution to fix but I make some change (I will PR)
- changing `PresentationConfiguration` from `struct `to `class`
- in `PresentationDesignable`, change `contextFrameForPresentation: CGRect?` to `var contextFrameForPresentation: (() -> CGRect)?`

phimage:
I am not sure that this is the best solution but I try to fix it.

To detect device orientation change I override `AnimatableModalViewController. viewDidLayoutSubviews`

When device orientation change I call the same code launched when updating `AnimatableModalViewController.contextFrameForPresentation`
(I make a method for that `configurePresenterFrameForPresentation`)

But how to update the frame without knowing the source, the segue, the parent view, ...?
I replace `contextFrameForPresentation` type from `CGRect` to closure to execute exactly the same code in `PresentOverCurrentContextSegue`
Maybe we can do better, find element in view hierarchy, make constraint, ...?

Then I see an issue, updating the `contextFrameForPresentation` in the `presenter`(`configurePresenterFrameForPresentation`) do nothing.
And I understand why. The `PresentationConfiguration` passed to the presentation controller is a `struct`, so a copy and no update in presentation controller.
By changing type to `class`, the object is shared.

### DIFF
--- a/Sources/Controllers/AnimatableModalViewController.swift
+++ b/Sources/Controllers/AnimatableModalViewController.swift
@@ -9,9 +9,9 @@ import UIKit
 open class AnimatableModalViewController: UIViewController, PresentationDesignable {
 
   // MARK: - AnimatablePresentationController
-  public var contextFrameForPresentation: CGRect? {
+  public var contextFrameForPresentation: (() -> CGRect)? {
     didSet {
-      presenter?.presentationConfiguration?.contextFrameForPresentation = contextFrameForPresentation
+      configurePresenterFrameForPresentation()
     }
   }
 
@@ -179,4 +179,10 @@ open class AnimatableModalViewController: UIViewController, PresentationDesignab
     super.viewDidAppear(animated)
     configureDismissalTransition()
   }
+
+  open override func viewDidLayoutSubviews() {
+    super.viewDidLayoutSubviews()
+    configurePresenterFrameForPresentation()
+  }
+
 }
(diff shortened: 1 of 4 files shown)


Sources/Controllers/AnimatableModalViewController.swift
```diff
@@ -179,4 +179,10 @@ open class AnimatableModalViewController: UIViewController, PresentationDesignab
     super.viewDidAppear(animated)
     configureDismissalTransition()
   }
+
+  open override func viewDidLayoutSubviews() {
```
JakeLin, line 183: I think that's safe to use `viewDidLayoutSubviews`, it will update the UI when any UI has been changed. Another way would be `viewWillTransitionToSize:withTransitionCoordinator:` for size class. But I think `viewDidLayoutSubviews` is fine.

JakeLin (approved):
Thanks for fixing that 👍

Sources/Protocols/Designable/PresentationDesignable.swift
```diff
@@ -103,12 +103,17 @@ public extension PresentationDesignable where Self: UIViewController {
     }
   }
 
+  public func configurePresenterFrameForPresentation() {
+       presenter?.presentationConfiguration?.contextFrameForPresentation = contextFrameForPresentation?()
+    print("presenter \( presenter?.presentationConfiguration?.contextFrameForPresentation)")
```
tbaranes: Could you remove this print?

tbaranes (approved):
Looks good!

Sources/Protocols/Designable/PresentationDesignable.swift
phimage: oups, I forgot to remove this one

PR #517, opened 2017-10-15
Extract code about MaskType drawing into UIBezierPath extension #no-public-changes

phimage:
This is the first part of issue #482 
Just a refactor to move code in UIBezierPath.

This is an old closed PR
https://github.com/IBAnimatable/IBAnimatable/pull/485

(swiftlint warnings will be fixed...)


Sources/Extensions/UIBezierPathExtension.swift
```diff
@@ -0,0 +1,171 @@
...
+
+  // MARK: - Circle
+
+  /**
+   Create a Bezier path for a circle shape.
+   
+   - Parameter bounds: The bounds of shape.
+   */
+  convenience init(circleIn bounds: CGRect) {
+    let diameter = ceil(min(bounds.width, bounds.height))
```
JakeLin, line 18: do we need to call `self.init()` here?

JakeLin (approved):
👍  great stuff, good to go

tbaranes (approved):

Sources/Extensions/UIBezierPathExtension.swift
phimage: `self.init(` is called 3 lines under

JakeLin: @phimage cool 👍

PR #518, opened 2017-10-18
Add new mask types: hear, ring, gear and superEllipse

phimage:
4 new mask types (idea from #482 )
- .heart ❤️ 
- .gear(radius, cogs) ⚙️ 
- .ring(radius) 💍  
- .superEllipse 💠  https://en.wikipedia.org/wiki/Superellipse

### DIFF
--- a/Sources/Extensions/UIBezierPathExtension.swift
+++ b/Sources/Extensions/UIBezierPathExtension.swift
@@ -168,7 +168,7 @@ extension UIBezierPath {
 
    - Parameter bounds: The bounds of shape.
    */
-  convenience init(heartIn bounds: CGRect)  {
+  convenience init(heartIn bounds: CGRect) {
     self.init()
     var x: CGFloat = bounds.origin.x
     var y: CGFloat = bounds.origin.y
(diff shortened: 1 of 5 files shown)


Sources/Extensions/UIBezierPathExtension.swift
```diff
@@ -162,10 +162,148 @@ extension UIBezierPath {
...
+
+   - Parameter bounds: The bounds of shape.
+   - Parameter radius: The radius of the shape.
+   - Parameter cogs: The number of cogs (min: 2)
+   */
+  convenience init(gearIn bounds: CGRect, radius: CGFloat, cogs: Int) {
+    let center = CGPoint(x: bounds.midX, y: bounds.midY)
+    let diameter = ceil(min(bounds.width, bounds.height))
+    let innerRadius = max(1, diameter / 2 - radius)
+    let outerRadius = diameter / 2
```
JakeLin: We may be able to encapsulate these four lines into a method because they are same as line 231.

Sources/Extensions/UIBezierPathExtension.swift
```diff
@@ -162,10 +162,148 @@ extension UIBezierPath {
...
 }
 
 private extension UIBezierPath {
   func point(from angle: CGFloat, radius: CGFloat, offset: CGPoint) -> CGPoint {
     return CGPoint(x: radius * cos(angle) + offset.x, y: radius * sin(angle) + offset.y)
   }
 }
+
+extension CGFloat {
+  static let 𝑒 = CGFloat(M_E)
```
JakeLin, line 302: cool

JakeLin (approved):
Can you please update the CHANGELOG.md as well?

Sources/Extensions/UIBezierPathExtension.swift
phimage: You want something like this ?
```swift
let (center, diameter, innerRadius, outerRadius) = bounds.ringParameters(radius)
```

I can make CGRect extension
like 
```swift
var center: CGPoint { // I do not know why this do not exist already in CoreGraphic
  return CGPoint(x: self.midX, y: self.midY)
}
var diameter: CGFloat {
  return ceil(min(self.width, self.height))
}
```

JakeLin: I prefer to have extensions. They look better to me.

tbaranes (approved):

phimage:
I am wondering why I did not see all this swiftlint warning in Xcode?
I have not problem with my other project (but the configuration file is in same path)

```sh
$ which swiftlint
/usr/local/bin/swiftlint

$ swiftlint version
0.23.1

```

tbaranes:
Is the IBAnimatable's swiftlint script is still here and valid? Otherwise, don't see why 🤔

phimage:
to make it working in my local repo I add to the swiflint conf file 
```
included:
    - ../Sources
```

So I fix the issue (and see that there is a lot of warnings in activity indicators classes)

ps: I know that the conf file is shared with other project and targets, so my 'edit' will not work for them. So I keep it in my local repo


I add `--path $SRCROOT/../Source` in run script build phase and that also working

PR #519, opened 2017-10-20
Add new mask types .drop, .plusSign and .moon Close #482

### LINKED ISSUE #482: New mask types and change proposal

phimage:
_(fix swiftlint warning about colon in demo app)_

- [x]  .drop 💧 
- [x]  .plus ➕

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ N/A
 
 #### Enhancements
 
+- Add new mask types `.drop`, `.plusSign`, `.moon`. [#519](https://github.com/IBAnimatable/IBAnimatable/pull/519) by [@phimage](https://github.com/phimage)
 - Add new mask types `.heart`, `.gear`, `.ring`, `.superEllipse`. [#518](https://github.com/IBAnimatable/IBAnimatable/pull/518) by [@phimage](https://github.com/phimage)
 
 #### Bugfixes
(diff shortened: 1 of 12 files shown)


tbaranes (approved):

phimage:
Added moon shape with angle
- [x]  .moon 🌙 

Added also the documentation in APIs.md

Update changelog

tbaranes (approved):

PR #520, opened 2017-10-22
Create `.compound` `AnimationType` to make sequential animation (Fix #504)

### LINKED ISSUE #504: Making an animation using several AnimationType
For example:
- Fade out + moveTo
- Fade out + scale
- ...

Not sure how we can make this real in IBAnimatable 🤔

phimage:
Just a try for #504

Add `indirect` on `AnimationType` to be able to reference itself

The `compound` animation type here could take only two animations, but we could image an array

I have tested on demo with only some animations, and no parameters for them.

### DIFF
--- a/Sources/Protocols/Animatable/Animatable.swift
+++ b/Sources/Protocols/Animatable/Animatable.swift
@@ -92,11 +92,24 @@ public extension Animatable where Self: UIView {
     let completion = {
       promise.animationCompleted()
     }
-    doAnimation(animation, configuration: configuration, completion: completion)
+    doAnimation(animation ?? self.animationType, configuration: configuration, completion: completion)
   }
 
-  internal func doAnimation(_ animation: AnimationType? = nil, configuration: AnimationConfiguration, completion: @escaping () -> Void) {
-    switch animation ?? animationType {
+  /**
+   `autoRunAnimation` method, should be called in layoutSubviews() method
+   */
+  func autoRunAnimation() {
+    if autoRun {
+      autoRun = false
+      animate(animationType)
+    }
+  }
+}
+
+fileprivate extension UIView {
+
+  func doAnimation(_ animation: AnimationType, configuration: AnimationConfiguration, completion: @escaping () -> Void) {
+    switch animation {
     case let .slide(way, direction):
       slide(way, direction: direction, configuration: configuration, completion: completion)
     case let .squeeze(way, direction):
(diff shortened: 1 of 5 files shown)


phimage:
Now `compound` animation support a list of animations and each children animation could have parameters

But I use `[`and `]` to pass parameters (to not modify the method to parse params)
So we have for instance in storyboard animation definition
```
compound(slide, pop[2], wobble[5])
```

I think that a more elegant solution could be achieve in `AnimationType` parsing

@JakeLin , @tbaranes, @lastMove what do you think of 
```
[slide, pop(2), wobble(5)]
```
or
```
slide, pop(2), wobble(5)
```
or
```
slide+ pop(2)+ wobble(5)
```
to specify a list of animations

JakeLin:
@phimage nice one, I like `[slide, pop(2), wobble(5)]`, it looks like a list of animations. `+` is also good too. It implies we will run them at the same time.

phimage:
Ok I will split into two case

- the list [a, b] with sequential animations like my current implementation
- the operation a +b with simultaneous animations

_I think some combination will be "crappy"_ 

I will split `.compound` into two enum case (name to find) or add a parameter to distinct the two case

phimage:
Finally I made only one `enum` with a `Run` mode with value `sequential` or `parallel`

I have made a simple parsing of [a1, a2, ..] or a1 + a2 + ...,  using some string replace
with only one level , could not do [a1, [a2, a3], a4, ...]

phimage:
I think it is ready. I make some code cleaning
I will add the changelog after checking that git merging didn't break anything 

Next step is maybe to do a better parsing of expression (maybe doing a lexer, etc...) to support more expression

Sources/Protocols/Animatable/Animatable.swift
```diff
@@ -148,6 +135,34 @@ fileprivate extension UIView {
...
+      case .parallel:
+        var finalized = 0
+        let finalCompletion: () -> Void = {
+          finalized += 1
+          if finalized == animations.count {
+            completion()
+          }
+        }
+        for animation in animations {
+          self.doAnimation(animation, configuration: configuration, completion: finalCompletion)
```
JakeLin, line 176: We don't need the `self.` here

JakeLin (approved):
👍 please rebased and fix the conflicts before you merge, thanks

tbaranes:
Shouldn't we add documentation before merging?

phimage:
where to explain that?
@tbaranes any idea?

Looking for the good place to document I see a broken link in APIs.md to `AnimatableType` file

JakeLin:
@phimage will be great if you can add a use case in [the Swift playground file]( https://github.com/IBAnimatable/IBAnimatable/blob/master/IBAnimatable.playground/Pages/Predefined%20Animations.xcplaygroundpage/Contents.swift), That playground file demonstrates all predefined Animations. And also add a new section may be called  **Compounding animations**  after [Chaining animations](https://github.com/IBAnimatable/IBAnimatable#chaining-animations) and before **Delaying animations**.  Thanks

PR #527, opened 2018-01-08
Radial Gradients

### LINKED ISSUE #508: Radial gradient
Would be great to support both gradient type: Linear AND Radial

tbaranes:
Close #508 

Currently in this PR:
- New enum `GradientMode`: currently supporting `linear` and `radial`
- Working version of radial gradient (basic configuration only)
- Moved "gradients" example to their own section (playground example), + add of radial gradients example
- Documentation and CHANGELOG

Currently missing:
- Add `startPoints` (and custom) support to radial gradient

Any feedbacks welcome :)
  
simulator screen shot - iphone 6 - 2018-01-08 at 11 13 57


tbaranes:
There's still rooms for improvements but I think we can already merge this then add the missing features in a future PR (support for `startPoint`), helps welcome 😬

Sources/Protocols/Designable/GradientDesignable.swift
```diff
@@ -6,145 +6,153 @@
...
 
-fileprivate extension CAGradientLayer {
+    let points = layerPoints()
+    layer.startPoint = points.0
+    layer.endPoint = points.1
+    return layer
+  }
 
-  func configurePoints(with startPoint: GradientStartPoint) {
+  private func layerPoints() -> (CGPoint, CGPoint) {
```
gkye: Could we keep the name `configurePoints` or `configureGradientPoints`? It's a lot more readable imo

tbaranes: I’d like to drop the `comfigure` because it doesn’t change the layer anymore, it will simply returns the points. Beside that, I’m completely open for another name!

tbaranes: Maybe `gradientPonts`?

gkye: Okay that makes sense. `gradientPoints` should work 😄

gkye (approved):

Sources/Protocols/Designable/GradientDesignable.swift
```diff
@@ -6,145 +6,153 @@
...
-    let statusBarHeight: CGFloat = 64 // UIApplication.shared.statusBarFrame.height
-    let navBarFrame = CGRect(x: 0, y: 0, width: bounds.width, height: statusBarHeight)
-    gradientLayer.frame = navBarFrame
+// MARK: - Layers
 
-    gradientLayer.cornerRadius = layer.cornerRadius
-    return gradientLayer
+extension CALayer {
+
+  fileprivate func makeImage() -> UIImage? {
```
JakeLin, line 57: I think we can use `private` now

Sources/Protocols/Designable/GradientDesignable.swift
```diff
@@ -6,145 +6,153 @@
...
 }
 
-private class PrivateGradientView: UIView {
+// MARK: - Gradients maker
 
-  init(frame: CGRect, gradientLayer: CAGradientLayer) {
-    super.init(frame: frame)
-    layer.insertSublayer(gradientLayer, at: 0)
-    autoresizingMask = [.flexibleWidth, .flexibleHeight]
+extension GradientDesignable {
```
JakeLin, line 69: Since all methods are `private`, we can just mark `GradientDesignable` as `private` and remove the others.

JakeLin (approved):
Great job👍, this is a very useful feature, please merge in💪.

Sources/Protocols/Designable/GradientDesignable.swift
tbaranes: In fact, that's not working in swift, you can try this in a playground as an example:

```swift
final class Test {

}

private extension Test {

    func test() {
        print("test")
    }

}

let test = Test()
test.test()
```

You can call the method from the private extension from outside even if the extension is private, that's why I moved the private to the methods instead of the extension.

Sources/Protocols/Designable/GradientDesignable.swift
tbaranes: Can't do that, it's still in use in the `GradientDesignable` extension used for the navigation bar, otherwise, I could move it, but it's quite useful to have it in the layer extension 😬

Sources/Protocols/Designable/GradientDesignable.swift
JakeLin: @tbaranes interesting, I didn't know that 🤣

JakeLin:
@tbaranes @gkye thanks, guys, I merged this useful feature in now 👏

Issue #528, opened 2018-01-16
IBAnimatable without Interface Builder


JakeLin:
Currently, we need to use the subclass like `AnimatableView`. If we can provide an extension property like `ib` to UIKit components, then we can use the properties from the protocols e.g. `Animatable`, `CornerRadiusDesignable` etc.  Then we can use this `ib` property to set the animatable or designable properties  like `view.ib.cornerRadius = 20` in code. `view` is a just pure `UIView` not `AnimatableView`.

Since `IBAnimatable` has some powerful protocols, some developer they don't want to use the subclasses (e.g. `AnimatableView`) and they don't want to make a subclass by themselves. Providing an extension property `ib` can help them use `IBAnimatable` easily.

tbaranes:
Support for `UIView`:
- Basics #529 
- `BorderDesignable` #531
- `RotationDesignable` #532
- `ShadowDesignable` #533 
- `BlurDesignable` #534
- `GradientDesignable` #535
- `MaskDesignable` #536

tbaranes:
Here a list of what's left once #536 is merged:
- Documentation: we need to clarify how we want to do it
- Demo app: is this necessary?
- Support of Animatable: that one is a hard one because everything is depending of `UIView`
- Extends `ib` to more types (BarButton, ...)

If anyone wants to pick one of these task, feel free to do it. I won't have time to work on these during the next weeks 😬

PR #529, opened 2018-01-17
Introducing "IB" #no-public-changes

tbaranes:
I take a look at #528 to make `IBAnimatable` working without interface builder, AND without subviews. 

This PR implements a working solution for this issue, currently supporting:
- FillDesignable
- CornerDesignable

Usage:

```swift
view.ib.cornerRadius = 12
view.ib.opacity = 0.1
view.ib.fillColor = .green
view.ib.predefinedColor = ColorType.flatAlizarin
```

In order to get this, I had to change each protocols to add a `configure(in view: UIView)` since `ib` is not a view itself. It will ask a bit of work to refactor all of them, but it will work in the end. There's one issue left: how to "reload" the properties when rotating, bounds did change... all the work currently in `layoutSubviews`

Other branches related to this feature ready to go (after a rebase):
- `feature/ib_borderdesignable`

Also, I don't know how we should document (documentation + demo project) this, any suggestions?


JakeLin:
@tbaranes 💪 thanks for picking up this feature quickly, it is a fun and challenging one. It is a great start 👍

 > I had to change each protocols to add a configure(in view: UIView) since ib is not a view itself.

I think we need to do more than just `UIView`, some of them are not `UIView`s. I haven't seen any better way to achieve the goal without doing that yet

> how to "reload" the properties when rotating,

Is it possible to provide a `reload` method like `view.ib.reload` to ask the user to call? It will call `configureAfterLayoutSubviews()` and `autoRunAnimation()`

>  I don't know how we should document (documentation + demo project) this, any suggestions?

I prefer adding code comments as documentation now. We may add documentation comment in `class IB`, we can refer to how the other libraies doing , for example, RxSwift. And using a playground is good for demonstrating the API in done.

tbaranes:
> I think we need to do more than just UIView, some of them are not UIViews. I haven't seen any better way to achieve the goal without doing that yet

Which component do you have in mind? Anyway, I think we can start by supporting only `UIView` then add the others later by keeping a list in #528 

> I prefer adding code comments as documentation now. We may add documentation comment in class IB, we can refer to how the other libraies doing , for example, RxSwift. And using a playground is good for demonstrating the API in done.

Sounds good to me, I don't really have a preference. Should we create a dedicated PR for this once the feature itself is more advanced?

tbaranes:
`reload()` method added, I think it's ready to go beside documentation (depending if we make the documentation in this PR or in a new one)

tbaranes:
Travis failing due to the tests, will pass once #530 has been merged and rebased

tbaranes:
@JakeLin Ready for final review considering that docs will be added in another PR!

JakeLin (approved):
👍  It's good to go

JakeLin:
@tbaranes 

> Which component do you have in mind? Anyway, I think we can start by supporting only UIView then add the others later by keeping a list in #528

I was thinking some components like `AnimatableBarButtonItem`. They are not `UIView`s

JakeLin:
@tbaranes great work 👍

PR #530, opened 2018-01-17
Fix view's borders when using it with corner radius `allSides` 

tbaranes:
The border doesn't respect the `allSides` radius when both are set. Using a custom layer for every radius fix this.

### DIFF
--- a/Sources/Protocols/Designable/CornerDesignable.swift
+++ b/Sources/Protocols/Designable/CornerDesignable.swift
@@ -26,18 +26,11 @@ public extension CornerDesignable where Self: UICollectionViewCell {
         layer.mask?.removeFromSuperlayer()
       }
       layer.cornerRadius = 0.0
+      contentView.layer.cornerRadius = 0.0
 
-      if cornerSides == .allSides {
-        contentView.layer.cornerRadius = cornerRadius
-      } else {
-        contentView.layer.cornerRadius = 0.0
-
-        // if a layer mask is specified, remove it
-        contentView.layer.mask?.removeFromSuperlayer()
-
-        contentView.layer.mask = cornerSidesLayer()
-      }
-
+      // if a layer mask is specified, remove it
+      contentView.layer.mask?.removeFromSuperlayer()
+      contentView.layer.mask = cornerSidesLayer()
       contentView.layer.masksToBounds = true
     } else {
       contentView.layer.masksToBounds = false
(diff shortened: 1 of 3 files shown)


Sources/Protocols/Designable/CornerDesignable.swift
```diff
@@ -48,16 +41,11 @@ public extension CornerDesignable where Self: UICollectionViewCell {
...
-        // if a layer mask is specified, remove it
-        layer.mask?.removeFromSuperlayer()
-
-        layer.mask = cornerSidesLayer()
-      }
+      layer.cornerRadius = 0.0
+      
+      // if a layer mask is specified, remove it
+      layer.mask?.removeFromSuperlayer()
+      layer.mask = cornerSidesLayer()
```
JakeLin: can we encapsulate the code since they are very similar?

JakeLin (approved):
👍  thanks for fixing it. Please have a look at my small comment.

Sources/Protocols/Designable/CornerDesignable.swift
tbaranes: If that's ok with you, I will do the refactor in #529 since I'm already refactoring a bit the class there

tbaranes: Done there

tbaranes:
Travis was failing because of the tests, I fixed most of them, but there's still one failing that I don't understand why. I welcome any idea 🤔

SD10:
@tbaranes I'm sorry I missed this. If there's a problem with something I wrote, please tag me in the future

tbaranes:
You didn’t really miss it, my comment has been posted less than 2h ago! The tests are failing due to my change, your code is really clear and understandable but path are a bit tricky sometimes 😀

SD10:
It's still going after 6 hours @_@ I hate travis

tbaranes:
@SD10 Yeah, travis is completely down these days, but it will fail since the test is failing locally.

tbaranes:
Finally understood why that last test didn't pass, travis should be green now!

SD10:
@tbaranes I'm not sure if you know, I just learned this a couple months ago to remove those annoying SwiftLint warnings:

`Xcode` -> `Preferences` -> `Text Editing`

screen shot 2018-01-20 at 1 40 12 pm

PR #532, opened 2018-01-31
Add `RotationDesignable` support to `ib` #no-public-changes

tbaranes:
Related to #528


Sources/IB.swift
```diff
@@ -0,0 +1,54 @@
...
+//  IBAnimatableApp
+//
+//  Created by Tom Baranes on 17/01/2018.
+//  Copyright © 2018 IBAnimatable. All rights reserved.
+//
+
+import Foundation
+
+private struct AssociatedKeys {
+  static var ib = "IBAnimatable"
```
JakeLin: can be `let`?

Sources/IB.swift
```diff
@@ -0,0 +1,54 @@
...
+}
+
+public final class IB: NSObject, CornerDesignable {
```
JakeLin: I don't quite get it, why do we have two `IB` classes?

Sources/Others/IB.swift
```diff
@@ -8,7 +8,8 @@
 
 import Foundation
 
-public final class IB: NSObject, CornerDesignable, FillDesignable, BorderDesignable {
+public final class IB: NSObject, CornerDesignable, FillDesignable, BorderDesignable,
```
JakeLin, line 11: a space before `{` please 🤣

JakeLin (commented):
Good progress, please have a look my comments

Sources/IB.swift
tbaranes: Rebase failure 😔

Sources/IB.swift
tbaranes: 👍

tbaranes: > Cannot pass immutable value as inout argument: 'ib' is a 'let' constant

This is why :D

JakeLin (approved):
👍  That's a lot cleaner

PR #533, opened 2018-02-01
Add ShadowDesignable support to IB #no-public-changes

tbaranes:
Related to #528


JakeLin (approved):
👍  Since we have found the pattern to support `ib`, that's great.

PR #539, opened 2018-03-06
Animate TabBarItem image view

phimage:
See the re-opened issue #452 (I closed it thinking of a possible duplicate)

I make UIBarItem animatable by passing a view in `animate` function

There is many way to get a view to animate it, and no solution is perfect
I do not want to force the way to get the view

but for `AnimatableTabBarController` I make an example to get the internal `UIImageView`
The demo is updated to use `AnimatableTabBarItem`

---

Then in Animatable I move some function from `extension Animatable where Self: UIView {`to `extension UIView {` to be able to animate any view

---

I think there is some lags because I animate after tab bar item selection
`func tabBar(_ tabBar: UITabBar, didSelect item: UITabBarItem)`
there is no
`func tabBar(_ tabBar: UITabBar, willSelect item: UITabBarItem)`

### DIFF
--- a/Sources/Controllers/AnimatableTabBarController.swift
+++ b/Sources/Controllers/AnimatableTabBarController.swift
@@ -79,14 +79,12 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable 
   }
 }
 
-fileprivate extension UITabBar {
+private extension UITabBar {
 
     func view(for item: UITabBarItem) -> UIControl? {
       // return item.value(forKeyPath: "view") as? UIControl // apple could not allow that
-      guard let items = self.items else {
-        return nil
-      }
-      guard let index = items.index(of: item) else {
+      guard let items = self.items,
+        let index = items.index(of: item) else {
         return nil
       }
       // get all buttons
--- a/Sources/Controllers/AnimatableTabBarController.swift
+++ b/Sources/Controllers/AnimatableTabBarController.swift
@@ -46,9 +46,8 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable 
         return
     }
     // Animate the children image view
-    let subviews = tabBarButton.subviews.filter { $0 is UIImageView }
-    if let view = subviews.first {
-      animatable.animate(view: view)
+    if let imageView = tabBarButton.subviews.first(where: { $0 is UIImageView }) {
+      animatable.animate(view: imageView)
     }
   }
 
--- a/Sources/Controllers/AnimatableTabBarController.swift
+++ b/Sources/Controllers/AnimatableTabBarController.swift
@@ -39,6 +39,19 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable 
     }
   }
 
+  // MARK: - UITabBarDelegate
+  open override func tabBar(_ tabBar: UITabBar, didSelect item: UITabBarItem) {
+    guard let animatable = item as? (UITabBarItem & Animatable),
+      let tabBarButton = tabBar.view(for: item) else {
+        return
+    }
+    // Animate the children image view
+    let subviews = tabBarButton.subviews.filter { $0 is UIImageView }
+    if let view = subviews.first {
+      animatable.animate(view: view)
+    }
+  }
+
   // MARK: - Private
   // Must have a property to keep the reference alive because `UITabBarController.delegate` is `weak`
   fileprivate var navigator: Navigator?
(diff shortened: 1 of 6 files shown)


Sources/Controllers/AnimatableTabBarController.swift
```diff
@@ -39,6 +39,19 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable
     }
   }
 
+  // MARK: - UITabBarDelegate
+  open override func tabBar(_ tabBar: UITabBar, didSelect item: UITabBarItem) {
+    guard let animatable = item as? (UITabBarItem & Animatable),
```
JakeLin, line 44: should it be `item as? AnimatableTabBarItem`

Sources/Controllers/AnimatableTabBarController.swift
```diff
@@ -65,3 +78,28 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable
...
 }
+
+fileprivate extension UITabBar {
+
+    func view(for item: UITabBarItem) -> UIControl? {
+      // return item.value(forKeyPath: "view") as? UIControl // apple could not allow that
+      guard let items = self.items else {
+        return nil
+      }
+      guard let index = items.index(of: item) else {
```
JakeLin: We can group those two `guard let`s together

Sources/Controllers/AnimatableTabBarController.swift
```diff
@@ -65,3 +78,28 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable
     delegate = navigator
   }
 }
+
+fileprivate extension UITabBar {
```
JakeLin: We can use `private` now 😁

Sources/Controllers/AnimatableTabBarController.swift
```diff
@@ -65,3 +78,28 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable
...
+        return nil
+      }
+      guard let index = items.index(of: item) else {
+        return nil
+      }
+      // get all buttons
+      // 1/ filter on control, not safe if apple add a new control in bar
+      let controls = self.subviews.filter { $0 is UIControl }
+      // 2/ filter on name, not safe if button class change
+      // let controls = self.subviews.filter { String(describing: type(of: $0 )) == "UITabBarButton" }
```
JakeLin, line 93: Both ways are hacky. I have seen a lot of code is using `UITabBarButton`, it is more specific than `UIControl`. But either way, I am fine.

Sources/Controllers/AnimatableTabBarController.swift
```diff
@@ -39,6 +39,19 @@ open class AnimatableTabBarController: UITabBarController, TransitionAnimatable
...
 
+  // MARK: - UITabBarDelegate
+  open override func tabBar(_ tabBar: UITabBar, didSelect item: UITabBarItem) {
+    guard let animatable = item as? (UITabBarItem & Animatable),
+      let tabBarButton = tabBar.view(for: item) else {
+        return
+    }
+    // Animate the children image view
+    let subviews = tabBarButton.subviews.filter { $0 is UIImageView }
+    if let view = subviews.first {
```
JakeLin: I assume it should be only one image inside. That's fine, we can get the first one in one let

```
if let imageView = self.subviews.first(where: { $0 is UIImageView }) {
  animatable.animate(view: imageView)
}
```

Sources/Protocols/Animatable/Animatable.swift
```diff
@@ -92,7 +92,24 @@ public extension Animatable where Self: UIView {
...
+   */
+  func autoRunAnimation() {
+    if autoRun {
+      autoRun = false
+      animate(animationType)
+    }
+  }
+}
+
+fileprivate extension UIView {
```
JakeLin, line 109: Since it is `private`, that's fine to extend `UIView`, then we can share it with `UIBarItem`'s subview.

JakeLin (approved):
thanks, Eric, it is a great PR for extending IBAnimatable with non-UIView components 👍👍👍. Can you please have a look at my comments and also please fix some Swift Lint formatting issues.

Sources/Controllers/AnimatableTabBarController.swift
phimage: I want to let dev make their own UITabBarItem, using protocol programming (IBAnimatable is a lot based on that)

JakeLin:
I don't quite get it why the CI is failing

It shows 
> ▸ Touching IBAnimatableApp.app
> ▸ Signing /Users/travis/Library/Developer/Xcode/DerivedData/IBAnimatable-fjjlvkrlpfaitrbircdibfavsvhl/Build/Products/Release-iphonesimulator/IBAnimatableApp.app
> ▸ Build Succeeded

But later on 
> Danger has failed this build. 
> Found 1 error.
> The command "bundle exec danger" exited with 1.

SD10:
@JakeLin It's the missing CHANGELOG entry. Danger will fail the build until one is present

Sources/Controllers/AnimatableTabBarController.swift
SD10: Whoa I've never seen a type cast like this before 😱

JakeLin:
@phimage I tried to rebuild the commit in Travis, but it didn't report back. I cannot easily rebase the RP because it sits on your own fork, can you rebase it? or I can create a merge commit for this PR and merge it? Thanks

phimage:
I made merge commit, I know it is crappy, but not a lot of time

JakeLin:
@phimage thanks, I am fine with merge commit too.

PR #544, opened 2018-03-14
Reinstate CornerDesignable .allSides case with no mask applied

DanielAsher:
This supports situations when `clipsToBounds == false` and no mask is required.

I have an `AnimatableButton` subclass that uses `clipsToBounds == false` and has a badge subview . This could no longer be seen using master with reverting the refactor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ N/A
 - Add suppport for radial gradient. Currently not working with `startPoint`. [#527](https://github.com/IBAnimatable/IBAnimatable/pull/527) by [@tbaranes](https://github.com/tbaranes)
 
 #### Bugfixes
-
+- Fix CornerDesignable refactor 24d308d5ca that caused changed rendering behaviour (always applied mask, even when case .allSides)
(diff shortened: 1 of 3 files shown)


JakeLin (approved):
👍 good to go, thanks

DanielAsher:
Very nice integration of IBAnimatableBot and Travis - makes of us our best selves!

JakeLin:
thanks, @DanielAsher seems you are a power user of `IBAnimatable`, can I invite you to join us as a core contributor to the project? We can help make the lib better.

DanielAsher:
@JakeLin I'd be honoured!

JakeLin:
@DanielAsher I have sent you an invite, do you use Slack? Can you send me your email address for Slack? You can find my email on my GitHub profile page, thanks.

tbaranes:
Should we update the changelog here since that refactoring has never been released? (if I'm remembering well 😬)

SD10:
Would be awesome if the entry was in the same format as the others too 😕

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

JakeLin:
Agree with @tbaranes since we don't change any feature for the new release. We may not need this item.

DanielAsher:
Thanks for merging this!

PR #546, opened 2018-03-23
Don't override view's alpha when using zoom out

### LINKED ISSUE #545: Zoom out animation is overriding default alpha
For example: set a custom alpha to a view, then make a zoom out, the alpha will be back to 1 before executing the animation.

tbaranes:
Close #545


JakeLin (approved):
👍 thanks @tbaranes

PR #553, opened 2018-04-20
Create `insetBy` mask type for a smaller rectangle than the view frame

phimage:
```
insetBy(10, 50)
insetBy(10) // same as insetBy(10, 10)
```


phimage:
Github do not let reopen the pr
I will make a new one...

PR #562, opened 2018-05-07
Fix activity indicator with delayed animations on some shape

phimage:
Trying to add a new activity indicator named `rupee` 

ezgif com-video-to-gif 1

and comparing the animation gif in doc and the current one in demo app, I see an issue :
some animation which use `beginTime` doesn't working

I see for working animation that the animation group is created only one time, not for each additional layer. So I edit the bugged ones.

### DIFF
--- a/Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationRupee.swift
+++ b/Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationRupee.swift
@@ -23,17 +23,17 @@ public class ActivityIndicatorAnimationRupee: ActivityIndicatorAnimating {
     let x = (layer.bounds.size.width - size.width) / 2
     let y = (layer.bounds.size.height - size.height) / 2
     let beginTime = layer.currentMediaTime
-    let beginTimes: [CFTimeInterval] = [0, 0.5, 0.0, 0.5, 0.0, 0.5, 0.0, 0.5]
+    let beginTimes: [CFTimeInterval] = [0.0, 0.5, 0.0, 0.5, 0.0, 0.0, 0.5, 0.0]
     let animation = defaultAnimation
 
     // Draw shapes
     for i in 0 ..< 8 {
       if i % 4 != 0 { // remove left and right
         let shapeLayer = makeLayer(angle: .pi / 4 * CGFloat(i),
-                                     size: maskSize,
-                                     origin: CGPoint(x: x, y: y),
-                                     containerSize: size,
-                                     color: color)
+                                   size: maskSize,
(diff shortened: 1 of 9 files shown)


Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationRupee.swift
```diff
@@ -0,0 +1,88 @@
...
+//  Copyright © 2018 IBAnimatable. All rights reserved.
+//
+
+import UIKit
+
+public class ActivityIndicatorAnimationRupee: ActivityIndicatorAnimating {
+
+  // MARK: Properties
+
+  fileprivate let duration: CFTimeInterval = 1
```
JakeLin, line 15: We can use `private` now

Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationBallSpinFadeLoader.swift
```diff
@@ -55,7 +56,7 @@ public class ActivityIndicatorAnimationBallSpinFadeLoader: ActivityIndicatorAnim
 
 private extension ActivityIndicatorAnimationBallSpinFadeLoader {
 
-  var animation: CAAnimationGroup {
+  var defaultAnimation: CAAnimationGroup {
```
JakeLin, line 59: Great fix 👍

JakeLin (approved):
Great animation 👍, can you also add it to the domo app?

Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationRupee.swift
phimage: Could be done later. I just copy paste.
I want to make it public, editable later.
I try to add this duration as protocol parameter, but failed, some must total animation duration must be calculated

phimage:
Demo app is updated automatically using enum ;)

JakeLin:
@phimage But I can't find the animation `rupee` in the demo app playground 🤔

phimage:
@JakeLin did you build the framework?
Then build and launch the demo app?

I do not know why and since when but I see no target or compile dependencies between projects

screen shot 2018-05-07 at 15 01 01
screen shot 2018-05-07 at 15 01 07

JakeLin:
@phimage thanks, you are absolutely right, I can see it now. I removed the dependence by accident when I upgraded the project to support Xcode 9.3. Can you add it back to this PR? Or I can do a comment in master later on?

phimage:
@JakeLin  I merge so you can edit main branch (or make new PR)

PR #566, opened 2018-05-18
Add `preferred` presentation modal size to use preferred size defined…

phimage:
… in storyboard

- demo app edited
- if defined `preferredContentSize` values are different from 0, if not I use default ie. `half`


tbaranes (approved):

JakeLin (approved):
👍

PR #567, opened 2018-05-20
Add `conical` gradient type

phimage:
sometimes called angular, sweep(on android)

screen shot 2018-05-20 at 17 26 03

The current code draw lines starting at the center point, and change the color for each lines.

An optimisation could be to create an image with that, to not redraw it each time

---

Then I can also make a more smooth conical grandient by repeating the colors
instead of `layer.colors = colors`(with two colors) 
I put that `layer.colors = [colors[0], colors[1], colors[0]]`
to get that
screen shot 2018-05-20 at 17 25 05

Do not know how to include that in library. Maybe with the repeating gradient issue, if we could create odd number colors when repeating

if not with only even number of colors :
screen shot 2018-05-20 at 17 34 43

### DIFF
--- a/IBAnimatableApp/IBAnimatableApp/Playground/UserInterfaces/Gradients/UserInterfaceGradientsTableViewController.swift
+++ b/IBAnimatableApp/IBAnimatableApp/Playground/UserInterfaces/Gradients/UserInterfaceGradientsTableViewController.swift
@@ -12,11 +12,25 @@ import UIKit
 final class UserInterfaceGradientsTableViewController: UITableViewController {
 
   override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
-    if let gradientVC = segue.destination as? GradientViewController {
-      gradientVC.usePredefinedGradient = segue.identifier?.lowercased().contains("predefined") ?? false
-      gradientVC.useRadialGradient = segue.identifier?.lowercased().contains("radial") ?? false
-    } else if let gradientVC = segue.destination as? GradientCustomStartPointViewController {
-      gradientVC.useRadialGradient = segue.identifier?.lowercased().contains("radial") ?? false
+    if let identifier = segue.identifier?.lowercased() {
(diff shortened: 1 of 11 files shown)


tbaranes (approved):

Sources/Others/RadialGradientLayer.swift
```diff
@@ -25,8 +25,9 @@ final class RadialGradientLayer: CALayer {
   override func draw(in ctx: CGContext) {
     ctx.saveGState()
     let colorSpace = CGColorSpaceCreateDeviceRGB()
-    let gradient = CGGradient(colorsSpace: colorSpace, colors: colors as CFArray, locations: [0, 1])!
-
+    guard !colors.isEmpty, let gradient = CGGradient(colorsSpace: colorSpace, colors: colors as CFArray, locations: [0, 1])  else {
```
JakeLin, line 28: Good fix 👍

IBAnimatableApp/IBAnimatableApp/Playground/UserInterfaces/Gradients/UserInterfaceGradientsTableViewController.swift
```diff
@@ -12,11 +12,25 @@ import UIKit
...
+        gradientVC.usePredefinedGradient = identifier.contains("predefined")
+        if identifier.contains("radial") {
+          gradientVC.gradientMode = .radial
+        } else if identifier.contains("conical") {
+          gradientVC.gradientMode = .conical
+        } else {
+          gradientVC.gradientMode = .linear
+        }
+      } else if let gradientVC = segue.destination as? GradientCustomStartPointViewController {
+        if identifier.contains("radial") {
```
JakeLin: There are some duplicated code here

JakeLin (commented):
Cool feature, well done 👍

JakeLin (approved):

PR #571, opened 2018-06-27
TimingFunction for all `AnimationType` Fix #564

### LINKED ISSUE #564: `timingFunction` not working for spring animations
All animations using `UIView.animate` doesn't apply the timing function factor. Is there a reason for that?

phimage:
If a timing function is defined for "spring" animation type, do not use damping and velocity 

see explanation in #564

demo app updated to be able to select the timing function all animation types


JakeLin (approved):
👍👍👍Thanks for fixing that, they can't work together 😞

tbaranes (approved):
Awesome!

PR #573, opened 2018-06-29
New activity indicator `newtonCradle` and `circlePendulum`

phimage:
newton
circlependulum
slow motion gif

### DIFF
--- a/Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationCirclePendulum.swift
+++ b/Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationCirclePendulum.swift
@@ -14,16 +14,16 @@ public class ActivityIndicatorAnimationCirclePendulum: ActivityIndicatorAnimatin
 
   // MARK: Properties
 
-  fileprivate let duration: CFTimeInterval = 1
+  public var duration: CFTimeInterval = 0.8
   fileprivate let ratio: CGFloat = 7
   fileprivate let ballCount: Int = 3
 
   #if LG
-  fileprivate let trueColor = true
+  public var primaryColor = true
   #else
-  fileprivate let trueColor = false
+  public var primaryColor = false
   #endif
-  fileprivate let colors: [UIColor] = [.red, .blue, .yellow]
+  fileprivate let colors: [UIColor] = [.magenta, .yellow, .cyan]
 
   fileprivate var ballSize: CGFloat = 0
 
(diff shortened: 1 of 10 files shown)


Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationCirclePendulum.swift
```diff
@@ -0,0 +1,97 @@
...
+
+public class ActivityIndicatorAnimationCirclePendulum: ActivityIndicatorAnimating {
+
+  // MARK: Properties
+
+  fileprivate let duration: CFTimeInterval = 1
+  fileprivate let ratio: CGFloat = 7
+  fileprivate let ballCount: Int = 3
+
+  #if LG
```
tbaranes: What means `LG`?

tbaranes (approved):

Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationCirclePendulum.swift
phimage: Life's Good
_This activity indicator has been inspired from my new tv, to watch the world cup_

This override the color defined and set 3 colors.
Maybe `LG` could be used accidentally, If I could keep my eastern eggs, a longer compilation flag could be used

JakeLin: Shall we have a property instead of a custom flag? Otherwise, the user needs to set the configuration in Podfile to enable it like `build_settings['OTHER_SWIFT_FLAGS'] = '-LG'`. If you think that's fun, shall we add it to ActivityIndicators.md . BTW: I have `LG` big TV at home too 📺

JakeLin (approved):
The effect looks great, well done 👍 please have a look at the small comment.

Sources/ActivityIndicators/Animations/ActivityIndicatorAnimationCirclePendulum.swift
phimage: ActivityIndicatorAnimation classes are not very accessible, but I could make `trueColor` as a `public var`
The compilation will change only the default value.

Yes I will add in doc

ps: same trick on triforce

JakeLin: @phimage cool, thanks, LG(Life's Good), you make it even LB (Life's Better)😉

PR #575, opened 2018-07-08
Add `MaskType.rounded` for a rounded rectangle on selected corners.

phimage:
I use `UIBezierPath(roundedRect:, byRoundingCorners: , cornerRadii:)`

and `CornerSides` converted into `UIRectCorner`


JakeLin:
@phimage thanks for the PR, can you please let me know what's the difference between `MaskDesignable.rounded` and `CornerDesignable.cornerRadius`. Look like both support `cornerSides`, And use `UIRectCorner` and `UIBezierPath(roundedRect:byRoundingCorners:cornerRadii:)` to draw. Did I miss something? 

BTW: for iOS 11, we can use `layer.maskedCorners = [.layerMaxXMaxYCorner]` to make it easier to draw. But we still need to support old version of iOS 😢

phimage:
There is no difference when executed, just adding a path to the layer as I can see. (that's why I factorise some code)

For my specific usage, I generate storyboards with some user defined runtime attributes,
and want to choose the view "form" as parameter.

- Having only one user defined runtime for that allow me to make juste a string replace (ie. my tag ___MASK_TYPE___ replaced by the value in storyboard xml)
- having two or more user defined attribute for the same things -> I must do some xml parsing and edit to add xml node or remove it

The truth is I want to have all `UIBezierPath` initialiser used in `MaskType` 😆

JakeLin:
@phimage I don't mind to have the same feature in two different properties. And `MASK_TYPE` need to have a special parameter `rounded`, which is not exactly same as `cornerRadius`. My concern is can we encapsulate the code into one place? Otherwise, we need to maintain two copies of code to do the same drawing.

phimage:
I already try to factorize the code
As you can see by the edition of CornerDesignable (and the use of CornerSides)

I have no other idea to make a better factorization yet

JakeLin:
@phimage yes, it is hard, because they located in protocols. and protocol should not depend on any others. Please go ahead.

JakeLin (approved):

JakeLin:
Well done @phimage  🎉

PR #576, opened 2018-07-10
Swift 4.2 support

djbe:
This is needed to support Xcode 10 (beta).
The codebase should still be compatible with Swift 4.1 (Xcode 9).

Currently the `test_predefinedColor` tests fail, because the `iterateEnum(from:)` function in `IBEnum` is broken somehow in Swift 4.2.

Note: I didn't change the Travis file yet.


JakeLin:
@djbe thanks for your PR. Since Xcode 10 is still in beta, we can merge it yet. We will do an update when Xcode 10 release public GM version.

djbe:
@JakeLin This PR is compatible with Xcode 9, that's why I use `swift(>=4.1.50)` instead of `swift(>=4.2)`.

I've fully tested it with both Xcode versions locally, and it works (except for the `iterateEnum(from:)` mentioned above).

JakeLin:
@djbe interesting, do you know why `iterateEnum(from:)` is broken? Only broken for the unit tests but it is OK for the Library and the Demo app? Or `iterateEnum(from:)` doesn't work in Swift 4.2 at all?

tbaranes:
Didn't swift 4.2 add an equivalent to `iterateEnum`? It would be worth to check it out to see if we can use it directly.

djbe:
@tbaranes They added `CaseIterable` yes, but that can only be added to the original declaration, and it's something that doesn't exist in Swift 4.1. If we were to use it, we'd have to duplicate whole enums to support it.

djbe:
@JakeLin From a quick test, it's also broken in the example app: the `makeRandomGradient()` crashes  because the `predefinedGradients` collection is empty (because it can't iterate):
https://github.com/IBAnimatable/IBAnimatable/blob/master/IBAnimatableApp/IBAnimatableApp/Playground/Utils/UIViewControllerExtension.swift#L13

djbe:
Huh, initial documentation seems to be wrong, I can add `CaseIterable` in extensions 🎉

djbe:
Aha, so there is a limitation:
> Implementation of 'CaseIterable' cannot be automatically synthesized in an extension in a different file to the type

Sources/Common/SwiftSupport.swift
```diff
@@ -0,0 +1,102 @@
+//
+//  Created by David Jennes on 7/10/18.
+//  Copyright © 2018 IBAnimatable. All rights reserved.
+//
+
+import UIKit
+
+/// Swift < 4.2 support
+#if !(swift(>=4.2))
+enum CAMediaTimingFunctionName {
```
JakeLin, line 10: Great idea 👍

IBAnimatableApp/IBAnimatableApp/AppDelegate.swift
```diff
@@ -38,3 +38,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
   }
 
 }
+
+/// Swift < 4.2 support
+#if !(swift(>=4.2))
```
JakeLin, line 43: Shall we put those in one place?

Sources/Animators/TransitionAnimator/SystemTransitionAnimator.swift
```diff
@@ -53,7 +53,11 @@ class SystemTransitionAnimator: NSObject, AnimatedTransitioning {
   fileprivate func performSystemTransition(using context: UIViewControllerContextTransitioning) {
     CALayer.animate({
       let transition = CATransition()
+      #if swift(>=4.2)
+      transition.type = CATransitionType(rawValue: systemTransitionType.rawValue)
```
JakeLin, line 57: Finally,  we have non-stringy type 🎉

JakeLin (commented):
@djbe thanks for fixing that, the build is green now. I am still thinking shall we wait for a bit to merge it when Xcode 10 GM is out.

IBAnimatableApp/IBAnimatableApp/AppDelegate.swift
djbe: These are for the demo app, they shouldn't influence the code of the library itself.

JakeLin:
@tbaranes @phimage how do you think? shall we wait for Xcode 10 GM and make a big release for it?

I am also thinking, shall we keep or remove those `#if swift(>=4.2)` code after Xcode 10 GM? Because most of the developer will upgrade their Xcode to version 10 once it is released publicly.

djbe:
@JakeLin when to merge is up to you. It would just be useful for developers preparing their app for the next iOS/Xcode release to have this library be compatible with the new swift version. As stated before, the changes in this PR are **backwards compatible**.

I would not remove the `#if swift(>=4.2)` bits, as it would break the library for users with Xcode 9. Keep in mind that quite a large user base doesn't directly upgrade to Xcode 9, because of bugs, company policies, lack of time, etc...

JakeLin:
@djbe I would like to consult our team member before I go ahead to merge it. We have a small Slack channel for our core developers, would you like to join us? If yes, please send me your Slack email to my email. You can find it on my profile page, thanks.

tbaranes:
If it's compatible with swift 4.1, I'm completely ok to merge it right now! It will make the migration much easier.

JakeLin (approved):
👍 the first step to Swift 4.2 🎉

JakeLin:
@djbe After discussed with team and they have done a few of tests, it looks good, let's go forward, thanks 👍

Sources/Common/SwiftSupport.swift
SD10: This could help me with another project. Thanks for the trick @djbe 💯

PR #588, opened 2018-09-18
Fix the build for Xcode 10

### LINKED ISSUE #587: CODE IS BREAKING ON XCODE 10 GM WITH SWIFT 4.2

JakeLin:
Fix #587

We need to make the `enum CornerSide` to be `CaseIterable`. And then we can `iterateEnum`. Compiled in Xcode Version 10.0 (10A254a)


tbaranes (approved):

phimage (approved):
CI is green

PR #590, opened 2018-09-20
Fix support of older Swift versions for some users

### LINKED ISSUE #585: 'Style' is not a member type of 'UIBlurEffect'
Hi there, 

I just tried to rebuild an old app using  `IBAnimatable 5.1.0`

How to solve that?

```
/IBanimatable/BlurEffectStyle.swift
8: extension UIBlurEffect.Style: IBEnum {   # <---
```

djbe:
Fixes #585.

Finally had a colleague with the same issue, and it was pretty easy to find a solution, we had something similar with [StencilSwiftKit](https://github.com/SwiftGen/StencilSwiftKit).

The issue is that, on all current swift versions (will be fixed in 5?), in some cases the swift compiler fails to find extensions defined in other files, if the filename is (alphabetically?) later than the current file. See https://bugs.swift.org/browse/SR-631.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ N/A
 
 #### Bugfixes
 
-N/A
+- Fix support for older Swift versions for some users. [#590](https://github.com/IBAnimatable/IBAnimatable/pull/590) by [@djbe](https://github.com/djbe)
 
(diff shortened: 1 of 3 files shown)


JakeLin (approved):
Thank you very much, I didn't know that trick.

djbe:
@JakeLin You might consider releasing this as a bugfix 5.2.1, in case (more) older Xcode users start getting the #585 issue.

PR #596, opened 2018-10-28
zoom changes

LuvGulati:


JakeLin:
@LuvGulati thanks for your PR, would you please let us know why do we need to change that?

phimage:
no activity and response on pr, I close it

PR #597, opened 2018-10-28
Fix all contributors link broken

cruisediary:
Fix all contributors link broken

from https://github.com/IBAnimatable/IBAnimatable/blob/master/graphs/contributors 404 Not found


JakeLin (approved):
Thanks

PR #611, opened 2019-05-07
Use a lexer/parser to support imbricated options for `IBEnum` created from user defined runtime attributes

phimage:
ie. for animation type we can do now [shake+pop,slide(10)]
https://github.com/IBAnimatable/IBAnimatable/pull/520#issuecomment-376091296

see more example in AnimationTypeTests

- parsing is done by Udra.swift: token list, lexer to split into tokens(using regex) and finally parser to create nodes
- all `params[safe:` has been replaced by more convenient functions
- `AnimationType` is the only one that support array and operator for the moment

Udra (user defined runtime attributes)


phimage:
for the fun I do numeric  operation `+`evaluation (more operators could be added)
`dash(10+10,10)` will be equals to `dash(20,10)`

## What I can do next 

### variables
supporting variables in parser (using for instance a token $) could allow to have some evaluated value like $screenHeight/5

if there is good user case I will make it

EDIT: I make some test variables parsing
```swift
types["shake($test)"] = .shake(repeatCount: 1) // without value the default one is used
IBParser.variables["test"] = 10  // Find a way to pass a context, a list of variable values
types["shake($test)"] = .shake(repeatCount: 10)
```
or using the controller and key path
```swift
shake(\test) //or 
shake(\Controller.test)
```
but because enum variable values are int, double etc.. and not closure which return a result,
the parser will evaluate immediately the value when creating the view controller
So the variable values must be set before that
(maybe variable could be global or from view controller)

### custom functions with params

for instance $myfunction(5,10)
like variable but with parameters

### traits

I think about allowing to have different value according to device ie. variation traits
For instance presenting a modal controller on ipad or iphone 5 could not be with same size

JakeLin (approved):
That's very cool, can you rebase?

PR #616, opened 2019-08-23
Use `Hasher` to implement `Hashable`

phimage:
better late than never

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ N/A
 
 #### Enhancements
 
-N/A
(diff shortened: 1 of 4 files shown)


tbaranes (approved):

phimage:
ci failed
doc/changelog maybe :p

JakeLin (approved):
Well done, Shall we deprecate Swift 4?

phimage:
> Well done, Shall we deprecate Swift 4?

it seems that xcode 11 build  swift 4 too, not only swift 4.2
https://docs.swift.org/swift-book/GuidedTour/Compatibility.html

PR #617, opened 2019-08-23
[WIP] For iOS13 do not use `automatic` modal presentation style for segue

phimage:
Segues set using storyboards, such as `PresentSlideSegue`, seem to failed to present

I see that iOS 13 add new navigation stack and `automatic` modal presentation style
Before `fullscreen`(=0) was the default one (If I debug)

I am wondering If IBAnimatable must force to fullscreen the modal presentation style like I do in my PR when the segue is used (to make it work)

or document and say you must select Presentation: Full Screen  (or current context if previous was also full screen etc...) 
Screenshot 2019-08-23 at 10 32 14
(xcode 11 screenshot, must not be available in xcode 10)


tbaranes:
I do think you have the right approach, won't we have issue with all the other segue overriden? I don't remember exactly how it was working.

JakeLin:
Because Segue doesn't support `@IBInspectable `, we have to set a default value for one Segue, the walkaround is to create different segue for different `Presentation`. 

If the current Segues only support fullscreen, shall we change the method  `func present(animated flag: Bool = true)` to `func presentFullScreen(animated flag: Bool = true)`?

phimage:
I can rename but I will make my mode less intrusive, so it is not always fullscreen

Set `fullscreen` only if in `automatic`, so if someone choose explicitly an other presentation it will continue to work. And if he want to have issues he can choose `pageSheet ` :)
But to compile with xcode 10.3 I must do something, maybe use rawValue of automatic of check on swift version 5.1...

phimage:
I has made my code more restrictive to fix the segue but!

I see that fixing the segue is not the solution
In you test the demo app, which is not using segue, the Transition part failed to "present" things too. The same code is used after all

Maybe some code to fix into UIPresentationController override or delegates

mingsai:
In Xcode 11.3, I had to apply the Presentation fullscreen setting onto specific View Controllers in order for the segue to complete successfully. Making sure to use the custom segues from IBAnimatable now works again.

PR #618, opened 2019-08-24
Use `firstIndex` in demo app #no-public-changes

phimage:
remove some warnings but allow to compile with swift 4 if we want to...
I put the code where we need it, this is the demo app, not the wonderful framework ;)


JakeLin (approved):
forward compatibility 😍

PR #623, opened 2019-11-08
Fix swift build by adding missing import and updating package.swift #no-public-changes

phimage:
opening with Xcode the Package.swift file or by using command line
```sh
sdk=`xcrun -sdk iphonesimulator -show-sdk-path`
sdkVersion=`echo $sdk | sed -E 's/.*iPhoneSimulator(.*)\.sdk/\1/'`
swift build  -Xswiftc "-sdk" -Xswiftc "$sdk" -Xswiftc "-target" -Xswiftc "x86_64-apple-ios$sdkVersion-simulator"
```

In fact swift build do not allow missing import

I think new release must be done with that, a 6.0.1 or 6.1.0

--

Also I have added a test with github workflow, but to be activated on IBAnimatable organisation
*  an admin must active it here https://github.com/features/actions

### DIFF
--- a/Sources/Enums/Utils.swift
+++ b/Sources/Enums/Utils.swift
@@ -4,6 +4,7 @@
 //
 
 import Foundation
+import UIKit
 
 extension Array {
   /// Returns the element at the specified index iff it is within bounds, otherwise nil.
@@ -116,13 +117,6 @@ func iterateEnum<T: Hashable>(from: T.Type) -> AnyIterator<T> {
   }
 }
 #endif
-
-import UIKit
-extension CALayer {
-    var currentMediaTime: CFTimeInterval {
-       return convertTime(CACurrentMediaTime(), from: nil)
-    }
-}
 
 extension String {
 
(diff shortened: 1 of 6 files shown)


tbaranes:
Thanks for this fix, I enabled GitHub Actions as you asked :)

Sources/Enums/Utils.swift
```diff
@@ -117,6 +117,7 @@ func iterateEnum<T: Hashable>(from: T.Type) -> AnyIterator<T> {
 }
 #endif
 
+import UIKit
```
tbaranes: Why importing it here instead at the top of the file with the other import?

tbaranes (approved):

Sources/Enums/Utils.swift
phimage: I am thinking for later, moving the import and the code to its own file

And I see better fix, I will update, move it to `CALayerExtension` now
and move import at top for CGFloat

PR #625, opened 2019-11-11
Fix navigationBar gradient height

phimage:
When looking at warning on `UIApplication.shared.statusBarFrame.height` call I see that gradient on navigation bar is not correct

Before the height was 64 with a comment about `UIApplication.shared.statusBarFrame.height`
I think 64 was 44 for nav bar and 20 for status bar
That has been replaced by `UIApplication.shared.statusBarFrame.height`, so without original navigation bar height and so the gradient is misplaced and repeated

Now on iOS13 I set the status bar height to 0 because with the default card presentation, it is not recommanded to colorise status bar with navigation bar. We must use the window scene status bar manager. It work pretty well with the demo app with page/card presentation (the default one)

I do an optimisation also, do not execute some code if there is no gradient colors defined

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ N/A
 #### Bugfixes
 
 - Fix `swift build` [#623](https://github.com/IBAnimatable/IBAnimatable/pull/623) by [@phimage](https://github.com/phimage)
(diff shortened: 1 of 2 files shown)


tbaranes (approved):

JakeLin (approved):
👍

PR #626, opened 2019-11-11
🪁 Add `kite` mask type and fix swiftlint warnings

phimage:
Definition https://en.wikipedia.org/wiki/Kite_(geometry)

* with an angle between 0 and 89
Screenshot 2019-11-11 at 05 49 02
* 90 is a triangle

* with an angle between 91 and 180 we have a `dart`
Screenshot 2019-11-11 at 05 49 12


phimage:
I know that I do two things at the same times but when fixing my swiftlint warnings, I see others so I fix it

tbaranes (approved):

PR #646, opened 2023-02-20
Fix Xcode and swiftlint warnings #no-public-changes

phimage:
Update Xcode project using Xcode 14.2 too

demo app tested

(see only the nav bar color that seems to be corrupted ; it's due to change on nav bar in storyboard since Xcode 13 maybe)


JakeLin (approved):
Thanks for fixing them 👍

